PR #1, opened 2015-07-30
Membership attribute service init

danmassie:
* Uses DynamoDB to store membership attributes. 
* Read/write endpoints provided 
  * /me protected by GU cookies
  * /user-id unprotected - needs client access token to be implemented
* Uses @adamnfish's "effective api design" ApiResponse
  * ApiResponse is the return type of all service, repos, etc
  * Errors are handled within the context of where they occur
* CloudFormation templates
  * Packer is used to create an AMI with Oracle Java 8
  * MembershipAttributeService to be used with the packer generated AMI
* Team city build and riff raff deployment


membership-attribute-service/app/Global.scala
```diff
@@ -0,0 +1,34 @@
...
+import models.{MembershipAttributes, ApiError, ApiErrors, ApiResponse}
+import play.api.Logger
+import play.api.mvc.{RequestHeader, Result, WithFilters}
+import play.filters.csrf._
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+object Global extends WithFilters(CSRFFilter(), AddEC2InstanceHeader) {
+
+  private val logger = Logger(this.getClass)
```
nlindblad: Across the other projects we use a `Logging` trait, which essentially just does this (e.g. https://github.com/guardian/identity-federation-api/blob/master/src/main/scala/com/gu/identity/oauth/util/Logging.scala)

For this simple case it makes no difference, but for Federation we ended up with some more advanced log message massaging, so it made sense to put it all in one place.

danmassie: That trait looks like its mainly for access logs and metrics rather than application logging. So its use is probably only appropriate in controllers?

membership-attribute-service/app/repositories/MembershipAttributesDynamo.scala
```diff
@@ -0,0 +1,64 @@
...
+    val membershipNumber = "MembershipNumber"
+    val tier  = "Tier"
+    val joinDate = "JoinDate"
+  }
+
+  val tableRequest =
+    new CreateTableRequest()
+      .withTableName(MembershipAttributesDynamo.tableName)
+      .withProvisionedThroughput(
+        Schema.provisionedThroughput(10L, 5L))
```
nlindblad: Move this to a configuration file?

provisioning/build-tc.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
nlindblad: Might be worth changing to `#!/bin/bash` and add `set -ex` so that it halts on errors and the commands being run are shown in the TeamCity build log.

membership-attribute-service/app/repositories/MembershipAttributesDynamo.scala
danmassie: That's actually only used in a test, so I'll move it there.

membership-attribute-service/app/Global.scala
nlindblad: Yeah, unless we want multiple log streams (access logs, metrics and application logs) separated we can just keep the inline `logger` initialisation.

danmassie: I'm going to use a request filter for the access logs btw

nlindblad: :+1:

nlindblad:
LGTM :+1: Great work!

PR #3, opened 2015-08-10
Environment specific config

danmassie:
- CODE and PROD dynamodb tables created
- cloud formation template updated to include security policies for the tables
- config file per env, specified in the upstart.conf script via an ec2 instance tag
- app secret accepted as a param to the cloud formation template and passed as a tag to the ec2 instance


markjamesbutler:
LGTM :+1:

PR #5, opened 2015-09-07
Transfer to membership

ostapneko:
This prepares the transfer from the Identity team to the Membership team

- Remove the upsert endpoint that we don't need
- Add the AMI creation recipe
- Adapt the CF template to conform to the Membership habits and to remove the dependencies on some Identity internal requirements.

### DIFF
--- a/cloudformation/membership-attribute-service.json
+++ b/cloudformation/membership-attribute-service.json
@@ -13,7 +13,7 @@
     "Stage": {
       "Description": "Environment name",
       "Type": "String",
-      "AllowedValues": [ "CODE", "PROD" ]
+      "Default" : "PROD"
     },
     "InstanceType": {
       "Description": "EC2 instance type",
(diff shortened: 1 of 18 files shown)


cloudformation/membership-attribute-service.json
```diff
@@ -303,23 +278,17 @@
           "Fn::Base64": {
             "Fn::Join": ["", [
               "#!/bin/bash -ev\n",
+              {"Fn::Join":["", ["aws --region ", { "Ref": "AWS::Region" }, " s3 cp s3://gu-membership-attribute-service-dist/set-env.sh .\n"]]},
```
dominickendrick: If we are using the `pip` version of `awscli`, we can probably remove the region lookups here and in the following 2 `aws s3` commands

dominickendrick:
:+1:

PR #6, opened 2015-09-08
Salesforce hook

ostapneko:
Add a hook that updates DynamoDB that comes from Salesforce outbound messages


membership-attribute-service/test/parsers/SalesforceTest.scala
```diff
@@ -0,0 +1,35 @@
...
+                <sObject xsi:type="sf:Contact" xmlns:sf="urn:sobject.enterprise.soap.sforce.com">
+                  <sf:Id>id</sf:Id>
+                  <sf:IdentityID__c>identity_id</sf:IdentityID__c>
+                  <sf:Membership_Number__c>membership_number</sf:Membership_Number__c>
+                  <sf:Membership_Tier__c>Supporter</sf:Membership_Tier__c>
+                </sObject>
+              </Notification>
+            </notifications>
+          </soapenv:Body>
+        </soapenv:Envelope>
```
rtyley: When the data's large like this, I prefer to take it and put it into it's own file (eg `salesforce-outbound.xml`) in `test/resources` - otherwise it tends to dominate the test!

project/Dependencies.scala
```diff
@@ -16,11 +16,12 @@ object Dependencies {
   val specs2 = PlayImport.specs2 % "test"
   val awsWrap = "com.github.dwhjames" %% "aws-wrap" % "0.7.2"
   val awsDynamo = "com.amazonaws" % "aws-java-sdk-dynamodb" % "1.9.31"
+  val scalaz = "org.scalaz" %% "scalaz-core" % "7.1.1"
 
   //projects
 
   val apiDependencies = Seq(identityCookie, identityPlayAuth, scalaUri,
-    playWS, playCache, playFilters, awsWrap, awsDynamo,
+    playWS, playCache, playFilters, awsWrap, awsDynamo, scalaz,
```
rtyley: Did compile time go up when you added scalaz? :-)

ostapneko: It's alright so long as you don't do huge wildcard imports!

```scala
import scalaz._
import Scalaz._
```

is a no-no if you want to keep a sane compile time.
Having said that, I'm a bit disappointed by Scalaz so far, so if at the end of it we don't get any benefits we can just remove it. For now it's a transitive dependency anyway so it doesn't have much impact.

membership-attribute-service/test/parsers/SalesforceTest.scala
ostapneko: No worries. The advantage though is that

- You have the expected values right at hand
- You ensure that everything is well-formed
- You don't have to load the file, so less indirection

Thoughts?

dominickendrick:
:+1:

PR #9, opened 2015-09-10
Add possibility to use fixtures

ostapneko:
Add a `use-fixtures` property.

In this mode, the response for the GET endpoint is pre-canned, and defined in `app/models/Fixtures.scala`


tudorraul:
LGTM :+1:

PR #10, opened 2015-09-11
Add a shared secret with Salesforce in the query string

ostapneko:
Also modify the CF template so it uses a config file in /etc/gu


rtyley:
:+1:

PR #11, opened 2015-09-15
Check payload signature for SF hook

ostapneko:
Instead of using a querystring containing a shared secret, this relies on the payload from Salesforce being signed by a Salesforce-issued certificate.

Also changes the format in which SF sends contact info.

This relies on some Apex code being present in Salesforce.

@rtyley

To ease the work in parallel with @tudorraul, the following points will be part of separate pull requests:

- Testing
- Download of the Salesforce certificate instead of relying on the one in the repo
- Logging
- Cloudwatch metrics


ostapneko:
Abandoned as the Outbound messages solution was preferred

PR #13, opened 2015-09-18
HTTPS cloudwatch configuration

tudorraul:
Configuration necessary to cloudform https://members-data-api.theguardian.com/


tudorraul:
@rtyley

rtyley:
Could you add a description telling me what problem this is aiming to solve?

chrisjowen:
This is already live, so :+1: and now there is a description which clarify the change to SSL

PR #15, opened 2015-09-23
Avoid overriding existing rows in batch load

ostapneko:
Make sure that the rows we import from the CSV during initial loading don't override whatever rows already exist in the DB.

The records in a fresh DB that is updated only by outbound messages are always fresher than the ones coming from the CSV file.

@dominickendrick @rtyley


dominickendrick:
:+1:

PR #17, opened 2015-09-24
Instantiate Dynamo client with a creds provider

ostapneko:
We currently instantiate the Dynamo client with credentials, which eventually expire. This uses a credentials provider instead, which is in charge of refreshing the creds.


tudorraul:
:+1:

PR #19, opened 2015-09-24
delete entries that no longer belong to members

tudorraul:
Upon membership cancellation, user data should be deleted from members attributes service.


tudorraul:
@ostapneko

ostapneko:
:+1:

PR #21, opened 2015-09-24
Add adfree endpoint

ostapneko:
Add an adfree enpoint at `GET /user-attributes/me/adfree` that returns a boolean result and drops a `gu_adfree_user` cookie


tudorraul:
:+1:

PR #22, opened 2015-09-25
Fix friend tier adfree allowance

tudorraul:
As a friend, I am entitled to an adfree experience. This is counter to our intent.


ostapneko:
:+1:

PR #23, opened 2015-09-25
return 404 if user is not found in db

tudorraul:
At the moment we return Unauthorised code, even though the authentication cookie is present and valid.


davidrapson:
:+1:

PR #24, opened 2015-09-25
Handle empty membership number

ostapneko:
Some member (including all friends) have an empty membership number, which cause their import to fail. This allow for an empty membership number


tudorraul:
:+1:

PR #28, opened 2015-09-29
Simplify the codebase

ostapneko:
- Remove unnecessary action abstractions
- Remove runtime DI
- Distinguish between set and delete actions


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -1,21 +1,22 @@
...
+import play.api.mvc.{Action, Cookie, Result}
+import services.{AttributeService, AuthenticationService, DynamoAttributeService, IdentityAuthService}
 
 import scala.concurrent.Future
 
-class AttributeController @Inject() (attributeService: AttributeService) extends CommonActions {
+class AttributeController {
   val ADFREE_COOKIE_MAX_AGE = 60 * 60 * 6 // 6 hours
+  lazy val authenticationService: AuthenticationService = IdentityAuthService
+  lazy val attributeService: AttributeService = DynamoAttributeService
```
tudorraul: I think the type hints are superfluous here.

tudorraul:
:+1: bar the comment

membership-attribute-service/app/controllers/AttributeController.scala
ostapneko: Nope, the type on the LHS is not the same as in the RHS. If you omit the type the class will expect concrete services instead of traits.

PR #35, opened 2015-10-06
Add support for test users

ostapneko:
- Also remove support for fixtures

@rtyley


rtyley:
:+1: ...though I feel that on the cookie-secured endpoints, my expectations lead me to have the authorization as an `action`, rather than mapping within the controller method.

PR #37, opened 2015-10-07
Add a gu_adblock_message cookie

ostapneko:
Add a cookie that drives the ad banner for membership, as per [the Trello ticket](https://trello.com/c/STuEhVmU/181-add-a-cookie-for-membership-banner) that come from a requirement from the commercial team

SUBJECT TO DEBATE

@davidrapson @rtyley

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -29,7 +29,10 @@ class AttributeController {
 
   private def adfreeResponse(adfree: Boolean) =
     Ok(Json.obj("adfree" -> adfree, "issuedAt" -> scala.compat.Platform.currentTime))
-      .withCookies(Cookie("gu_adfree_user", adfree.toString, maxAge = Some(ADFREE_COOKIE_MAX_AGE)))
+      .withCookies(
+        Cookie("gu_adfree_user", adfree.toString, maxAge = Some(ADFREE_COOKIE_MAX_AGE)),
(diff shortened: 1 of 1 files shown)


rtyley:
I'm confused as to why we have two identical-looking cookies here - can you explain what's going on in a bit more detail?

ostapneko:
So:
  - the adfree cookie drives the display of the ads and can be driven by user preferences
  - the adblock cookie drives the display of the adblock banner and has nothing to do with the user preferences

@kenlim made a good case that the policy that drives the display of this banner should be owned by membership, so the commercial team shouldn't have to infer it from the membership tier for instance.

Also they probably don't want to update their code when we decide to separate those policy. And for us the change is minimal.

@rtyley

davidrapson:
I get the reason for the separate cookies, but I think the naming convention is far too granular. Why drop `gu_adblock_message` which seems like a *very* specific case? Is the thing they need to know in this case just that they are a Member, if so should we just drop that as `gu_is_member` or something similar?

Also, worth asking but do *we* need to be doing the cookie dropping? Surely *how* the preference is stored is better determined in frontend. i.e., I would think storing the preference in localStorage would be better in many cases which we can't do from this endpoint. I'd think they just need us to own the answer and then they determine how best to store that answer for their use-case.

Am I chatting bubbles?

ostapneko:
@davidrapson You're probably right on the storage mechanism; in particular I'm not sure what would happen for mobile... I remember having a discussion with the frontend team where some were positive they needed a cookie but why? I'll ask again

ostapneko:
Made obsolete by the decision to send a simple JSON payload instead of a cookie

davidrapson:
:gem:

PR #39, opened 2015-10-09
Ensure that the org id is 18 chars

ostapneko:


rtyley:
:+1: bless salesforce.

PR #40, opened 2015-10-09
Change the www.thegulocal to m.thegulocal in CORS

ostapneko:
This is to conform to the local domain that is chosen in the front end repo. See https://github.com/guardian/frontend/blob/master/nginx/frontend.conf

@rtyley


rtyley:
:+1:

PR #41, opened 2015-10-13
Fix sentry

ostapneko:
Actually start Sentry on application startup, and also move the config to /etc/gu
@rtyley @afiore @chrisjowen @tomverran


rtyley:
:+1: - do you need to regenerate the Sentry DSN, as this one seems to be compromised?

ostapneko:
@rtyley Please - I don't know how critical this is; @tudorraul didn't seem to worry too much

rtyley:
Done - will message you

PR #42, opened 2015-10-13
Increase cookie max-age to 3 months

ostapneko:
As per the commercial team recommendation

@jbreckmckye @rtyley


ostapneko:
Made obsolete by the decision to send a simple JSON payload instead of a cookie

PR #45, opened 2015-10-22
Only provide adfree to commercial team members

ostapneko:
The commercial team want to test this feature with their own accounts only before they roll it out to the general public

@rtyley @kenlim @tomverran 

Note that this is not even smoke tested as the CODE Identity API is down ATM


ostapneko:
bump @rtyley @kenlim @tomverran

tomverran:
looks ok to me :+1:

tomverran:
identity CODE API is back btw

PR #48, opened 2015-11-05
Re-enable the "adfree" whitelist

kenlim:
If we are to test the adfree feature live on the site, ad free capable users should be confined to the whitelist of dev accounts (which was previously provided). 

Reverts pull request: guardian/membership-attribute-service#47


ostapneko:
:+1: , but merge only when I get the whitelist

kenlim:
Been informed by Thomas Franquelin that the whitelist is live. Merging.

PR #49, opened 2015-11-10
Add content access details to `/user-attributes/me/membership`

rtyley:
With `contentAccess` added, https://members-data-api.theguardian.com/user-attributes/me/membership will look like this:

```
 {
   "tier": "patron",
   "membershipNumber": "abc",
   "userId": "123",
   "contentAccess":{"member":true,"paidMember":true}
 }
```

This data is going under `/membership` rather than `/features` because it directly discloses that the user is a paying customer, which we're going to consider a piece of personal information.

Note also that we're going to want Staff to be able to see 'paid' member content, even though they don't pay.

https://trello.com/c/KhTyeT4p/193-restricted-content-change-user-me-endpoint-to-member-data-api

cc @ostapneko

### DIFF
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -1,10 +1,17 @@
 package models
 
+import json._
 import play.api.libs.json._
 import play.api.mvc.Result
 import play.api.mvc.Results.Ok
 
 import scala.language.implicitConversions
+
+case class ContentAccess(member: Boolean, paidMember: Boolean)
+
+object ContentAccess {
+  implicit val jsWrite = Json.writes[ContentAccess]
+}
 
 case class Attributes(userId: String, tier: String, membershipNumber: Option[String]) {
   require(tier.nonEmpty)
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/models/Attributes.scala
```diff
@@ -1,21 +1,31 @@
...
   lazy val isFriendTier = tier.equalsIgnoreCase("friend")
   lazy val isPaidTier = !isFriendTier
+
+  lazy val contentAccess = ContentAccess(member = true, paidMember = isPaidTier) // we want to include staff!
 }
 
 object Attributes {
-  implicit val jsWrite = Json.writes[Attributes]
+  implicit val jsWrite: Writes[Attributes] =
+    Json.writes[Attributes].asInstanceOf[OWrites[Attributes]].addField("contentAccess", _.contentAccess)
```
rtyley: This annoying cast is necessary because currently the `Json.writes` macro has a declared return type of `Writes` rather than `OWrites` (a type which guarantees to return a `JsObject`, rather than just a `JsValue`).

This is actually fixed in a forthcoming release of Play: https://github.com/playframework/playframework/pull/4772

rtyley:
Looks like the tests are a little flakey in TeamCity - failed first time, passed second:

https://teamcity.gutools.co.uk/viewType.html?buildTypeId=Membership_AttributesService&branch_Membership=add-content-access-data&tab=buildTypeStatusDiv

ostapneko:
Is it worth reconciling https://github.com/guardian/membership-attribute-service/blob/master/membership-attribute-service/app/models/Features.scala#L16

with this, and use `contentAccess` there to make the logic consistent?

rtyley:
> Is it worth reconciling https://github.com/guardian/membership-attribute-service/blob/master/membership-attribute-service/app/models/Features.scala#L16
> 
> with this, and use contentAccess there to make the logic consistent?

Do you mean the user-whitelisting done at https://github.com/guardian/membership-attribute-service/blob/e2f6718d/membership-attribute-service/app/models/Features.scala#L19 ? I don't think the content access stuff needs the whitelisting- we can just apply it to everyone.

There doesn't seem to be much else in Features.scala that relates to contentAccess, especially because we don't want to include contentaccess in there because it gives away whether or not the user is a paying user?

ostapneko:
:+1:

rtyley:
Confirmed on PROD:

```
{"userId":"xxxxxx","tier":"Staff","membershipNumber":"yyyyy","contentAccess":{"member":true,"paidMember":true}}
```

PR #50, opened 2015-11-12
MMA API endpoints

tomverran:
These will need to change when the subs model changes go live, but not that much


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -0,0 +1,50 @@
...
+
+    private def memberDetails(contact: Contact[MemberStatus, PaymentMethod]): JsObject = contact.memberStatus match {
+      case m: Member => Json.obj("regNumber" -> m.regNumber.mkString, "tier" -> m.tier.name, "isPaidTier" -> m.tier.isPaid)
+      case _ => Json.obj()
+    }
+
+    private def basicDetails(contact: Contact[MemberStatus, PaymentMethod]) =
+      Json.obj("joinDate" -> contact.joinDate)
+
+    private def toJson(paymentDetails: PaymentDetails): JsObject = {
```
afiore: Why not creating  `Write[PaymentDetails]` and of `Writer[Card]` instances here?

tomverran: So this was taken from membership-frontend and since the JSON format is a bit odd and specific to the MMA page I didn't think it'd be worth making it reusable

membership-attribute-service/app/actions/WithTouchpointFromCookieAction.scala
```diff
@@ -0,0 +1,19 @@
...
+import play.api.mvc.{Result, Request, ActionRefiner}
+import services.IdentityAuthService
+import play.api.libs.concurrent.Execution.Implicits._
+import scala.concurrent.Future
+
+object WithTouchpointFromCookieAction extends ActionRefiner[Request, TouchpointRequest] {
+  override protected def refine[A](request: Request[A]): Future[Either[Result, TouchpointRequest[A]]] = Future {
+    val backendConf =
+      if (IdentityAuthService.username(request).exists(Config.testUsernames.isValid))
+        TouchpointBackendConfig.byEnv(Config.testTouchpointBackendStage, Config.config.getConfig("touchpoint.backend"))
```
afiore: can this file be simplified a bit? Like at least creating a val for `Config.config.getConfig("touchpoint.backend") `

membership-attribute-service/app/actions/package.scala
```diff
@@ -1,9 +1,13 @@
+import com.gu.membership.touchpoint.TouchpointBackendConfig
+import com.gu.monitoring.ServiceMetrics
 import configuration.Config.BackendConfig
 import play.api.mvc.{WrappedRequest, Request, Action}
 import services.AttributeService
 
 package object actions {
   val BackendFromCookieAction = Action andThen WithBackendFromCookieAction
+  val TouchpointFromCookieAction = Action andThen WithTouchpointFromCookieAction
```
ostapneko: Can we reconcile this with `backendAction`?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -40,4 +49,27 @@ class AttributeController {
     onSuccess = Features.fromAttributes,
     onNotFound = Features.unauthenticated
   )
+
+  def membershipDetails = paymentDetails("Membership")
+  def digitalPackDetails = paymentDetails("Digital Pack")
+
+  def paymentDetails(service: String) = TouchpointFromCookieAction.async { implicit request =>
+    authenticationService.userId.fold[Future[Result]](Future(cookiesRequired)){ userId =>
```
ostapneko: This should be moved out of the controllers, and the services should be instantiated once per backend type

membership-attribute-service/app/framework/DualRouter.scala
```diff
@@ -0,0 +1,17 @@
+package framework
+
+import play.api.mvc.{Handler, RequestHeader}
+import play.api.routing.Router
+import play.api.routing.Router.Routes
+
+/**
+  * Created by tverran on 23/11/2015.
```
ostapneko: !!!

PR #51, opened 2015-11-12
Support authentication through Identity Access Tokens as well as cookies

rtyley:
The Guardian mobile apps authenticate through Identity Access Tokens rather than Cookies (ie the SC_GU_U, etc cookies), and so we want to support the mobile clients on `members-data-api.theguardian.com`.

https://trello.com/c/d7jD8CUW/120-members-data-api-enable-access-by-mobile-access-tokens

See also https://github.com/guardian/identity-play-auth/pull/4

cc @tomverran @ostapneko @markjamesbutler


tomverran:
:+1:

PR #53, opened 2015-11-13
Don't block for EC2 Instance-Id Metadata

rtyley:
The `X-EC2-instance-id` header is useful for us when we trawl through Fastly access logs, and the EC2 metadata call only occurs once on start-up, but in the dev environment we restart frequently, and don't want to wait the 30-seconds or so it takes for the request to time-out!

The behaviour was a little strange to observe before we knew what was going on - we would curl http://localhost:9400/healthcheck, and the application log would immediately say a response had been returned:

```
2015-11-13 10:02:12,423 [ForkJoinPool-1-worker-1] INFO  play.api.Play$ - Application started (Dev)
2015-11-13 10:02:12,729 [application-akka.actor.default-dispatcher-6] INFO  filters.LogRequestsFilter$ - REQUEST 127.0.0.1 GET /healthcheck took 194 ms and returned 200
```

...but curl wouldn't return a response for at least half a minute!

Thanks to @ostapneko for investigating this with me :microscope:

### DIFF
--- a/membership-attribute-service/app/filters/AddEC2InstanceHeader.scala
+++ b/membership-attribute-service/app/filters/AddEC2InstanceHeader.scala
@@ -10,11 +10,13 @@ import scala.concurrent.Future
 object AddEC2InstanceHeader extends Filter {
 
   // http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
-  lazy val instanceIdOptF = WS.url("http://169.254.169.254/latest/meta-data/instance-id").get().map(resp => Some(resp.body).filter(_.nonEmpty)).recover { case _ => None }
+  lazy val instanceIdF = WS.url("http://169.254.169.254/latest/meta-data/instance-id").get().map(_.body)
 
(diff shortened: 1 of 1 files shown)


rtyley:
Confirmed on PROD that in a real AWS environment, the instance id still comes through:

```
$ curl -I https://members-data-api.theguardian.com/user-attributes/me/features
HTTP/1.1 401 Unauthorized
Content-Length: 61
Content-Type: application/json; charset=utf-8
Date: Fri, 13 Nov 2015 12:10:52 GMT
Set-Cookie: PLAY_SESSION=4b6a544b45a3ff4421ed0f60fb4d82315ba97da8-csrfToken=60ffa5270b672939f4c1377e70bfc15d95ce2d77-1447416652338-34b1c80ec289974346bb5c86; Path=/; HTTPOnly
X-EC2-instance-id: i-f98f4d40
Connection: keep-alive
```

rtyley:
Note this was re-encountered in membership-frontend, and fixed there with https://github.com/guardian/membership-frontend/pull/1426

PR #55, opened 2015-11-19
Subscription information with globals instead of di

tomverran:


membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -0,0 +1,43 @@
...
+import play.api.Play.current
+import services.{AttributeService, DynamoAttributeService}
+import play.api.libs.concurrent.Execution.Implicits._
+
+trait TouchpointComponents {
+
+  val stage: String
+  val sfConfig: SalesforceConfig
+
+  lazy val conf = Config.config.getConfig("touchpoint.backend")
```
afiore: Not sure it is a good idea to make these lazy, as this will make harder to spot configuration errors which would be otherwise evident right away.

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -40,4 +47,21 @@ class AttributeController {
     onSuccess = Features.fromAttributes,
     onNotFound = Features.unauthenticated
   )
+
+  def membershipDetails = paymentDetails({t: TouchpointComponents => t.membershipPlans})
+  def digitalPackDetails = paymentDetails({t: TouchpointComponents => t.digitalPackPlans})
+
+  def paymentDetails(func: (TouchpointComponents)=>ProductFamily) = backendAction.async { implicit request =>
```
afiore: Can't this just accept a `ProductFamily` ?

tomverran: unfortunately we can't refer to a specific instance of a ProductFamily without an HTTP request as the rate plan IDs contained within vary depending on whether you're a test user or not

afiore:
```scala

sealed trait ProductFamily
case object DigiPack
case object Membership




````

File: membership-attribute-service/app/components/TouchpointComponents.scala [hunk 3]
```diff
@@ -0,0 +1,42 @@
...
+  lazy val sfSecret = environmentConf.getString("salesforce.hook-secret")
+  lazy val dynamoTable = environmentConf.getString("dynamodb.table")
+
+  lazy val digitalPackPlans = DigitalPack.fromConfig(digitalPackConf)
+  lazy val membershipPlans = Membership.fromConfig(membershipConf)
+
+  lazy val tpConfig = TouchpointBackendConfig.byEnv(stage, conf)
+  lazy val metrics = new ServiceMetrics(tpConfig.zuoraRest.envName, Config.applicationName,_: String)
+
+  lazy val stripeService = new StripeService(tpConfig.stripe, metrics("stripe"), WS.client)
```
afiore: Probably need to force the instantiation at boot time

afiore:
:+1: except for comments

membership-attribute-service/app/components/TouchpointComponents.scala
tomverran: The conclusion to this was that it will actually be okay

ostapneko:
:+1:

PR #56, opened 2015-11-26
No CORS by default, only allow profile.theguardian.com for MMA

tomverran:


afiore:
:+1: looks good

tomverran:
Confirmed on PROD

I wrote some runscope tests to make sure we remain locked down

PR #58, opened 2015-12-01
Membership common 107

tomverran:
The MMA page won't work very well with multicurrency but I can fix that in once we can have float multicurrency prices

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -45,6 +45,7 @@ object AccountDetails {
           "plan" -> Json.obj(
             "name" -> paymentDetails.plan.name,
             "amount" -> paymentDetails.plan.amount.amount * 100,
+            "currency" -> paymentDetails.plan.amount.currency.glyph,
             "interval" -> paymentDetails.plan.interval.mkString
           )))
       )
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -44,7 +44,8 @@ object AccountDetails {
           "trialLength" -> paymentDetails.remainingTrialLength,
           "plan" -> Json.obj(
             "name" -> paymentDetails.plan.name,
-            "amount" -> paymentDetails.plan.amount,
+            "amount" -> paymentDetails.plan.amount.amount * 100,
+            "currency" -> paymentDetails.plan.amount.currency.glyph,
```
afiore: It would be probably better to put the currency 3-letters code here instead of the glyph.

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -44,7 +44,8 @@ object AccountDetails {
           "trialLength" -> paymentDetails.remainingTrialLength,
           "plan" -> Json.obj(
             "name" -> paymentDetails.plan.name,
-            "amount" -> paymentDetails.plan.amount,
+            "amount" -> paymentDetails.plan.amount.amount * 100,
```
afiore: wouldn't be a good idea to rename plan amount into plan `price`?

afiore:
:+1: overall ok

tomverran:
shall do! The multicurrency stuff is likely to change anyway

PR #60, opened 2015-12-02
Return a JSON object and a 200 even if no MMA info can be found

tomverran:
This is because the old MMA page expects this behaviour


ostapneko:
:+1: if it stays only a few days

tomverran:
definitely! I like my HTTP APIs to use HTTP response codes

PR #64, opened 2015-12-03
Update stripe endpoint

tomverran:
Push CC updates straight through to Zuora

### DIFF
--- a/membership-attribute-service/app/configuration/Config.scala
+++ b/membership-attribute-service/app/configuration/Config.scala
@@ -49,6 +49,13 @@ object Config {
   val mmaCorsConfig = CORSConfig.denyAll.copy(
     allowedOrigins = config.getStringList("mma.cors.allowedOrigins").asScala.toSet
   )
+
+  lazy val mmaCardCorsConfig = Config.mmaCorsConfig.copy(
+    isHttpHeaderAllowed = Seq("accept", "content-type", "csrf-token").contains(_),
+    isHttpMethodAllowed = _ == "POST",
+    supportsCredentials = true
+  )
+
   // TODO: remove once the adfree feature is generally available to the public
   val preReleaseUsersIds = config.getStringList("identity.prerelease-users").asScala.toSet
 }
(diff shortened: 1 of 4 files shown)


membership-attribute-service/conf/DEV.conf
```diff
@@ -4,7 +4,6 @@ include "application.conf"
 
 identity {
   production.keys = false
-  test.users.secret="a-non-secure-key-for-our-dev-env-only"
```
tomverran: this was removed as it breaks test users on dev which is pretty annoying

membership-attribute-service/app/configuration/Config.scala
```diff
@@ -49,6 +49,13 @@ object Config {
   val mmaCorsConfig = CORSConfig.denyAll.copy(
     allowedOrigins = config.getStringList("mma.cors.allowedOrigins").asScala.toSet
   )
+
+  lazy val mmaCardCorsConfig = Config.mmaCorsConfig.copy(
+    isHttpHeaderAllowed = Seq("accept", "content-type", "csrf-token").contains(_),
```
afiore: Shoudn't these headers be capitalised?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -49,6 +55,25 @@ class AttributeController {
...
+    val updateForm = Form { single("stripeToken" -> nonEmptyText) }
+    val productFamily = request.touchpoint.ratePlanIds(product)
+    val tp = request.touchpoint
+
+    (for {
+      user <- OptionT(Future.successful(authenticationService.userId))
+      sfUser <- OptionT(tp.contactRepo.get(user))
+      subscription <- OptionT(tp.subService.findByProductFamily(sfUser, productFamily))
+      stripeCardToken <- OptionT(Future.successful(updateForm.bindFromRequest().value))
+      result <- OptionT(tp.subService.setPaymentCardWithStripeToken(subscription.accountId, stripeCardToken))
```
afiore: result is unbound... can it be replaced with `_`

membership-attribute-service/app/configuration/Config.scala
tomverran: turns out the cors filter does some normalisation after all so the case is correct

afiore:
:+1: looks good

PR #71, opened 2016-01-05
Allow the "origin" header safari sends along with CORS requests

tomverran:
I am struggling with their dev tools but I think they send origin during the preflight access-control-request-headers


afiore:
:+1:

PR #78, opened 2016-02-01
Revert "Enable CloudWatch logs"

tomverran:
Reverts guardian/membership-attribute-service#77

We were missing what turned out to be quite an important newline from the bash script


mario-galic:
:+1:

PR #79, opened 2016-02-02
Add a logger.conf for cloudwatch

tomverran:
We're soft deploying the logging by modifying the cloudformation in AWS but it fails due to this file not being in git


tomverran:
@AWare

AWare:
🤖
:+1:

PR #82, opened 2016-02-05
Make controllers singletons

mario-galic:
@rtyley @paulbrown1982 

Fixes runaway thread pool instantiation in CloudWatch class where each
request created a new thread pool at [CloudWatch.scala#L23](https://github.com/guardian/membership-attribute-service/blob/8a18590fd0a55bd0e46cd6489b62f261ef1f8a70/membership-attribute-service/app/monitoring/CloudWatch.scala#L23):

```
    val client = new AmazonCloudWatchAsyncClient(new DefaultAWSCredentialsProviderChain)
```

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -12,17 +12,18 @@ import play.api.data.Form
 import play.api.data.Forms._
 import play.api.libs.concurrent.Execution.Implicits._
 import play.api.libs.json.Json
-import play.api.mvc.Result
+import play.api.mvc.{Controller, Result}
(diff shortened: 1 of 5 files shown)


rtyley:
:+1: The newly introduced Dependency Injection in Play 2.4 is playing havoc with us here. Going to singletons is the simplest fix here...

https://www.playframework.com/documentation/2.4.x/Highlights24#Dependency-Injection

mario-galic:
We can remove `AttributeControllerTest` in a separate pull request after discussing with @tomverran

mario-galic:
The fix seems to have stabilised membership-attribute-service instances:

thread_leak_fix

PR #84, opened 2016-02-08
Less dramatic controller fix

tomverran:
So when you use play's _static_ routes generator the @ sign indicates you want your controllers to be classes with zero arguments

When you use the _injected_ routes generator however it indicates you want to use a provider function to get your controller instances, which is **re-run on every request**

From InjectedRoutesGenerator.scala
```scala
// If it's using the @ syntax, we depend on the provider (ie, look it up each time)
```

As such we can still use the InjectedRoutesGenerator and don't have to fall back to legacy static controllers if we just remove the @ signs from the routes file, which we did anyway as part of the fix.

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -12,18 +12,17 @@ import play.api.data.Form
 import play.api.data.Forms._
 import play.api.libs.concurrent.Execution.Implicits._
 import play.api.libs.json.Json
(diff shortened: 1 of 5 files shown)


tomverran:
@AWare mind taking a look?

cc @rtyley

tomverran:
Alex is searching for a gif to use as a thumbs up but has verbally given the okay so I'm merging this

AWare:
lgtm :(

PR #85, opened 2016-02-24
Support membership-common 161

tomverran:


AWare:
Does what it says on the tin. 
:+1:

PR #86, opened 2016-03-08
Update to newer versions of various libraries

tomverran:


AWare:
This PR is to make sure that everything is pointing at the latest version of identity. We can't be having things with different identity deps to their libraries. It might be worth considering how we deal with this maybe?

:+1:

PR #87, opened 2016-03-09
Fix monthly downgrades

tomverran:
Currently if you downgrade to a friend you're told your subscriber benefits will run out after the _term end date_ of your subscription. This is true for annual subscriptions where you're billed once per term but not for monthly subscriptions, where your benefits will run out after the end of your current charged period.


AWare:
:+1:

tomverran:
image

PR #91, opened 2016-04-05
Stripe webhook

tomverran:
So this adds a webhook that stripe will call to allow us to update marketing preferences in Salesforce

To ensure you can't just call this with bad information I only use the ID of the content passed through to the webhook then look up the event with the stripe API

Let me know if you see anything that looks exploitable

This also includes some misc fixes as membership-common was quite out of date


AWare:
:+1:

PR #93, opened 2016-04-05
Don't return so many 400s from the stripe hook

tomverran:
Stripe retries non 200 responses so we don't want to return 400 for this non-error case


AWare:
200 OK

paulbrown1982:
👍

PR #94, opened 2016-04-07
Update identity rather than salesforce with the webhook

tomverran:
So the stripe webhook was updating salesforce which doesn't help much
now it updates identity


AWare:
👍 0
0 :+1:

PR #97, opened 2016-04-28
Ability to ping members-data-api to update user details

tomverran:
This PR does two big things

 - Add an endpoint to trigger attribute updates
 - Move MMA methods to a new controller (they have nothing to do with attributes)


rtyley:
:+1:

PR #98, opened 2016-04-29
Use a more active SBT Dynamo plugin

philwills:
This one doesn't require a custom repo and seems to be being actively maintained.


tomverran:
👍

PR #99, opened 2016-05-06
Friendly tailor

annebyrne:
cc @philwills

### DIFF
--- a/membership-attribute-service/app/controllers/HealthCheckController.scala
+++ b/membership-attribute-service/app/controllers/HealthCheckController.scala
@@ -17,7 +17,7 @@ class BoolTest(name: String, exec: () => Boolean) extends Test {
 }
 
 class HealthCheckController extends Results {
-  val zuora = NormalTouchpointComponents.zuoraService
+  lazy val zuora = NormalTouchpointComponents.zuoraService
 
   val tests: Seq[Test] = Seq(
     new BoolTest("ZuoraPing", () => zuora.lastPingTimeWithin(2.minutes))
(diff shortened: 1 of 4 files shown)


philwills:
This adds an endpoint to help with the Discover work that allows us to retrieve from Dynamo the number of EU referendum articles that a users has seen.

If this ends up being more than an experiment, it probably wouldn't be in the Members Data API, but @rtyley suggested this would be a relatively painless way for us to get it available on a `theguardian.com` domain, so that we get the right cookies.

membership-attribute-service/app/controllers/HealthCheckController.scala
```diff
@@ -17,7 +17,7 @@ class BoolTest(name: String, exec: () => Boolean) extends Test {
 }
 
 class HealthCheckController extends Results {
-  val zuora = NormalTouchpointComponents.zuoraService
+  lazy val zuora = NormalTouchpointComponents.zuoraService
```
philwills: This hid some spurious errors for those of us who don't have Zuora credentials, but aren't working on it.

project/Dependencies.scala
```diff
@@ -17,6 +17,7 @@ object Dependencies {
   val scalaTest =  "org.scalatestplus" %% "play" % "1.4.0-M3" % "test"
   val specs2 = PlayImport.specs2 % "test"
   val awsWrap = "com.github.dwhjames" %% "aws-wrap" % "0.7.2"
+  val scanamo =  "com.gu" %% "scanamo" % "0.4.0"
```
tomverran: is this being / going to be used? I can't see any scanamo imports

tomverran:
One thing that springs to mind is that the members-data-API does not apply any CORS policies by default, so you probably need to copy what is happening in AttributesController to allow access from dotcom

rtyley:
I've added CORS, as suggested by @tomverran , and checked the output looks ok with @joelochlann -merging now.

PR #101, opened 2016-05-19
Add a 7-day threshold so only recently-read articles are considered

rtyley:
The editorial choice is that we only want to tailor articles based on a value that won't potentially just increase and increase over time (ie all articles on EU Referendum read by the user since the start of time). So the threshold for an 'expert' user can be '10 articles in the last week' for instance.

We're finally using Scanamo to read the results (thanks @philwills!) because we've changed the data format to not use a custom-column for each tag:

https://github.com/guardian/friendly-tailor/pull/11

cc @annebyrne

### DIFF
--- a/membership-attribute-service/app/repositories/BrowserIdStats.scala
+++ b/membership-attribute-service/app/repositories/BrowserIdStats.scala
@@ -1,7 +1,12 @@
 package repositories
 
-import com.amazonaws.services.dynamodbv2.model.{ComparisonOperator, Condition, QueryRequest}
-import com.github.dwhjames.awswrap.dynamodb._
+import java.time.Instant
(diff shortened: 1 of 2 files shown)


annebyrne:
👍

rtyley:
Confirmed on PROD that the `/tailor/me` API endpoint is still returning results, now that it's using our new data model that includes the time. Though I slightly regret using 'Number' to denote time when we get no benefit from it - should probably have used String and ISO time format.

https://members-data-api.theguardian.com/tailor/me?tag=politics/eu-referendum

image


https://members-data-api.theguardian.com/tailor/me?tag=politics/eu-referendum&tag=profile/roberto-tyley

PR #102, opened 2016-05-23
Add CORS headers for hitting members-data-api endpoint

annebyrne:
This will allow us to retrieve user article history to tailor additional content on Brexit (interactive-brexit-companion).

cc @philwills @rtyley


philwills:
👍

PR #103, opened 2016-05-26
Parse Guardian Identity cookies and use them to perform lookup of stats

rtyley:
This means that the page view stats returned by the Friendly Tailor /tailor/me endpoint will be for the logged-in user, across all their devices, if the user is logged in. Otherwise the stats are based on the
Ophan browser cookie as before.

The response will include the user id if it found one, to make it clearer where the stats are coming from.

This follows on from https://github.com/guardian/friendly-tailor/pull/12

cc @philwills

### DIFF
--- a/membership-attribute-service/app/controllers/FriendlyTailor.scala
+++ b/membership-attribute-service/app/controllers/FriendlyTailor.scala
@@ -2,14 +2,23 @@ package controllers
 
 import com.typesafe.scalalogging.LazyLogging
 import configuration.Config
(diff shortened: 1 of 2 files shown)


rtyley:
With https://github.com/guardian/members-data-api/pull/104 and https://github.com/guardian/members-data-api/commit/595b3a4d5 I've finally verified that this works on PROD.

https://members-data-api.theguardian.com/tailor/me?tag=politics/eu-referendum

image

...the stat updates as I surf other eu-referendum articles in my browser.

PR #104, opened 2016-05-26
Fix userId query by actually querying the Global Secondary Index

rtyley:
This is what I get for not writing any tests, and also _not_ waiting for @philwills to review https://github.com/guardian/members-data-api/pull/103


philwills:
👍

PR #106, opened 2016-06-01
Return status code in error json

mario-galic:
@tomverran 

Apps team would like to switch on the actual status code instead of on string.


nlindblad:
👍

PR #109, opened 2016-06-09
Public tier API

tomverran:
Ability to say you're happy for your membership tier to be public and a corresponding endpoint to retrieve tiers for those who allow them to be shown

This PR also refactors DynamoAttributeService to use Scanamo

Also this removes ScalaTest which breaks running tests locally as SBT tries to start two dynamoDBs and then later breaks when trying to stop one of them


paulbrown1982:
👍

paulbrown1982:
👍

PR #113, opened 2016-06-14
Stripe web hook to SNS infrastructure

jranks123:
This pull request adds a new webhook from Stripe that receives a charge event from Stripe and sends it to an SNS topic. 

Also adds a debug start script.

### DIFF
--- a/cloudformation/membership-attribute-service.json
+++ b/cloudformation/membership-attribute-service.json
@@ -424,16 +424,14 @@
   "TopicGiraffe": {
     "Type": "AWS::SNS::Topic",
     "Properties": {
-      "DisplayName": "Giraffe",
-      "TopicName": "Giraffe"
+      "TopicName": "giraffe-prod"
     }
   },
 
   "TopicGiraffeUAT": {
     "Type": "AWS::SNS::Topic",
     "Properties": {
-      "DisplayName": "GiraffeUAT",
-      "TopicName": "GiraffeUAT"
+      "TopicName": "giraffe-pre-prod"
       }
     }
 }
--- a/cloudformation/membership-attribute-service.json
+++ b/cloudformation/membership-attribute-service.json
@@ -177,6 +177,16 @@
                 "Resource": "arn:aws:dynamodb:*:*:table/friendly-tailor-PROD"
               },
               {
+                "Effect":"Allow",
+                "Action":["sns:Publish"],
+                "Resource":{"Ref":"TopicGiraffe"}
+              },
+              {
+                "Effect":"Allow",
+                "Action":["sns:Publish"],
+                "Resource":{"Ref":"TopicGiraffePreProd"}
+              },
+              {
                 "Effect": "Allow",
                 "Action": [
                   "dynamodb:Query"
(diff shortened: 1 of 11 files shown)


cloudformation/membership-attribute-service.json
```diff
@@ -177,6 +177,16 @@
...
               },
               {
+                "Effect":"Allow",
+                "Action":["sns:Publish"],
+                "Resource":{"Ref":"TopicGiraffe"}
+              },
+              {
+                "Effect":"Allow",
+                "Action":["sns:Publish"],
+                "Resource":{"Ref":"TopicGiraffeUAT"}
```
philwills: It would be nice if this was parameterised so that only a production instance could talk to the production topic.

membership-attribute-service/app/controllers/StripeHookController.scala
```diff
@@ -1,35 +1,57 @@
...
+  }
+
+
+  def publishToSns = Action.async { implicit request =>
+    request.body.asJson.map(Json.fromJson[Event[StripeObject]](_)).fold[Future[Result]] {
+      Future.successful(BadRequest("No JSON found in request body"))
+    } { event =>
+      (for {
+        e <- OptionT(Future.successful(event.asOpt))
+        tp = if (e.liveMode) NormalTouchpointComponents else TestTouchpointComponents
```
philwills: It feels odd to switch on the event, rather than have live events go to a different instance from test ones, but I don't know enough about Stripe to know whether this is a necessary evil.

membership-attribute-service/app/controllers/StripeHookController.scala
```diff
@@ -1,35 +1,57 @@
...
+        eventFromStripe <- OptionT(tp.giraffeStripeService.Event.findCharge(e.id))
+      } yield {
+        tp.snsGiraffeService.publish(eventFromStripe.`object`)
+      }).run
+        .map(_.fold[Result](Ok(Json.obj("event found" -> false)))
```
philwills: I'm more familiar with the Cats OptionT, than the scalaz one, but I'd tend to yield a result, then have a get or else for the not found case, rather than having the map/fold.

philwills:
Aside from my confusion over how test events work, this looks good to me.

membership-attribute-service/app/controllers/StripeHookController.scala
jranks123: @philwills  When you get a moment lets have a conversation about this, I'd be interested to discuss the tradeoffs

tomverran: Ah we don't have a code environment on members-data-api

philwills: OK, I got confused by the 'pre-prod' SNS topic.

cloudformation/membership-attribute-service.json
jranks123: @markjamesbutler could you please have a look at this

markjamesbutler: @philwills Agreed.  Unfortunately due to the membership architecture, the production instance needs to populate the UAT Topic for test users otherwise we pollute the production data with test users.

markjamesbutler:
@jranks123 Good work here.  Let's talk about encryption before this is merged.

PR #115, opened 2016-06-15
Add sns path for prod and dev

jranks123:
As we are going to route all charges through the same stripe -> sns -> lambda pipeline, the sns path should be the same for live and test


AWare:
👍

PR #116, opened 2016-06-16
Update membership common dependency and change map then fold to getOrElse in StripeHookController

jranks123:
- Updates membership common dependency. 
- Changes map then fold to getOrElse
- Add stripe event and aws arn to request response when calling stripe-sns webhook


AWare:
👍

PR #118, opened 2016-06-17
Updated health check parameters and enabled connection draining

paulbrown1982:
Updated health check parameters and enabled connection draining - boxes were running out of CPU credits and traffic was not failing over.

Sorry about the "Effect": "Allow" reordering...

cc @mario-galic


mario-galic:
👍

PR #119, opened 2016-06-20
Add sns publish permissions for giraffe to membership IAM role

jranks123:
Add sns publish permissions for giraffe to membership IAM role.
Create new SNS topic in cloudformation for giraffe prod
Configure prod touchpoint backend to point to new SNS topic

This sns topic is used to send Stripe charge events to a lambda


paulbrown1982:
👍

PR #120, opened 2016-06-21
Update membership common dependency to 2.19

jranks123:
This update to MC adds the livemode to the Stripe Charge object


markjamesbutler:
👍

PR #123, opened 2016-07-12
Remove lookup metrics

mario-galic:
@paulbrown1982 

[Trello card](https://trello.com/c/YIhjG1c8/19-optimise-cloudwatch-monitoring-costs)

**Save over £1000 per month.**

The culprit for our high CloudWatch cost were the following 4 metrics:

* `membership-user-not-found`
* `features-user-not-found`
* `membership-lookup-successful`
* `features-lookup-successful`

As the graph below shows these metrics were posted to cloudwatch around **five million** times per day, which at pricing of $0.01 per 1,000 `PutMetricData` API requests, adds to around $1500 per month:

image

Beside being expensive, these metrics do not add much value. For instance, the fact that user is not in DynamoDB does not represent an error, it simply means the user has not become member yet.


paulbrown1982:
👍

PR #125, opened 2016-07-13
Fix Salesforce sync

mario-galic:
@paulbrown1982 @tomverran 

[Trello ticket](https://trello.com/c/PquWKnxV/42-membership-dynamo-db-out-of-sync)

The issue was that we were returning OK to Salesforce in all cases, even if insertion into DynamoDB failed. Call to DynamoDB was executed in separate Future, after which we would immediately return OK without checking the result of the Future: [SalesforceHookController.scala: L54](https://github.com/guardian/members-data-api/compare/fix-salesforce-sync?expand=1#diff-69104f1fecf890597795bd8d7aadf135L54)


tomverran:
👍

PR #126, opened 2016-07-14
Add converted amount and country to Stripe object before sending to SNS

jranks123:
We needed to have the converted amount and the card country in the stripe event for data analysis purposes. 

As the converted amount is not in the Stripe Charge object that comes from the webhook, we need to make a call to the Stripe API to get the Balance Transaction object, which we then send to SNS along with the Charge object


jranks123:
This will fail the tests until this PR goes live: https://github.com/guardian/membership-common/pull/274

tomverran:
👍

PR #127, opened 2016-08-01
Use membership common 248

tomverran:
This changes prices from being strange float pence values to being ints of pence

image

image

The frontend JS can handle this okay

image


paulbrown1982:
👍 👍 👍

PR #131, opened 2016-10-05
remove contribution specific code

alexduf:
I'm not familiar with this project so please be extra careful when reviewing this PR.

I have removed everything that was making sense as being a direct dependency of the stripe web hook.

I have compiled and tested the branch but haven't tried to run the project as the setup is non trivial.

let me know if I missed anything


paulbrown1982:
Looks good, however we've got a big change (to delete a load of subs stuff) which we'd like to go ahead of this which may cause conflicts with this checking, which we're happy to resolve.

johnduffell:
is this still going anywhere or should be closed?

paulbrown1982:
cc @alexduf ?

alexduf:
I think this should be merged, contribution's webhook doesn't go through this service, so this is essentially dead code

johnduffell:
this is winning the prize for oldest PR surely!  pretty heavily conflicted too.

johnduffell:
looks pretty uncontroversial to me, I'm not sure anyone knows the code that well! 👍 but needs the conflicts sorting before it can go. Also @pvighi is shipping some changes in the near future so it may be worth coordinating

johnduffell:
🔪

PR #135, opened 2016-10-25
update membership common with the deleted old code

johnduffell:
So this pulls in the membership common PR with the deleted code - https://github.com/guardian/membership-common/pull/331

I ran the tests, which passed, but when I run locally, it's using a dataset that doesn't give me the directly testable endpoints.  Is it enough to run the tests, or should I be running it up myself pointing somewhere?

@paulbrown1982 @guardian/membership-and-subscriptions any comments or +1s?


paulbrown1982:
If you spin it up locally and test with a cookied session, does calling /user-attributes/me/features return a full complement of data?

johnduffell:
no I just tried and I just get a 401 error and nothing in the logs.  I definitely sent over a cookie

johnduffell:
ok I managed to test locally and got supporter data back correctly

PR #138, opened 2016-10-26
Grab the right aws creds, end dev headaches

AWare:
At the moment, we just grab the standard aws creds- now this uses membership common to check if this is appropriate or whether it should grab them out of the membership profile.


AWare:
@johnduffell

nginx/members-data-api.conf
```diff
@@ -3,7 +3,7 @@ server {
 
     location / {
         proxy_pass http://localhost:9400/;
-        proxy_set_header Host $http_host;
+          proxy_set_header Host $http_host;
```
johnduffell: probably not intentional?

johnduffell:
great, a bit of an odd selection, but good work! 👍 🚤

PR #139, opened 2016-11-08
Fix features response when user not found

paulbrown1982:
Revert the /features controller to work as it did pre-June 1st: serving its own custom error response, and not falling back to the  default which /membership uses.

This means we don't need to do: https://github.com/guardian/frontend/pull/14920

cc @philwills @mario-galic @rtyley @jacobwinch

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -50,13 +50,15 @@ class AttributeController extends Controller with LazyLogging {
     ))
   }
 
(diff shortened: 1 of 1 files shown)


mario-galic:
Is 404 designed for this kind of situation? Perhaps we should return 200 with JSON describing the result of the query.

rtyley:
> Is 404 designed for this kind of situation? Perhaps we should return 200 with JSON describing the result of the query.

Similar comment from me :) https://github.com/guardian/frontend/pull/14920#issuecomment-259101700

paulbrown1982:
Yup, that's what /features is doing now -returning an appropriate JSON describing the result of the query. 

I agree /membership should do it too, but we need to check what the apps / other clients are expecting.

PR #140, opened 2016-11-15
Fix Salesforce Parser in order to tolerate multiple objects per message

jacobwinch:
This PR allows our Salesforce Parser to handle multiple objects/notifications sent within a single message.

Previously the assumption seems to have been that each message would only include one sfObject, however the Salesforce docs show that multiple sfObjects can be included in a single message:
https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_om_outboundmessaging_notifications.htm
https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_om_outboundmessaging_wsdl.htm

The important parts are:

_A single SOAP message can include up to 100 notifications. Each notification contains the object ID and a reference to the associated sObject data. Note that if the information in the object changes after the notification is queued but before it is sent, only the updated information will be delivered.
...
If you issue multiple discrete calls, the calls may be batched together into one or more SOAP messages._

And later:

_This element is the schema for sending an acknowledgement (ack) response to Salesforce.
...
You acknowledge all notifications in the message if there is more than one._

We have seen huge numbers of members missing from Dynamo DB in the recent period of heavy traffic, and I suspect that this may be why. The logging included in this PR will also allow us to know for sure whether we are getting multiple sfObjects per message.

@paulbrown1982 @mario-galic @johnduffell

### DIFF
--- a/membership-attribute-service/app/controllers/SalesforceHookController.scala
+++ b/membership-attribute-service/app/controllers/SalesforceHookController.scala
@@ -46,19 +46,19 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
   val attributeService = touchpoint.attrService
   implicit val pf = Membership
 
-  def processDeletion(membershipDeletion: MembershipDeletion) = {
+  def deleteMemberRecord(membershipDeletion: MembershipDeletion): Future[Object] = {
     val userId = membershipDeletion.userId
-    attributeService.delete(userId).map { _ =>
+    attributeService.delete(userId).map { deleteItemResult =>
       logger.info(s"Successfully deleted user $userId from ${touchpoint.dynamoTable}.")
       metrics.put("Delete", 1)
-      Success
+      deleteItemResult
     }.recover { case e: Throwable =>
       logger.warn(s"Failed to delete user $userId from ${touchpoint.dynamoTable}. Salesforce should retry.", e)
       Failure
     }
   }
 
-  def processUpdate(membershipUpdate: MembershipUpdate) = {
+  def updateMemberRecord(membershipUpdate: MembershipUpdate): Future[Object] = {
     val attrs = membershipUpdate.attributes
     (for {
       sfId <- OptionT(touchpoint.contactRepo.get(attrs.UserId))
@@ -74,7 +74,7 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
       attributeService.set(attrsUpdatedWithZuoraOpt.getOrElse(attrs)).map { putItemResult =>
         logger.info(s"Successfully inserted ${attrsUpdatedWithZuoraOpt.getOrElse(attrs)} into ${touchpoint.dynamoTable}.")
         metrics.put("Update", 1)
-        Success
+        putItemResult
       }.recover { case e: Throwable =>
         logger.warn(s"Failed to insert ${attrsUpdatedWithZuoraOpt.getOrElse(attrs)} into ${touchpoint.dynamoTable}. Salesforce should retry.", e)
         Failure
@@ -94,11 +94,13 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
       logger.info(s"Parsed Salesforce message successfully. Salesforce sent ${outboundMessageChanges.length} objects to update: $outboundMessageChanges")
       // Take the Seq and apply the appropriate action for each notification item, based on its type
       val updates = outboundMessageChanges.map {
-        case membershipDelete: MembershipDeletion => processDeletion(membershipDelete)
-        case membershipUpdate: MembershipUpdate => processUpdate(membershipUpdate)
+        case membershipDelete: MembershipDeletion => deleteMemberRecord(membershipDelete)
+        case membershipUpdate: MembershipUpdate => updateMemberRecord(membershipUpdate)
       }
       // Gather up the results of the futures and check for failures. Only send a success response to Salesforce if every processUpdate/processDeletion for the message succeeds
-      Future.sequence(updates).map { updateSeq => if (updateSeq.contains(Failure)) ApiErrors.internalError else ack }
+      Future.sequence(updates).map { updateSeq =>
+        if (updateSeq.contains(Failure)) ApiErrors.internalError else ack
+      }
   }
  }
 }
--- a/membership-attribute-service/app/controllers/SalesforceHookController.scala
+++ b/membership-attribute-service/app/controllers/SalesforceHookController.scala
@@ -48,7 +48,7 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
 
   def processDeletion(membershipDeletion: MembershipDeletion) = {
     val userId = membershipDeletion.userId
-    attributeService.delete(userId).map { x =>
+    attributeService.delete(userId).map { _ =>
       logger.info(s"Successfully deleted user $userId from ${touchpoint.dynamoTable}.")
       metrics.put("Delete", 1)
       Success
@@ -67,6 +67,7 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
       val tierFromSalesforceWebhook = attrs.Tier
       val tierFromZuora = membershipSubscription.plan.charges.benefit.id
       if (tierFromZuora != tierFromSalesforceWebhook) logger.error(s"Differing tier info for $sfId : sf=$tierFromSalesforceWebhook zuora=$tierFromZuora")
+      // If the tier info does not match, we trust the info we get from Zuora, instead of the tier sent to us in the outbound message from Salesforce
       attrs.copy(Tier = tierFromZuora)
     }).run.flatMap { attrsUpdatedWithZuoraOpt =>
       if (attrsUpdatedWithZuoraOpt.isEmpty) logger.error(s"Couldn't update $attrs with information from Zuora")
@@ -92,10 +93,10 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
     case \/-(outboundMessageChanges @ Seq(_*)) =>
       logger.info(s"Parsed Salesforce message successfully. Salesforce sent ${outboundMessageChanges.length} objects to update: $outboundMessageChanges")
       // Take the Seq and apply the appropriate action for each notification item, based on its type
-      val updates = outboundMessageChanges.map( outboundMessage => outboundMessage match {
+      val updates = outboundMessageChanges.map {
         case membershipDelete: MembershipDeletion => processDeletion(membershipDelete)
         case membershipUpdate: MembershipUpdate => processUpdate(membershipUpdate)
-      })
+      }
       // Gather up the results of the futures and check for failures. Only send a success response to Salesforce if every processUpdate/processDeletion for the message succeeds
       Future.sequence(updates).map { updateSeq => if (updateSeq.contains(Failure)) ApiErrors.internalError else ack }
   }
--- a/membership-attribute-service/app/controllers/SalesforceHookController.scala
+++ b/membership-attribute-service/app/controllers/SalesforceHookController.scala
@@ -6,15 +6,15 @@ import com.gu.memsub.subsv2.SubscriptionPlan
 import com.typesafe.scalalogging.LazyLogging
 import models.ApiErrors
 import monitoring.CloudWatch
-import parsers.Salesforce.{MembershipDeletion, MembershipUpdate, OrgIdMatchingError, ParsingError}
+import parsers.Salesforce._
 import parsers.{Salesforce => SFParser}
-import play.Logger
 import play.api.libs.concurrent.Execution.Implicits._
 import play.api.mvc.BodyParsers.parse
 import play.api.mvc.Results.Ok
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.memsub.subsv2.reads.ChargeListReads._
 import scala.concurrent.Future
+import scala.util.{Failure, Success}
 import scalaz.std.scalaFuture._
 import scalaz.{-\/, OptionT, \/-}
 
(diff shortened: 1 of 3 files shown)


jacobwinch:
In addition to the tests I added to the code, I ran the app in DEV mode and hit the /salesforce-hook endpoint with various SOAP requests in order to validate that we were still correctly updating Dynamo DB after making these changes.

membership-attribute-service/app/parsers/Salesforce.scala
```diff
@@ -1,48 +1,73 @@
...
-      id <- obj.getText("IdentityID__c")
-    } yield {
-      val tier = (obj \ "Membership_Tier__c").map(_.text).headOption
-      tier.fold[OutboundMessageChange](MembershipDeletion(id)) { t =>
-        val num = (obj \ "Membership_Number__c").headOption.map(_.text)
-        MembershipUpdate(Attributes(id, t, num))
+    def orgIdValidation: OutboundMessageParseError \/ Boolean = {
+      for {
+        orgId <- (payload \\ "notifications").getText("OrganizationId")
+        isValid <- if (orgId === organizationId) true.right else OrgIdMatchingError(orgId).left
```
johnduffell: you never use false here so unit was enough

membership-attribute-service/app/parsers/Salesforce.scala
```diff
@@ -1,48 +1,73 @@
...
-      orgId <- (payload \\ "notifications").getText("OrganizationId")
-      _ <- if (orgId === organizationId) ().right else OrgIdMatchingError(orgId).left
-      obj <- (payload \\ "Notification").getTag("sObject")
-      id <- obj.getText("IdentityID__c")
-    } yield {
-      val tier = (obj \ "Membership_Tier__c").map(_.text).headOption
-      tier.fold[OutboundMessageChange](MembershipDeletion(id)) { t =>
-        val num = (obj \ "Membership_Number__c").headOption.map(_.text)
-        MembershipUpdate(Attributes(id, t, num))
+    def orgIdValidation: OutboundMessageParseError \/ Boolean = {
```
johnduffell: I'd probably called it orgIdValid (although it is the validation code, the function is evaluating to whether it's valid)  you can prefix with "is" although I heard it's more java style, but  I still do it sometimes

membership-attribute-service/app/parsers/Salesforce.scala
```diff
@@ -1,48 +1,73 @@
...
+        // If the Salesforce Contact has a Tier, we mark them for an update
+        case Some(tier) => MembershipUpdate(Attributes(id, tier, num))
       }
     }
+
+    orgIdValidation match {
+      case -\/(ParsingError(msg)) => ParsingError(msg).left
+      case -\/(OrgIdMatchingError(orgId)) => OrgIdMatchingError(orgId).left
+      // If we can validate the organization ID, then we get a Right and can attempt to process the notifications.
+      case \/-(_) => getSalesforceObjects.map { _.map(sfObject => processSalesforceObject(sfObject)) }
```
johnduffell: not sure what the general wisdom is, but I feel like we could just map the orgidvalidation with getSalesforceObjects, and then flatMap that with the bit you put in braces there, rather than breaking it out with a pattern match and then putting it back together

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -39,48 +39,65 @@ class SalesforceHookController extends LazyLogging {
...
+      logger.error(s"Could not parse payload. \n$msg")
+      Future(ApiErrors.badRequest(msg))
+    case -\/(OrgIdMatchingError(orgId)) =>
+      logger.error(s"Wrong organization Id: $orgId")
+      Future(ApiErrors.unauthorized.copy("Wrong organization Id"))
+    // On successful parse, we should end up with a Seq of Salesforce objects to update
+    case \/-(outboundMessageChanges @ Seq(_*)) =>
+      logger.info(s"Parsed Salesforce message successfully. Salesforce sent ${outboundMessageChanges.length} objects to update: $outboundMessageChanges")
+      // Take the Seq and apply the appropriate action for each notification item, based on its type
+      val updates = outboundMessageChanges.map( outboundMessage => outboundMessage match {
```
johnduffell: you can just miss out the `( outboundMessage => outboundMessage match ` and it automatically assimes you're effectively pattern matching

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -39,48 +39,65 @@ class SalesforceHookController extends LazyLogging {
...
-        Logger.error(s"Wrong organization Id: $orgId")
-        Future(ApiErrors.unauthorized.copy("Wrong organization Id"))
+  val touchpoint = request.touchpoint
+  val validOrgId = touchpoint.sfOrganisationId
+  val attributeService = touchpoint.attrService
+  implicit val pf = Membership
+
+  def processDeletion(membershipDeletion: MembershipDeletion) = {
+    val userId = membershipDeletion.userId
+    attributeService.delete(userId).map { x =>
```
johnduffell: if `x` is unused, you can use `_` to make that clear

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -39,48 +39,65 @@ class SalesforceHookController extends LazyLogging {
...
+    (for {
+      sfId <- OptionT(touchpoint.contactRepo.get(attrs.UserId))
+      membershipSubscription <- OptionT(touchpoint.subService.current[SubscriptionPlan.Member](sfId).map(_.headOption))
+    } yield {
+      val tierFromSalesforceWebhook = attrs.Tier
+      val tierFromZuora = membershipSubscription.plan.charges.benefit.id
+      if (tierFromZuora != tierFromSalesforceWebhook) logger.error(s"Differing tier info for $sfId : sf=$tierFromSalesforceWebhook zuora=$tierFromZuora")
+      attrs.copy(Tier = tierFromZuora)
+    }).run.flatMap { attrsUpdatedWithZuoraOpt =>
+      if (attrsUpdatedWithZuoraOpt.isEmpty) logger.error(s"Couldn't update $attrs with information from Zuora")
```
johnduffell: is this a problem?  I can see this is existing code!

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -39,48 +39,65 @@ class SalesforceHookController extends LazyLogging {
...
+      attributeService.set(attrsUpdatedWithZuoraOpt.getOrElse(attrs)).map { putItemResult =>
+        logger.info(s"Successfully inserted ${attrsUpdatedWithZuoraOpt.getOrElse(attrs)} into ${touchpoint.dynamoTable}.")
+        metrics.put("Update", 1)
+        Success
+      }.recover { case e: Throwable =>
```
johnduffell: in an ideal world we would wrap anything that might throw an exception with a Try and turn the result into a `\/` straight away, to avoid having exception handling code dotted around

membership-attribute-service/test/parsers/SalesforceTest.scala
```diff
@@ -110,5 +191,25 @@ class SalesforceTest extends Specification {
...
+              <EnterpriseUrl>https://cs17.salesforce.com/services/Soap/c/34.0/enterprise_id</EnterpriseUrl>
+              <PartnerUrl>https://cs17.salesforce.com/services/Soap/u/34.0/enterprise_id</PartnerUrl>
+              <Notification>
+                <Id>notification_id</Id>
+              </Notification>
+            </notifications>
+          </soapenv:Body>
+        </soapenv:Envelope>
+
+      Salesforce.parseOutboundMessage(payload, orgId) shouldEqual ParsingError(s"No Salesforce Objects were found when parsing the message. \n $payload").left
```
jacobwinch: Yes, that's possible. We don't respond to Salesforce until after we've actually actioned all of the updates (in SalesforceHookController.scala) and as there doesn't seem to be any way of responding with granularity about what did/didn't succeed, we just return an API error if there are any failures during the update sequence.

membership-attribute-service/app/controllers/SalesforceHookController.scala
jacobwinch: Nice - thanks for the tip

membership-attribute-service/app/controllers/SalesforceHookController.scala
jacobwinch: It looks like we mistrust the data that we get sent via the Salesforce webhook, and prefer to calculate the correct tier for ourselves (if possible). I have put a comment in to this effect for future reference, but I won't remove this code until we work out whether this mistrust is justified...

membership-attribute-service/app/parsers/Salesforce.scala
jacobwinch: I tidied up the pattern matching as per your suggestion above, which makes that solution a bit easier on the eye!

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -39,48 +39,66 @@ class SalesforceHookController extends LazyLogging {
...
-        Logger.error(s"Wrong organization Id: $orgId")
-        Future(ApiErrors.unauthorized.copy("Wrong organization Id"))
+  val touchpoint = request.touchpoint
+  val validOrgId = touchpoint.sfOrganisationId
+  val attributeService = touchpoint.attrService
+  implicit val pf = Membership
+
+  def processDeletion(membershipDeletion: MembershipDeletion) = {
+    val userId = membershipDeletion.userId
+    attributeService.delete(userId).map { _ =>
```
johnduffell: what does delete return? is it just unit?

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -39,48 +39,66 @@ class SalesforceHookController extends LazyLogging {
...
+    // On successful parse, we should end up with a Seq of Salesforce objects to update
+    case \/-(outboundMessageChanges @ Seq(_*)) =>
+      logger.info(s"Parsed Salesforce message successfully. Salesforce sent ${outboundMessageChanges.length} objects to update: $outboundMessageChanges")
+      // Take the Seq and apply the appropriate action for each notification item, based on its type
+      val updates = outboundMessageChanges.map {
+        case membershipDelete: MembershipDeletion => processDeletion(membershipDelete)
+        case membershipUpdate: MembershipUpdate => processUpdate(membershipUpdate)
+      }
+      // Gather up the results of the futures and check for failures. Only send a success response to Salesforce if every processUpdate/processDeletion for the message succeeds
+      Future.sequence(updates).map { updateSeq => if (updateSeq.contains(Failure)) ApiErrors.internalError else ack }
```
johnduffell: just split the if onto a new line

membership-attribute-service/app/parsers/Salesforce.scala
```diff
@@ -1,48 +1,70 @@
...
-      _ <- if (orgId === organizationId) ().right else OrgIdMatchingError(orgId).left
-      obj <- (payload \\ "Notification").getTag("sObject")
-      id <- obj.getText("IdentityID__c")
-    } yield {
-      val tier = (obj \ "Membership_Tier__c").map(_.text).headOption
-      tier.fold[OutboundMessageChange](MembershipDeletion(id)) { t =>
-        val num = (obj \ "Membership_Number__c").headOption.map(_.text)
-        MembershipUpdate(Attributes(id, t, num))
+    def orgIdValid: OutboundMessageParseError \/ Unit = {
+      (payload \\ "notifications").getText("OrganizationId").flatMap(orgId => if (orgId != organizationId) OrgIdMatchingError(orgId).left else ().right)
```
johnduffell: use braces instead of () in the flatmap and split the if onto a separate line

membership-attribute-service/app/parsers/Salesforce.scala
```diff
@@ -1,48 +1,70 @@
...
+
+    def getSalesforceObjects: OutboundMessageParseError \/ NodeSeq = {
+        val salesforceObjects = (payload \\ "sObject")
+        if (salesforceObjects.isEmpty) {
+          ParsingError(s"No Salesforce Objects were found when parsing the message. \n $payload").left
+        }
+        else salesforceObjects.right
+    }
+
+    def processSalesforceObject(salesforceObject: Node): OutboundMessageChange = {
```
johnduffell: "process" is a word I generally avoid, because all methods are processing something.  I suppose this is reading the membership and identity information from a salesforce object

membership-attribute-service/app/parsers/Salesforce.scala
```diff
@@ -1,48 +1,70 @@
...
+      // Match on the Tier to determine the required action
+      tier match {
+        // If Salesforce Contact object has no Tier, we assume the user has expired/cancelled and mark them for deletion
+        case None => MembershipDeletion(id)
+        // If the Salesforce Contact has a Tier, we mark them for an update
+        case Some(tier) => MembershipUpdate(Attributes(id, tier, num))
       }
     }
+
+    orgIdValid match {
```
johnduffell: this whole match statement is still just doing a flatMap explicitly
```scala
orgIdValid.flatMap { _ =>
  getSalesforceObjects.map { _.map(sfObject => processSalesforceObject(sfObject)) }
}
```

johnduffell:
great, good work, let's see if it works! 👍 🚀

jacobwinch:
From the logs post-merge:

_INFO: Parsed Salesforce message successfully. Salesforce sent 2 objects to update: List(MembershipUpdate(Attributes(17018308,Supporter,Some(688075),None)), MembershipUpdate(Attributes(17018298,Supporter,Some(688076),None)))_
...
_INFO: Successfully inserted Attributes(17018298,Supporter,Some(688076),None) into MembershipAttributes-PROD._
...
_INFO: Successfully inserted Attributes(17018308,Supporter,Some(688075),None) into MembershipAttributes-PROD._

PR #142, opened 2016-12-02
update membership common version

pvighi:
this is needed to keep this from failing when weekly one off plans are added to the zuora catalog


jacobwinch:
👍

PR #145, opened 2016-12-14
Load user ad-free status from database

JustinPinner:
In preparation for the ad-free trial, we now read the user's ad-free participation state from Dynamo DB.

We can manually update the table for specific users while we're preparing the backend services that will switch people in or out of the trial. This will work in conjunction with dotcom frontend changes to enable the trial.

### DIFF
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -15,13 +15,14 @@ object ContentAccess {
   implicit val jsWrite = Json.writes[ContentAccess]
 }
 
-case class Attributes(UserId: String, Tier: String, MembershipNumber: Option[String], PublicTier: Option[Boolean] = None) {
+case class Attributes(UserId: String, Tier: String, MembershipNumber: Option[String], PublicTier: Option[Boolean] = None, AdFree: Option[Boolean] = None) {
   require(Tier.nonEmpty)
   require(UserId.nonEmpty)
 
   lazy val allowsPublicTierDisplay = PublicTier.exists(identity)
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/models/Attributes.scala
```diff
@@ -15,13 +15,14 @@ object ContentAccess {
...
 
-case class Attributes(UserId: String, Tier: String, MembershipNumber: Option[String], PublicTier: Option[Boolean] = None) {
+case class Attributes(UserId: String, Tier: String, MembershipNumber: Option[String], PublicTier: Option[Boolean] = None, AdFree: Option[Boolean] = None) {
   require(Tier.nonEmpty)
   require(UserId.nonEmpty)
 
   lazy val allowsPublicTierDisplay = PublicTier.exists(identity)
   lazy val isFriendTier = Tier.equalsIgnoreCase("friend")
   lazy val isPaidTier = !isFriendTier
+  lazy val isAdFree = AdFree.exists(identity)
```
paulbrown1982: Is 'identity' an imported function that can be used to determine if an Option[Boolean] contains a true boolean value?

JustinPinner: The reference in the code is to a scala predef of an identity function that, apparently, always returns the same value as the passed argument. It's used to catch the `PublicTier` value, which is also an `Option[Boolean]`, so I figured it'd be ok for `AdFree` too 🤕

JustinPinner:
Re-deployed manually as it looked like boxes were in the process of restarting which caused the automatic deploy to bail.

PR #146, opened 2016-12-15
Allow http code env requests...

JustinPinner:
Prevent these errors in the code environment;

```No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://m.code.dev-theguardian.com' is therefore not allowed access. The response had HTTP status code 403.```


johnduffell:
good spot! 👍 🚀

PR #150, opened 2017-02-03
Fix DynamoDB Local download by upgrading sbt-dynamodb to v1.5.3

rtyley:
Just like https://github.com/guardian/membership-common/pull/420, to get the fix in https://github.com/localytics/sbt-dynamodb/pull/30 for AWS shifting their download url.

cc @rupertbates


rupertbates:
👍

PR #152, opened 2017-02-03
Version updates and tweaks

rtyley:
I'm bunching these disparate updates into one PR, the commit messages should describe them fairly well.

cc @svillafe


rtyley:
Using MMA as a barometer, it appears to be working just fine.

https://profile.theguardian.com/membership/edit

PR #155, opened 2017-02-07
Track incomplete checkouts take 2

JustinPinner:
Capture the pre-validated data from sent from membership frontend and send that to the database instead of attempting to extract the data from a re-validated call to the identity service, which worked when running locally, but not when deployed to PROD.

And since the data is sent in the POST request for both capturing and removing behaviour events, we don't have to pass the activity in the URL, so those are gone too.

There are some log captures in the corresponding membership-frontend PR here: https://github.com/guardian/membership-frontend/pull/1487


JustinPinner:
Thanks @svillafe - I'll merge now and try testing it from my local membership-frontend.

PR #157, opened 2017-02-08
Add Guardian dentity headers

rtyley:
@alfavata on the Mobile team had seen a recent spike in new errors coming from the members-data-api - a total of 17K errors starting Feb 1st, with a couple of spikes on Feb 2nd (3pm and 10pm). This is in addition to the 'regular background noise' of empty-response errors that the iOS app says it gets from the members-data-api.

After trying to investigate this, I feel that better logging on the members-data-api is required to understand what's going on : specifically, we need to turn on Fastly logging (it's not currently turned on for the members-data-api), and we need to add Guardian Identity information to the response headers so that it can be recorded in the Fastly logs. At the very least, this should allow us to look at a specific example of an error and say - _"this is how Fastly says we responded to that request"_.

The "Ensure cache headers are set on all HTTP responses" commit was added because in the past we've only returned Guardian Identity

Unfortunately, although we normally just add the relevant headers with the `AddGuIdentityHeaders` filter, these response headers are thrown away for `onBadRequest`, `onHandlerNotFound` and `onError` (eg a HTTP 500 internal server error response). Because we're particularly concerned about these, I've added the Identity response headers into the error handlers in `Global` as well.

https://trello.com/c/i3BzCE6i/320-add-fastly-logging-to-members-data-api-to-help-diagnose-errors

cc @paulbrown1982

### DIFF
--- a/membership-attribute-service/app/Global.scala
+++ b/membership-attribute-service/app/Global.scala
@@ -1,13 +1,19 @@
-import filters.{AddEC2InstanceHeader, LogRequestsFilter}
+import filters.AddEC2InstanceHeader
 import models.ApiErrors._
 import monitoring.SentryLogging
-import play.api.{Application, Logger}
 import play.api.libs.concurrent.Execution.Implicits._
 import play.api.mvc.{RequestHeader, Result, WithFilters}
+import play.api.{Application, Logger}
 import play.filters.csrf._
+
 import scala.concurrent.Future
 
-object Global extends WithFilters(CSRFFilter(), AddEC2InstanceHeader) {
+
+object Global extends WithFilters(
+  CheckCacheHeadersFilter,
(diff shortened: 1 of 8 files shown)


alfavata:
Cheers! We're adding some more logging on our side too, anyway.

rtyley:
[Add Fastly logging to members-data-api to help diagnose errors](https://trello.com/c/i3BzCE6i/320-add-fastly-logging-to-members-data-api-to-help-diagnose-errors)

rtyley:
Related Fastly changes:

https://manage.fastly.com/configure/services/3oeL55htke1szmt67tNxEq/versions/11/logging
https://manage.fastly.com/configure/services/3oeL55htke1szmt67tNxEq/diff/8,12

PR #160, opened 2017-02-08
Fix 'Cache-Control not set' error on robots.txt and other pages

rtyley:
Unfortunately the `CheckCacheHeadersFilter` added with https://github.com/guardian/members-data-api/pull/157/commits/0bcf6afca233862d37a5e55d8cb5487de3950c90 ...was throwing a lot of noisy errors because our error pages didn't have cache headers set :) Those are triggered on 404s (eg robots.txt) and other pages.

https://sentry.io/the-guardian/members-data-api/issues/215640173/

...also it turns out that the `AddGuIdentityHeaders.headersFor()` wrapping is only necessary for `onError()`, as the `AddGuIdentityHeaders` filter-added response-headers are not wiped out for `onHandlerNotFound()` and `onBadRequest()`

cc @svillafe @paulbrown1982

### DIFF
--- a/membership-attribute-service/app/Global.scala
+++ b/membership-attribute-service/app/Global.scala
@@ -25,12 +25,12 @@ object Global extends WithFilters(
 
   override def onBadRequest(request: RequestHeader, error: String): Future[Result] = {
     logger.debug(s"Bad request: $request, error: $error")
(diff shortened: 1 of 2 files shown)


rtyley:
The 404 errors are now being returned correctly:

image

PR #163, opened 2017-02-22
Track incomplete checkouts with email

JustinPinner:
Change abandoned cart DynamoDb `add` to be an `upsert` operation instead. This makes it far easier and cleaner to send updates to the database.

Also, this PR adds an `abandonded-cart/email` route for the abandonedCartWatcher lambda to call back to when it encounters old-enough records. Currently this is a logging feature only while we test it and prepare an actual email generation pipeline for it. The data received from the lambda is used to check the Identity record for the user id before any email is sent, and doing this means we don't have to store the email address anywhere other than in Identity.

There is a corresponding PR in membership-frontend (https://github.com/guardian/membership-frontend/pull/1518)

### DIFF
--- a/membership-attribute-service/app/controllers/BehaviourController.scala
+++ b/membership-attribute-service/app/controllers/BehaviourController.scala
@@ -9,6 +9,7 @@ import play.api.libs.concurrent.Execution.Implicits._
 import play.api.libs.json.JsValue
 import play.api.mvc.{AnyContent, Controller}
 import play.filters.cors.CORSActionBuilder
+import services.IdentityService.IdentityId
 import services.{AuthenticationService, IdentityAuthService}
 
 class BehaviourController extends Controller with LazyLogging {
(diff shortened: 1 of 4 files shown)


membership-attribute-service/app/controllers/BehaviourController.scala
```diff
@@ -23,30 +20,63 @@ class BehaviourController extends Controller with LazyLogging {
...
+        val updateResult = for {
+          current <- request.touchpoint.behaviourService.get(receivedBehaviour.userId)
+          upserted = current.map { bhv =>
+            bhv.copy(
+              activity = receivedBehaviour.activity.orElse(bhv.activity),
+              lastObserved = receivedBehaviour.lastObserved.orElse(bhv.lastObserved),
+              note = receivedBehaviour.note.orElse(bhv.note),
+              emailed = receivedBehaviour.emailed.orElse(bhv.emailed))
+          }.getOrElse(receivedBehaviour)
+          res <- request.touchpoint.behaviourService.set(upserted)
```
rtyley: This upserting code is kind of interesting to me because it looks like a place where maybe the [DynamoDB upsert functionality](http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html) would be useful? Under the hood, we are using Scanamo here, and I think Scanamo has some kind of functionality around this: https://github.com/guardian/scanamo#updating

JustinPinner: Ok thanks. I'm sure there's scope for improving this as we progress. I'll merge this as-is so we can begin testing the overall approach - specifically the requests made to identity when prompted by the lambda.

PR #165, opened 2017-02-27
Abandoned cart SQS plumbing for membership-workflow

JustinPinner:
Rather than burden the API with the additional complications of interacting with ExactTarget, we can just make use of the existing features of membership-workflow which already does all this for us, with the additional benefit of coping with backend outages because we only interact with it through the queues.

This PR sends messages into the queue, and a corresponding one in workflow (https://github.com/guardian/membership-workflow/pull/66) processes those messages via ET.

Note: the message format here is subject to change, if only to remove the test elements when we're about to go live.

cc: @Ap0c @svillafe @rupertbates @rtyley

### DIFF
--- a/membership-attribute-service/app/controllers/BehaviourController.scala
+++ b/membership-attribute-service/app/controllers/BehaviourController.scala
@@ -1,20 +1,16 @@
 package controllers
 
 import actions._
-import com.amazonaws.regions.{Regions, Region}
-import com.amazonaws.services.sqs.AmazonSQSClient
-import com.amazonaws.services.sqs.model.{SendMessageRequest, CreateQueueRequest}
-import com.gu.aws._
 import com.typesafe.scalalogging.LazyLogging
 import configuration.Config
 import models.Behaviour
 import monitoring.Metrics
 import play.api.libs.concurrent.Execution.Implicits._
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/controllers/BehaviourController.scala
```diff
@@ -33,28 +39,59 @@ class BehaviourController extends Controller with LazyLogging {
...
+      "Message" -> completionLink,
+      "TestData" -> ignoredTestData,
+      "DataExtensionName" -> "supporter-abandoned-checkout-email"
+    )
+
+    try {
+      sqsClient.sendMessage(new SendMessageRequest(emailQueueUrl, msg.toString()))
+      true
+    } catch {
+      case e: Exception => logger.warn("email queue operation failed")
```
Ap0c: Is it worth also logging the details of the exception here? e.g.:

```
logger.warn(s"email queue operation failed: ${e}")
```

JustinPinner:
@Ap0c Sorry Jamie - any chance you could have another quick look over this? I've made a couple of changes since you last reviewed it. Many thanks.

Ap0c:
Yeah looks good!

PR #169, opened 2017-04-03
Make the api work properly with test users

rupertbates:
There were 2 separate issues which were causing problems:
* The lookup to see if a user was a test user was being done by the full name rather than first name which is what Identity expects
* In DEV the identity.test.users.secret config key was always being overwritten by the dummy value in DEV.conf because of the ordering of config imports

Trello card is [here](https://trello.com/c/p24hSehz/419-members-data-api-is-not-recognising-test-users)

NB: To get this to work locally you will need to refetch the private conf from aws, as the key was not set in the previous version.

`aws s3 cp s3://members-data-api-private/DEV/members-data-api.conf /etc/gu/ --profile membership`


jacobwinch:
👍

JustinPinner:
Nice! 🚢

PR #171, opened 2017-04-03
Convert cloudformation from json to yaml

davidfurey:
Improves readability


cloudformation/membership-attribute-service.yaml
```diff
@@ -0,0 +1,329 @@
...
+    - t2.micro
+    - t2.small
+    - t2.medium
+    - m3.medium
+    - m3.large
+    - m3.xlarge
+    - m3.2xlarge
+    - c3.large
+    - c3.xlarge
+    - c3.2xlarge
```
jacobwinch: I know it's not been introduced in this PR, but we can probably trim this list down a bit a some stage.

cloudformation/membership-attribute-service.yaml
```diff
@@ -0,0 +1,329 @@
...
+      - Ref: GuardianAccessSecurityGroup
+      - Ref: VulnerabilityScanningSecurityGroup
+      InstanceType:
+        Ref: InstanceType
+      AssociatePublicIpAddress: 'True'
+      IamInstanceProfile:
+        Ref: InstanceProfile
+      UserData:
+        "Fn::Base64":
+          !Sub
```
rtyley: :sparkling_heart:  :champagne: :fireworks: :sparkles:  [`!Sub`](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html) is definitely the cherry on the cake of switching cloudformation from JSON to YAML

davidfurey: I used json2yaml.com for most of it, and then manually tidied up the UserData to use the !Sub, which is so much easier to read.

PR #174, opened 2017-04-04
Use play's defaultContext execution context

lmath:
Using the Play's default context is considered best practice:
https://www.playframework.com/documentation/2.5.x/ThreadPools

So, this change is to use play's defaultContext executionContext rather than Scala's global execution context. 

This isn't to address a particular performance problem, just a matter of moving to best practice. 

See also: 
https://trello.com/c/7CTCozNx/58-use-play-s-default-execution-context

cc @paulbrown1982 @jacobwinch @mario-galic @johnduffell @pvighi

### DIFF
--- a/membership-attribute-service/app/BatchLoader.scala
+++ b/membership-attribute-service/app/BatchLoader.scala
@@ -11,7 +11,7 @@ import repositories.MembershipAttributesSerializer
 import sources.SalesforceCSVExport
 
 import scala.collection.JavaConverters._
-import scala.concurrent.ExecutionContext.Implicits.global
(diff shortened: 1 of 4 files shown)


johnduffell:
that card is for monitoring the number of open files not changing the context?

we should mention on this PR that this is considered best practice but we haven't seen any issues as a result of doing it wrong.
once this is deployed, it's worth keeping an eye on what happens when a breaking news alert goes out, I'm not expecting anything but it's worth checking
👍

paulbrown1982:
👍

lmath:
Good points @johnduffell. I've updated the card and the description here.

PR #175, opened 2017-04-11
Improve logging in paymentDetails

jacobwinch:
It's currently very hard to debug this area of the code without running everything locally and testing manually. Hopefully this additional logging will make it slightly easier to investigate problems in the future.

cc @paulbrown1982 @pvighi @AWare


jacobwinch:
Having problems with TeamCity (hence the Overdue-on-PROD label) - investigating now...

PR #176, opened 2017-04-21
Config changes in preparation for systemd and deb packages

pvighi:
Systemd doesnt support passing the config file as a parameter when starting the app so we have to change how we load the configuration.

The config include-chain is now this (each file contains an include for the next):
application.conf
/etc/gu/members-data-api.private.conf (downloaded from S3)
[STAGE].public.conf (from a stage-specific include in members-data-api.private.conf

Note that members-data-api.private.conf is now performing a double-duty:
- Storing your private credentials!
- Including the correct public conf for your stage: [STAGE].public.conf

@paulbrown1982 @johnduffell @jacobwinch @rtyley


membership-attribute-service/conf/TEST.public.conf
```diff
@@ -0,0 +1,59 @@
+include "touchpoint.DEV.conf"
```
pvighi: I added a custom test configuration (just copied from application.conf and dev.public.conf) otherwise the tests did not pass in team city. I'm not sure if this is the most elegant solution but it works

provisioning/gu-membership-attribute-service-dist/upstart.conf
```diff
@@ -19,7 +19,7 @@ chdir /<APP>
 script
 TOTAL_MEMORY=$(grep MemTotal /proc/meminfo | awk '{ print $2 }')
 HEAP_SIZE_IN_MB=$(python -c "print int($TOTAL_MEMORY * 0.8 / 1024)")
-COMMAND="/<APP>/<APP>-1.0-SNAPSHOT/bin/<APP> -J-Xmx${HEAP_SIZE_IN_MB}m -J-Xms${HEAP_SIZE_IN_MB}m -Dplay.crypto.secret=<PLAY_APP_SECRET> -Dconfig.resource=<Stage>.conf"
+COMMAND="/<APP>/<APP>-1.0-SNAPSHOT/bin/<APP> -J-Xmx${HEAP_SIZE_IN_MB}m -J-Xms${HEAP_SIZE_IN_MB}m -Dplay.crypto.secret=<PLAY_APP_SECRET>"
 echo "$COMMAND" >/<APP>/cmd.txt
 $COMMAND >$LOGFILE 2>&1
 end script
```
pvighi: This script is downloaded from s3 in the launch configuration so we have to make sure to update it in the bucket before deploying these changes

PR #177, opened 2017-04-24
Update membership-common to 0.395

lmath:
Use the newest version of membership common. This includes [my change](https://github.com/guardian/membership-common/pull/460) that ensures that membership common will use the execution context passed implicitly from members-data-api instead of creating another one within the common.


jacobwinch:
👍

PR #179, opened 2017-05-18
Auto-update AMI on deploy & so also use deb packages and systemd

lmath:
* Generates a deb artifact instead of a zip
* Updated launch configuration in the cloudformation to expect a deb artifact
* Updated cloudformation to be useable to create a CODE stack, if a value for the certificate is added. 
* Specifies in deploy.json to check for an ami update before deploy


pvighi:
I think you can delete the upstart.conf file

pvighi:
other than my previous message looks good, but we might want someone else who wasn't involved in this to look at it

membership-attribute-service/conf/deploy.json
```diff
@@ -1,6 +1,20 @@
...
+      "type": "ami-cloudformation-parameter",
+      "apps": [
+        "membership-attribute-service"
+      ],
+      "data": {
+        "amiTags": {
+          "Recipe": "xenial-membership",
+          "AmigoStage": "PROD"
+        },
+        "amiParameter": "ImageId",
```
pvighi: I think this is called AmiId in the cloudformation for this

membership-attribute-service/conf/deploy.json
```diff
@@ -1,6 +1,20 @@
...
+        "cloudFormationStackByTags": true
```
pvighi: remember to add the App,Stage & Stack tags on the prod template

cloudformation/membership-attribute-service.yaml
```diff
@@ -320,7 +309,8 @@ Resources:
   TopicGiraffe:
     Type: AWS::SNS::Topic
     Properties:
-      TopicName: giraffe
+      TopicName:
+        Fn::FindInMap: [ StageVariables, { Ref: Stage }, GiraffeTopic ]
```
paulbrown1982: Just for notes, the removal of this stuff is in https://github.com/guardian/members-data-api/pull/131, which may need to resolve a merge conflict.

PR #182, opened 2017-06-05
Card expiry date in /features

paulbrown1982:
- Updated the Attributes case class to contain the card expiry year and month. This will get serialised to DynamoDB during the Salesforce webhook callback or a backfill of data from Stripe.
- Updated the Features case class to return a boolean of whether the card has expired already alongside the expiry local date in YYYY-mm-dd format.
- Fully removed old PublicTier stuff as this was hackday code.

cc @johnduffell @pvighi @jacobwinch @lmath @AWare @rtyley

### DIFF
--- a/membership-attribute-service/app/controllers/SalesforceHookController.scala
+++ b/membership-attribute-service/app/controllers/SalesforceHookController.scala
@@ -95,7 +95,8 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
       val cardExpiryFromStripeF = (for {
         account <- OptionT(touchpoint.zuoraService.getAccount(membershipSubscription.accountId).map(Option(_)))
         paymentMethod <- OptionT(touchpoint.zuoraService.getPaymentMethod(account.defaultPaymentMethodId.get).map(Option(_)))
-        stripeCustomer <- OptionT(touchpoint.stripeService.Customer.read(paymentMethod.secondTokenId.get).map(Option(_)))
+        customerToken <- OptionT(Future.successful(paymentMethod.secondTokenId))
+        stripeCustomer <- OptionT(touchpoint.stripeService.Customer.read(customerToken).map(Option(_)))
       } yield {
         (stripeCustomer.card.exp_month, stripeCustomer.card.exp_year)
       }).run
--- a/membership-attribute-service/app/controllers/SalesforceHookController.scala
+++ b/membership-attribute-service/app/controllers/SalesforceHookController.scala
@@ -6,7 +6,7 @@ import com.gu.memsub.subsv2.SubscriptionPlan
 import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.typesafe.scalalogging.LazyLogging
-import models.ApiErrors
+import models.{ApiErrors, Attributes}
 import monitoring.Metrics
 import parsers.Salesforce._
 import parsers.{Salesforce => SFParser}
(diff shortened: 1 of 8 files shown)


membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -60,26 +60,47 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
...
-        metrics.put("Update", 1)
-        putItemResult
-      }.recover { case e: Throwable =>
-        logger.warn(s"Failed to insert ${attrsUpdatedWithZuoraOpt.getOrElse(attrs)} into ${touchpoint.dynamoAttributesTable}. Salesforce should retry.", e)
-        Failure
+      val attributesWithZuoraTier = {
+        val tierFromSalesforce = salesforceAttributes.Tier
+        val tierFromZuora = membershipSubscription.plan.charges.benefit.id
+        if (tierFromZuora != tierFromSalesforce) logger.error(s"Differing tier info for $sfId : sf=$tierFromSalesforce zuora=$tierFromZuora")
+        salesforceAttributes.copy(Tier = tierFromZuora)
```
johnduffell: you went back to the copying solution in the end?

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -60,26 +60,44 @@ def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
   }
 
   def updateMemberRecord(membershipUpdate: MembershipUpdate): Future[Object] = {
-    val attrs = membershipUpdate.attributes
+    def updateDynamo(attributes: Attributes) = attributeService.set(attributes).map { putItemResult =>
```
johnduffell: small point I'd put the `=` last on the line and put the rest on a new line

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -1,41 +1,48 @@
...
 }
 
 object Attributes {
 
-  val ignore = OWrites[Any](_ => Json.obj())
   implicit val jsWrite: OWrites[Attributes] = (
     (__ \ "userId").write[String] and
     (__ \ "tier").write[String] and
     (__ \ "membershipNumber").writeNullable[String] and
     (__ \ "adFree").writeNullable[Boolean] and
```
johnduffell: I thought adfree wasn't actually used in the end?  or it's still needed for future?

johnduffell:
looks good! 👍

membership-attribute-service/app/models/Attributes.scala
paulbrown1982: It's read from Dynamo but never written. Thanks for the reminder though, I'll double check it doesn't get overwritten by the webhook...

johnduffell:
fyi @svillafe is also working on this area of code to put contributors in, since we found salesforce doesn't seem to call back the hook for contributors, we are thinking maybe it's better to have zuora call through to update it as Jacob mentioned the salesforce hook isn't the long term solution.

paulbrown1982:
Ah yes, this solution is limited to existing members, not recurring contributors or subscribers. Also should Contributors shouldn't go in members-data-api's Dynamo table? A more generic card expiry solution should be built alongside this. So the overarching question is, do we go forward with this, or close the branch?

johnduffell:
I think it's right to have one table to everything, however we did have a debate earlier about where the data should come from (should it be pushed in by the contributions frontend or pulled from zuora).
But fundamentally I think members data api's dynamo cache is all about "here's an identity user, please tell me what we need to do/reflect on the web site and app based on their paying relationship" so the table aligns with how we use/look it up rather than a product level.
I think card expiry and whether they have a non-membership paying relationship fits in to that?

paulbrown1982:
Agreed. Even though we don't fulfil any benefits to recurring contributors, we still need to fulfil the fact that they are a Contributor so the requesting channel can stop showing asks, etc.

So the same question applies, should I merge now or or wait for @svillafe to merge his changes, and rebase again? This task has little urgency, @svillafe what's your urgency/appetite to rebase off this PR?

johnduffell:
merge it!  I'd say get it merged sooner rather than later, so far we just sketched out the approach for Satiago's stuff but haven't really started coding as he is looking into triggering when contributors are added or deleted.  Originally he was going to POST to the update endpoint but we discussed it and went slightly off that approach.

PR #185, opened 2017-06-08
let members-data-api log to kibana

johnduffell:
based on https://github.com/guardian/subscriptions-frontend/pull/946
Pushing logs into kinesis and thereafter kibana from members-data-api
This all part of my evil plan to have one source of truth for tech/debug that's easy to search so we can diagnose issues more reliably.

Comments/suggestions/:+1:s! @guardian/membership-and-subscriptions

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -86,6 +86,7 @@ Mappings:
       DynamoDBBehaviourTable: arn:aws:dynamodb:*:*:table/MembershipBehaviour-UAT
       DynamoDBBehaviourTableTestUsers: arn:aws:dynamodb:*:*:table/MembershipBehaviour-UAT
       GiraffeTopic: giraffe-code
+      SSLCertificate: arn:aws:iam::865473395570:server-certificate/members-data-api.theguardian.com
 
 Resources:
   MembershipRole:
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -60,6 +60,12 @@ Parameters:
   VulnerabilityScanningSecurityGroup:
     Description: Security group that grants access to the account's Vulnerability Scanner
     Type: AWS::EC2::SecurityGroup::Id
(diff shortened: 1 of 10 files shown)


cloudformation/membership-attribute-service.yaml
```diff
@@ -80,6 +86,7 @@ Mappings:
       DynamoDBBehaviourTable: arn:aws:dynamodb:*:*:table/MembershipBehaviour-UAT
       DynamoDBBehaviourTableTestUsers: arn:aws:dynamodb:*:*:table/MembershipBehaviour-UAT
       GiraffeTopic: giraffe-code
+      SSLCertificate: arn:aws:iam::865473395570:server-certificate/members-data-api.theguardian.com
```
johnduffell: this is just a duplicate of the above ssl cert, if we really want to use this properly we should register members-data-api.code.dev-etc and get a cert for that

membership-attribute-service/conf/TEST.public.conf
```diff
@@ -1,6 +1,4 @@
 include "DEV.public"
 include "application"
 
-
-
-
+logstash.enabled=true
```
johnduffell: not sure what Test.public.conf is used for, if anything?

membership-attribute-service/conf/CODE.public.conf
```diff
@@ -0,0 +1,3 @@
+include "DEV.public"
+
+logstash.enabled=true
```
johnduffell: I've updated the CODE props in S3 to point to this file instead of the DEV one as they did before

membership-attribute-service/conf/TEST.public.conf
jacobwinch: Yes, not sure either. It looks like @pvighi added this here: https://github.com/guardian/members-data-api/commit/97f10aa4a93212595a85601e4ab22430f40ba6cb

jacobwinch:
Looks good to me 👍

AWare:
This fixes. #166

PR #186, opened 2017-06-08
Remove PublicTier field from Attributes data model

svillafe:
During the implementation of the contribution field in Members Data API, we discovered the `publicTier` field in the `Attributes.scala` file. After talking about this with @rtyley, he thinks that we are not using that field at all and probably the best thing to do, is to delete it. Therefore, this PR. 

cc: @paulbrown1982 @JustinPinner please review this PR.

### DIFF
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -15,11 +15,14 @@ object ContentAccess {
   implicit val jsWrite = Json.writes[ContentAccess]
 }
 
-case class Attributes(UserId: String, Tier: String, MembershipNumber: Option[String], PublicTier: Option[Boolean] = None, AdFree: Option[Boolean] = None) {
+case class Attributes(
+                       UserId: String,
+                       Tier: String,
+                       MembershipNumber: Option[String],
+                       AdFree: Option[Boolean] = None) {
   require(Tier.nonEmpty)
   require(UserId.nonEmpty)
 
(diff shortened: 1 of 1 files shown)


JustinPinner:
If you're worried about breaking ad-free, it's only an internal test at the moment so it's not the end of the world if it does break for a while. I don't think it will, but it can be checked by hitting the `user-attributes/me/features` endpoint after this is merged.

rtyley:
Note that the `PublicTier` was added part of a hack day (!) in June 2016 with https://github.com/guardian/members-data-api/pull/109, but unfortunately wasn't removed with the clean-up PR in https://github.com/guardian/members-data-api/pull/129 :cry: 

We think that the `AdFree` and `PublicTier` attributes were later inadvertently [flipped over](https://github.com/guardian/members-data-api/pull/145/files#diff-f6762c4a98ab2b33595daa97438d314c) in the 'Attributes' endpoint code, but thankfully the values were never actually output on https://members-data-api.theguardian.com/user-attributes/me/membership, so no negative consequences arose... apart from confusion...!

svillafe:
Thank you for your point of view @JustinPinner,  I would like to wait for @paulbrown1982's review since he could have more information about the purpose of this field and if we are actually using it for something.

svillafe:
Thank you Paul, I will merge now and test against @JustinPinner's test.

svillafe:
It is working in **PROD** a.k.a. @JustinPinner's test is passing.
picture 278

PR #187, opened 2017-06-12
Added logging to the Salesforce hook.

svillafe:
Currently, Salesforce is being called when someone becomes a contributor. Nevertheless, that information is not in the member's data API. This PR adds logging to try to know why is that happening.


johnduffell:
👍 great!
we should add request id more generally in future I think, to make it easier to trace calls through the system

PR #188, opened 2017-06-12
Fixed unsafe get which I missed which caused https://sentry.io/the-gu…

paulbrown1982:
Fixed unsafe get which I missed which caused https://sentry.io/the-guardian/members-data-api/issues/292706477/

cc @johnduffell @jacobwinch @svillafe


johnduffell:
👍

PR #189, opened 2017-06-14
Updated Scanamo to the latest version.

svillafe:
We were using version `0.4.0`, that version does not support update operations. According to [Scanamo's changelog](https://github.com/guardian/scanamo/blob/master/CHANGES.md) this was introduced in version `0.6.0`.

### DIFF
--- a/membership-attribute-service/app/services/ScanamoBehaviourService.scala
+++ b/membership-attribute-service/app/services/ScanamoBehaviourService.scala
@@ -15,8 +15,8 @@ class ScanamoBehaviourService(client: AmazonDynamoDBAsyncClient, table: String) 
   def run[T] = ScanamoAsync.exec[T](client) _
 
   override def get(userId: String): Future[Option[Behaviour]] =
-    run(scanamo.get('userId -> userId).map(_.flatMap(_.toOption)))
-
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/services/ScanamoBehaviourService.scala
```diff
@@ -15,8 +15,8 @@ class ScanamoBehaviourService(client: AmazonDynamoDBAsyncClient, table: String)
   def run[T] = ScanamoAsync.exec[T](client) _
 
   override def get(userId: String): Future[Option[Behaviour]] =
-    run(scanamo.get('userId -> userId).map(_.flatMap(_.toOption)))
-
+    run(scanamo.get('userId -> userId).map(_.flatMap(_.right.toOption)))
```
johnduffell: would be nice to know what the (new?) left represents - if it's a message we could at least leftMap it into the logger so at least it's not lost completely

johnduffell:
thanks! that extra logging is a definite improvement 👍 🚀

svillafe:
It is working on **PROD**, I created a Supporter in UAT and the test user is in the DynamoDB UAT table.

PR #191, opened 2017-06-15
Endpoint for contributors

svillafe:
# Why are you doing this:

In order to connect the support-workers step functions with members-data-api, we created the following route.

`PUT  /user-attributes/:identityId?testUser=[true|false]`

This endpoint will retrieve from Zuora and Salesforce all the information about that id and update the attribute's table. The entity which makes this request must sign it via a shared secret.

This PR changes the Attributes definition, it adds a new field (column in DynamoDB) called `Contributor`.  Additionally, it makes the member's tier optional. 

The idea is that a user can be a member and a contributor. The user can start being a member and after a while become a monthly-contributor or vice-versa. 

## Breaking change:

This PR will collide with https://github.com/guardian/members-data-api/pull/188. After going through it with @jacobwinch we think that credit card introduced in  https://github.com/guardian/members-data-api/pull/188  should be specific for membership and create new ones for contributors. Moreover, we could extract them in a separate table. cc: @paulbrown1982 

### 1 Member, then Contributor

1.  Become a member.
2. Execute the call:

`GET        /user-attributes/me/membership`

#### Answer:

```
{
"userId": "30000597",
"tier": "Supporter",
"membershipNumber": "1234",
"contentAccess": {
"member": true,
"paidMember": true,
"contributor": false
}
}
```

3. Execute the call :
`GET        /user-attributes/me`

#### Answer:

```
{
"userId": "30000597",
"tier": "Supporter",
"membershipNumber": "1234",
"contentAccess": {
"member": true,
"paidMember": true,
"contributor": false
}
}
```

4.  Become a monthly contributor.
5. Execute the call:
`GET        /user-attributes/me`

#### Answer:

{
"userId": "30000597",
"tier": "Supporter",
"membershipNumber": "324143",
"contributionFrequency": "Monthly Contribution",
"contentAccess": {
"member": true,
"paidMember": true,
"contributor": true
}
}

6. Execute the call :
`GET        /user-attributes/me/membership`

#### Answer:
{
"userId": "30000597",
"tier": "Supporter",
"membershipNumber": "324143",
"contributionFrequency": "Monthly Contribution",
"contentAccess": {
"member": true,
"paidMember": true,
"contributor": true
}
}

### 2 Contributor, then Member

1.  Become a monthly contributor.
2. Execute the call:

`GET        /user-attributes/me`

#### Answer:
```
{
    "userId": "30000585",
    "contributionFrequency": "Monthly Contribution",
    "contentAccess": {
        "member": false,
        "paidMember": false,
        "contributor": true
    }
}
```
3. Execute the call :
`GET        /user-attributes/me/membership`

#### Answer:
```
{
    "message": "Not found",
    "details": "User was found but they are not a member",
    "statusCode": 404
}
```

4.  Become a member.

5. Execute the call:
`GET        /user-attributes/me`

#### Answer:
```
{
"userId": "30000592",
"tier": "Supporter",
"membershipNumber": "1234",
"cardExpirationMonth": 12,
"cardExpirationYear": 2042,
"contributionFrequency": "Monthly Contribution",
"contentAccess": {
"member": true,
"paidMember": true,
"contributor": true
}
}
```
6. Execute the call :
`GET        /user-attributes/me/membership`

#### Answer:
```
{
"userId": "30000592",
"tier": "Supporter",
"membershipNumber": "1234",
"cardExpirationMonth": 12,
"cardExpirationYear": 2042,
"contributionFrequency": "Monthly Contribution",
"contentAccess": {
"member": true,
"paidMember": true,
"contributor": true
}
}
```


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -70,4 +82,35 @@ class AttributeController extends Controller with LazyLogging {
...
+        res <- EitherT(tp.attrService.update(attributes).map(\/.right))
+      } yield attributes
+
+    result.run.map(_.fold(
+      error => {
+        logger.error(s"Failed to update attributes - $error")
+        ApiErrors.badRequest(error)
+      },
+      attributes => {
+        logger.info(s"${attributes.UserId} -> ${attributes.Tier} || ${attributes.Contributor}")
```
rupertbates: We probably shouldn't log this information in production because of GDPR

membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -2,11 +2,15 @@ package services
...
-import com.amazonaws.services.dynamodbv2.model.{DeleteItemResult, PutItemResult}
+import com.amazonaws.services.dynamodbv2.model.{DeleteItemResult, PutItemResult, UpdateItemResult}
+import com.gu.memsub.Benefit.Contributor
 import play.api.libs.concurrent.Execution.Implicits._
+
 import scala.concurrent.Future
 import com.gu.scanamo.syntax._
 import com.gu.scanamo._
+import com.gu.scanamo.error.DynamoReadError
+import com.gu.scanamo.syntax.{set => scanmoSet}
```
rupertbates: Typo: should be scanamoSet

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -13,19 +13,31 @@ import models._
...
+        block(request)
+      } else {
+        Future.successful(Forbidden)
+      }
+    }
+  }
+
   lazy val corsFilter = CORSActionBuilder(Config.corsConfig)
   lazy val backendAction = NoCacheAction andThen corsFilter andThen BackendFromCookieAction
+  lazy val backendForSyncWithZuora = NoCacheAction andThen  signed andThen WithBackendFromUserIdAction
```
johnduffell: as mentioned previously I see no reason to have it signed as the only malicious action would be a DOS attack which would be just as possible through any of the other endpoints that hit zuora directly, also unsigned it makes it easier for us to do an update from other tools e.g. useradmin

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -56,7 +68,7 @@ class AttributeController extends Controller with LazyLogging {
         request.touchpoint.attrService.get(id).map {
           case Some(attrs) =>
             onSuccess(attrs).withHeaders(
-              "X-Gu-Membership-Tier" -> attrs.Tier,
+              "X-Gu-Membership-Tier" -> attrs.Tier.get,
```
johnduffell: if someone with an identity id calls this but not a member, will they get a 500 error?

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -18,17 +18,19 @@ object ContentAccess {
...
                        CardExpirationMonth: Option[Int] = None,
-                       CardExpirationYear: Option[Int] = None) {
-  require(Tier.nonEmpty)
+                       CardExpirationYear: Option[Int] = None,
+                       Contributor: Option[String] = None) {
+
   require(UserId.nonEmpty)
 
-  lazy val isFriendTier = Tier.equalsIgnoreCase("friend")
+  lazy val isFriendTier = Tier.getOrElse("").equalsIgnoreCase("friend")
```
johnduffell: `Tier.exists(_.equalsIgnoreCase("friend"))`

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -18,17 +18,19 @@ object ContentAccess {
...
                        UserId: String,
-                       Tier: String,
+                       Tier: Option[String] = None,
                        MembershipNumber: Option[String],
                        AdFree: Option[Boolean] = None,
                        CardExpirationMonth: Option[Int] = None,
-                       CardExpirationYear: Option[Int] = None) {
-  require(Tier.nonEmpty)
+                       CardExpirationYear: Option[Int] = None,
+                       Contributor: Option[String] = None) {
```
johnduffell: `contributor` should be a boolean? or is this actually `contributorFrequency`? or `contributorAmount`?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -70,4 +82,35 @@ class AttributeController extends Controller with LazyLogging {
...
+
+    val result: EitherT[Future, String, Attributes] =
+      for {
+        contact <- EitherT(tp.contactRepo.get(identityId).map(_ \/> s"No contact for $identityId"))
+        memSubF = EitherT[Future, String, Option[Subscription[SubscriptionPlan.Member]]](tp.subService.current[SubscriptionPlan.Member](contact).map(a => \/.right(a.headOption)))
+        conSubF = EitherT[Future, String, Option[Subscription[SubscriptionPlan.Contributor]]](tp.subService.current[SubscriptionPlan.Contributor](contact).map(a => \/.right(a.headOption)))
+        memSub <- memSubF
+        conSub <- conSubF
+        _ <- EitherT(Future.successful(if (memSub.isEmpty && conSub.isEmpty) \/.left("No paying relationship") else \/.right(())))
+        attributes = Attributes( UserId = identityId, Tier = memSub.map(_.plan.charges.benefit.id), MembershipNumber = contact.regNumber, Contributor = conSub.map(_.plan.charges.benefit.id))
```
svillafe: The benefitId of a contributor currently is Contributor. I believe we should change it to monthly-contribution, so in the future, if we want to add annual or other type is easily extensible.

The API returns the following data for a member who is a contributor:
picture 279

membership-attribute-service/app/repositories/MembershipAttributesSerializer.scala
```diff
@@ -24,13 +24,13 @@ case class MembershipAttributesSerializer(tableName: String)
...
-      AttributeNames.tier -> membershipAttributes.Tier
+      AttributeNames.tier -> membershipAttributes.Tier.getOrElse("")
     ).filter(_._2.nonEmpty).map(mkAttribute[String])
 
   override def fromAttributeMap(item: collection.mutable.Map[String, AttributeValue]) =
     Attributes(
       UserId = item(AttributeNames.userId),
       MembershipNumber = item.get(AttributeNames.membershipNumber).map(_.getS),
-      Tier = item(AttributeNames.tier)
+      Tier = Some(item(AttributeNames.tier))
```
johnduffell: Tier is just wrapped in a Some, but how about if it was written as a ""?  It would be filtered out and not written, so would `item` have thrown an exception? or returned ""?  whatever actually happens, it should end up returning a None presumably so that it's symmetrical with toAttributeMap

membership-attribute-service/app/services/ScanamoAttributeService.scala
johnduffell: scanamoSet
missing an 'a'

membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -32,5 +36,7 @@ class ScanamoAttributeService(client: AmazonDynamoDBAsyncClient, table: String)
 
   override def set(attributes: Attributes): Future[PutItemResult] = run(scanamo.put(attributes))
 
+  override def update(attributes: Attributes): Future[Either[DynamoReadError, Attributes]] = run(scanamo.update('UserId -> attributes.UserId, scanmoSet('Tier -> attributes.Tier) and scanmoSet('MembershipNumber -> attributes.MembershipNumber) and scanmoSet('Contributor -> attributes.Contributor)))
```
johnduffell: I'd split this out one line per attribute for readability

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: it's not coming in for a year! also info only goes to the instance's disk, cloudwatch, and membership-logs kibana, which I believe are all access controlled and encrypted.

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: Contributor starts with caps in the json key when the others are lower case.
also if it's going to hold the frequency it should be called confributorFrequency or at least contributorType

membership-attribute-service/app/services/ScanamoAttributeService.scala
svillafe: Addressed in f909ae9e6799d0909136bc608289c19c5de0a614

membership-attribute-service/app/services/ScanamoAttributeService.scala
svillafe: Addressed in f909ae9e6799d0909136bc608289c19c5de0a614

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -44,11 +46,12 @@ object Attributes {
...
   implicit val jsWrite: OWrites[Attributes] = (
     (__ \ "userId").write[String] and
-    (__ \ "tier").write[String] and
+    (__ \ "tier").writeNullable[String] and
     (__ \ "membershipNumber").writeNullable[String] and
     (__ \ "adFree").writeNullable[Boolean] and
     (__ \ "cardExpirationMonth").writeNullable[Int] and
-    (__ \ "cardExpirationYear").writeNullable[Int]
+    (__ \ "cardExpirationYear").writeNullable[Int] and
+    (__ \ "Contributor").writeNullable[String]
```
paulbrown1982: This should be lower case "contributor" or "contributorFrequency" as per @johnduffell's comment.

membership-attribute-service/app/services/ScanamoAttributeService.scala
svillafe: Addressed in f909ae9e6799d0909136bc608289c19c5de0a614

membership-attribute-service/app/models/Attributes.scala
svillafe: Addressed here: c28e8c09fc89f544613e090e09e151f50dc16afa `Contributor` became `ContributionFrequency`

membership-attribute-service/app/controllers/AttributeController.scala
svillafe: Addressed here: c28e8c0 Contributor became ContributionFrequency

membership-attribute-service/app/models/Attributes.scala
svillafe: The field will be called: `ContributionFrequency`. Introduced in: Addressed here: c28e8c0 Contributor became ContributionFrequency

membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -41,7 +41,7 @@ class ScanamoAttributeService(client: AmazonDynamoDBAsyncClient, table: String)
       scanamo.update('UserId -> attributes.UserId,
        scanamoSet('Tier -> attributes.Tier) and
         scanamoSet('MembershipNumber -> attributes.MembershipNumber) and
-        scanamoSet('Contributor -> attributes.Contributor))
+        scanamoSet('Contributor -> attributes.ContributionFrequency))
```
paulbrown1982: Should the key be ContributionFrequency?

svillafe: Addressed in 37855c4cebef2dd3c06e190519ca42102f8d774c

membership-attribute-service/app/controllers/AttributeController.scala
svillafe: There is an extra error here. If the user is only a contributor we should return 404 in order to have backwards compatibility.

membership-attribute-service/app/controllers/AttributeController.scala
svillafe: After an offline discussion, it is ok since the logs are encrypted.

membership-attribute-service/app/models/Attributes.scala
svillafe: Addressed in f1350a42c5f3a475455bfa37c46ee5953bef86d1

membership-attribute-service/app/controllers/AttributeController.scala
svillafe: Addressed in b7a12f88a249367db9755bfbaa9aa415f47633f5

membership-attribute-service/app/controllers/AttributeController.scala
svillafe: After talking with @adamnfish, he mentioned that as an advantage of having the signed request we can reduce the surface of a DOS attack because we won't hit Zuora or Salesforce if the request is not valid. At the same time, this introduces an extra layer of maintenance because if the shared secret changes a lot, we should update all the app that use this endpoint.

membership-attribute-service/app/actions/WithBackendFromUserIdAction.scala
```diff
@@ -0,0 +1,21 @@
...
+import components.{NormalTouchpointComponents, TestTouchpointComponents}
+import play.api.mvc.{ActionRefiner, Request, Result}
+import play.api.libs.concurrent.Execution.Implicits._
+
+import scala.concurrent.Future
+
+object WithBackendFromUserIdAction extends ActionRefiner[Request, BackendRequest] {
+
+  override protected def refine[A](request: Request[A]): Future[Either[Result, BackendRequest[A]]] =
+    Future {
```
davidfurey: Use `Future.successful {` instead of `Future {`

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -13,19 +13,31 @@ import models._
...
+import scala.concurrent.{Future, Promise}
 import scalaz.std.scalaFuture._
 import scalaz.syntax.std.option._
 import scalaz.{EitherT, \/}
+import configuration.Config.authentication
 
 
 class AttributeController extends Controller with LazyLogging {
 
+  def signed(): ActionBuilder[Request] = new ActionBuilder[Request] {
```
davidfurey: I don't think that `signed()` is a good name for this, since the request has an API key present, it isn't actually signed. Maybe `apiKeyFilter`?

membership-attribute-service/app/configuration/Config.scala
```diff
@@ -36,6 +37,10 @@ object Config {
     val region = Regions.EU_WEST_1
   }
 
+  object authentication {
+    val key = config.getString("authentication.key")
```
davidfurey: It will be easier to rotate keys if you allow multiple keys to be valid at a time.  A simple way to do this would be `config.getString("authentication.keys").split(',')`

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -13,19 +13,31 @@ import models._
...
+    def invokeBlock[A](request: Request[A], block: (Request[A]) => Future[Result]) = {
+      if (request.headers.get("authentication").contains(authentication.key)) {
+        block(request)
+      } else {
+        Future.successful(Forbidden)
```
davidfurey: Could make this more helpful `Forbidden("This endpoint requires an API key")`

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -13,19 +13,31 @@ import models._
...
 import scalaz.syntax.std.option._
 import scalaz.{EitherT, \/}
+import configuration.Config.authentication
 
 
 class AttributeController extends Controller with LazyLogging {
 
+  def signed(): ActionBuilder[Request] = new ActionBuilder[Request] {
+    def invokeBlock[A](request: Request[A], block: (Request[A]) => Future[Result]) = {
+      if (request.headers.get("authentication").contains(authentication.key)) {
```
davidfurey: In Identity, the authentication key is prefixed by the string "Bearer", so it might be good to do the same?

adamnfish: Unless I'm mistaken, Identity uses the `Authorization` header. This header was originally used for Basic Auth. the use of `Bearer` is encouraged when you are providing a token (and indeed, specified in use cases like OAuth).

It looks like you've used `authentication`. I can't think of a standard where use of this header is specified so you can use whatever format you want. I'd recommend using `Authorization: Bearer token` for consistency, but if you want to stick to `authentication` it's probably worth at least capitalising the first letter to be consistent with other HTTP headers.

davidfurey: Thanks Adam.  I misread the line and thought that it said `Authorization`.  Agree on the capitalisation change, although the request headers map is case insensitive.

svillafe: Great, thank you for your reviews guys, I will push a new commit using the correct headers.

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: fair enough! thanks for checking this out!  If the shared secret is in AWS it should be easier to keep in step.

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: if it's a custom header, it should start with `X-`

johnduffell: sorry, actually not, I'm 5 years out of date according to https://tools.ietf.org/html/rfc6648

membership-attribute-service/test/repositories/DynamoAttributeServiceTest.scala
```diff
@@ -68,17 +68,17 @@ class DynamoAttributeServiceTest extends Specification {
...
+      Attributes("4567", Some("Partner"), None)
     )
 
     "Fetch many people by user id" in {
       Await.result(Future.sequence(testUsers.map(repo.set)), 5.seconds)
       Await.result(repo.getMany(List("1234", "3456", "abcd")), 5.seconds) mustEqual Seq(
-        Attributes("1234", "Partner", None),
-        Attributes("3456", "Partner", None)
+        Attributes("1234", Some("Partner"), None),
+        Attributes("3456", Some("Partner"), None)
```
johnduffell: are any of the new bits worth automated testing?

membership-attribute-service/app/repositories/MembershipAttributesSerializer.scala
johnduffell: ok, so if Tier is not set we would get an exception thrown at this point http://www.scala-lang.org/api/current/scala/collection/mutable/Map.html#apply(key:K):V

membership-attribute-service/conf/DEV.public.conf
```diff
@@ -24,6 +24,11 @@ play.filters.cors.allowedOrigins = [
   "https://thegulocal.com"
 ]
 
+authentication {
```
johnduffell: don't bother changing it, but for a single value you cn just do 
`authentication.key="a-nice-secret"`
all on one line

svillafe:
After testing the apps with the iOS (Antonio) and Android (Max) team, everything looks like is working correctly.

johnduffell:
great work getting this one out there! :1st_place_medal:

PR #192, opened 2017-06-19
store membership join date in dynamo and return from /features call

johnduffell:
This is to let us target messages on the client side to people who joined at certain times.  This lets us serve appropriate messages on anniversaries to make people feel more valued as a member based on what motivated joiners at that time.

Note: you can try viewing this with [?w=1](https://github.com/guardian/members-data-api/pull/192/files?w=1) to ignore all the indentation changes, but then it's a lot harder to comment...

The technical side is when we get the message from salesforce that someone has joined, we mirror the [contractEffectiveDate](https://github.com/guardian/membership-common/blob/919ed929b8f07bc4b399caf64d3e42be43aabe9b/src/main/scala/com/gu/memsub/subsv2/reads/SubJsonReads.scala#L110) which is called startDate for some reason (someone should fix that)
Then when someone hits the features endpoint, we return the extra field `startDate`
```json
{"userId":"30000594","adFree":false,"adblockMessage":false,"cardHasExpired":false,"cardExpires":"2022-02-28","startDate":"2017-06-19"}
```
This also works for existing users who don't have a record with the start/join date by just omitting the field completely.

I'm not sure about naming, since this is specifically the memberhip start date I'm not sure whether to call it joinDate.  I wonder how this would interact with contributions, as we might want to know when they started too.  We can add another contributionStartDate for that if we need it?

In due course we need to back populate the existing users, this would be by doing a query on zuora, and I found a batch loader in the codebase which may be old, but this is an essential step to actually being able to target existing members.

I will have a quick look after lunch and try to work out which bits could do with tests.

Into the mix I did a few small bits of tidy up, deleting an unused endpoint for update (no one called it through fastly as I grepped the logs) and also removed one layer from the MembershipUpdate object.

@lmath @jacobwinch @svillafe @pvighi would be great to get a bit of a review on this then I can fix things up this afternoon and potentially merge @svillafe 's [PR](https://github.com/guardian/members-data-api/pull/191) into it once he's merged.

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -29,28 +29,6 @@ class AttributeController extends Controller with LazyLogging {
   lazy val authenticationService: AuthenticationService = IdentityAuthService
   lazy val metrics = Metrics("AttributesController")
 
-  def update = BackendFromCookieAction.async { implicit request =>
-
-    val result: EitherT[Future, String, Attributes] = for {
-      id <- EitherT(Future.successful(authenticationService.userId \/> "No user"))
-      contact <- EitherT(request.touchpoint.contactRepo.get(id).map(_ \/> s"No contact for $id"))
-      sub <- EitherT(request.touchpoint.subService.current[SubscriptionPlan.Member](contact).map(_.headOption \/> s"No sub for $id"))
-      attributes = Attributes(id, sub.plan.charges.benefit.id, contact.regNumber)
-      res <- EitherT(request.touchpoint.attrService.set(attributes).map(\/.right))
-    } yield attributes
-
-    result.run.map(_.fold(
-      error => {
-        logger.error(s"Failed to update attributes - $error")
-        ApiErrors.badRequest(error)
-      },
-      attributes => {
-        logger.info(s"${attributes.UserId} -> ${attributes.Tier}")
-        Ok(Json.obj("updated" -> true))
-      }
-    ))
-  }
-
   private def lookup(endpointDescription: String, onSuccess: Attributes => Result, onNotFound: Option[Result] = None) = backendAction.async { request =>
       authenticationService.userId(request).map[Future[Result]] { id =>
         request.touchpoint.attrService.get(id).map {
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -43,10 +43,10 @@ class AttributeController extends Controller with LazyLogging {
   lazy val authenticationService: AuthenticationService = IdentityAuthService
   lazy val metrics = Metrics("AttributesController")
 
-  private def lookup(endpointDescription: String, onSuccess: Attributes => Result, onNotFound: Option[Result] = None) = backendAction.async { request =>
+  private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessMemberAndOrContributor: Attributes => Result, onNotFound: Result) = backendAction.async { request =>
       authenticationService.userId(request).map[Future[Result]] { id =>
         request.touchpoint.attrService.get(id).map {
-          case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _)) =>
+          case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _, _)) =>
             logger.info(s"$id is a member - $endpointDescription - $attrs")
             onSuccessMember(attrs).withHeaders(
               "X-Gu-Membership-Tier" -> tier,
@@ -86,7 +86,13 @@ class AttributeController extends Controller with LazyLogging {
         memSub <- memSubF
         conSub <- conSubF
         _ <- EitherT(Future.successful(if (memSub.isEmpty && conSub.isEmpty) \/.left("No paying relationship") else \/.right(())))
-        attributes = Attributes( UserId = identityId, Tier = memSub.map(_.plan.charges.benefit.id), MembershipNumber = contact.regNumber, ContributionFrequency = conSub.map(_.plan.name))
+        attributes = Attributes(
+          UserId = identityId,
+          Tier = memSub.map(_.plan.charges.benefit.id),
+          MembershipNumber = contact.regNumber,
+          ContributionFrequency = conSub.map(_.plan.name),
+          MembershipJoinDate = memSub.map(_.startDate)
+        )
         res <- EitherT(tp.attrService.update(attributes).map(\/.right))
       } yield attributes
 
--- a/membership-attribute-service/app/controllers/SalesforceHookController.scala
+++ b/membership-attribute-service/app/controllers/SalesforceHookController.scala
@@ -40,99 +40,153 @@ class SalesforceHookController extends LazyLogging {
     </soapenv:Envelope>
   )
 
-def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
+  def createAttributes = BackendFromSalesforceAction.async(parse.xml) { request =>
 
-  val touchpoint = request.touchpoint
-  val validOrgId = touchpoint.sfOrganisationId
-  val attributeService = touchpoint.attrService
-  val requestId = scala.util.Random.nextInt
+    val touchpoint = request.touchpoint
+    val validOrgId = touchpoint.sfOrganisationId
+    val attributeService = touchpoint.attrService
 
-  implicit val pf = Membership
+    object ContextLogging {
+      case class RequestId(value: Int)
+      implicit val requestId = RequestId(scala.util.Random.nextInt)
+      def info(message: String)(implicit requestId: RequestId) =
+        logger.info(s"$requestId: $message")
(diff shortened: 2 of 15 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -29,28 +29,6 @@ class AttributeController extends Controller with LazyLogging {
   lazy val authenticationService: AuthenticationService = IdentityAuthService
   lazy val metrics = Metrics("AttributesController")
 
-  def update = BackendFromCookieAction.async { implicit request =>
```
johnduffell: this was not used by anyone according to the fastly logs from 11th Jun 00:00 until 12th Jun 10:19am

AWare:
I was looking at having frontend call the update endpoint when a user hit the mma page, and kind of put that on hold when everyone wanted to make changes to the schema. (Although deleting it may be more sensible, we need to think about how we want to populate members data api, as the salesforce call isn't 100% reliable)

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -40,99 +40,153 @@ class SalesforceHookController extends LazyLogging {
...
+      }
+
     }
-  }
+    import ContextLogging._
+
+    implicit val pf = Membership
 
-  def updateMemberRecord(membershipUpdate: MembershipUpdate): Future[Object] = {
+    info(s"Called from Saleforce to modify the Members Data API")
```
johnduffell: we only are logging the dynamo record or information that goes into it.  the extra logging is basically this:
```
INFO: RequestId(-1247831390): Called from Saleforce to modify the Members Data API
INFO: RequestId(-1247831390): Parsed Salesforce message successfully. Salesforce sent 1 objects to update: List(MembershipUpdate(30000594,Supporter,Some(003g000001VLiO6AAL)))
INFO: RequestId(-1247831390): Salesforce called has been parsed. Attrs: MembershipUpdate(30000594,Supporter,Some(003g000001VLiO6AAL))
INFO: RequestId(-1247831390): contact id from SF {Some(003g000001VLiO6AAL)}
INFO: RequestId(-1247831390): current subscriptions {List(Id(2c92c0f95cac811c015cbff6a181648b))}
id='2c92c0f95cac811c015cbff6a14c6487'
INFO: RequestId(-1247831390): ready to update dynamo with Attributes(30000594,Supporter,Some(003g000001VLiO6AAL),None,Some(2),Some(2022),Some(2017-06-19))
INFO: RequestId(-1247831390): Successfully inserted Attributes(30000594,Supporter,Some(003g000001VLiO6AAL),None,Some(2),Some(2022),Some(2017-06-19)) into MembershipAttributes-DEV.

```

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -40,99 +40,153 @@ class SalesforceHookController extends LazyLogging {
...
+        }
+      }).run.flatMap {
+        case Some(zuoraAttributesF) =>
+          zuoraAttributesF.onSuccess{ case attr =>
+            info(s"ready to update dynamo with $attr")
+          }
+          zuoraAttributesF.flatMap(updateDynamo)
+        case None =>
+          error(s"Couldn't update $membershipUpdate with information from Zuora")
+          Future.successful(Failure)
```
johnduffell: we no longer touch the record at all if we couldn't find the user in zuora

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -17,12 +17,14 @@ object ContentAccess {
...
-                       AdFree: Option[Boolean] = None,
-                       CardExpirationMonth: Option[Int] = None,
-                       CardExpirationYear: Option[Int] = None) {
+  UserId: String,
+  Tier: String,
+  MembershipNumber: Option[String],
+  AdFree: Option[Boolean] = None,
+  CardExpirationMonth: Option[Int] = None,
+  CardExpirationYear: Option[Int] = None,
+  StartDate: Option[LocalDate] // TODO startDate shouldn't be optional once we've backfilled it everywhere
```
johnduffell: so this is the membership joining date, not the contributions start date

membership-attribute-service/app/repositories/MembershipAttributesSerializer.scala
```diff
@@ -24,13 +26,16 @@ case class MembershipAttributesSerializer(tableName: String)
...
     ).filter(_._2.nonEmpty).map(mkAttribute[String])
 
-  override def fromAttributeMap(item: collection.mutable.Map[String, AttributeValue]) =
+  override def fromAttributeMap(item: collection.mutable.Map[String, AttributeValue]) = {
     Attributes(
       UserId = item(AttributeNames.userId),
       MembershipNumber = item.get(AttributeNames.membershipNumber).map(_.getS),
-      Tier = item(AttributeNames.tier)
+      Tier = item(AttributeNames.tier),
+      StartDate = item.get(AttributeNames.startDate).map(LocalDate.parse(_))
```
johnduffell: @svillafe this is the `.get` call needed to stop the exception, but sounds like Jacob's going to delete it anyway :1st_place_medal:

membership-attribute-service/app/controllers/AttributeController.scala
jacobwinch: :1st_place_medal:

membership-attribute-service/app/controllers/SalesforceHookController.scala
jacobwinch: Looks great 👍

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -40,99 +40,153 @@ class SalesforceHookController extends LazyLogging {
...
+          case None =>
+            Attributes(
+              UserId = membershipUpdate.UserId,
+              Tier = tierFromZuora,
+              MembershipNumber = membershipUpdate.MembershipNumber,
+              AdFree = None,
+              CardExpirationMonth = None,
+              CardExpirationYear = None,
+              StartDate = Some(startDate)
+            )
```
pvighi: isn't there a nice way to turn this Option(expMonth,expYear) into Option(expmonth), Option(exyear) and eliminate the duplication ?

johnduffell: thanks, good point!

johnduffell:
@aware as per jacob's comments we should use the new endpoint to refresh everything from zuora if we're updating things

membership-attribute-service/app/controllers/SalesforceHookController.scala
jacobwinch: Good idea, Salesforce retries anyway if there is a failure, so we are better off waiting and trying again. Data quality is more important than performance here 👍

johnduffell:
This is tested and ready to go now - if anyone wants to try to get it out while it's still conflict free that would be great.

We are still needing a backfill for everyone, at least the people who joined just over a year ago for a start.  We could get all their identity ids and run them through @svillafe 's new endpoint one at a time as a cheating fix.

If it's still here next Tuesday, I'll merge it myself! cc @lmath  @paulbrown1982

paulbrown1982:
Agreed, we should merge this ASAP. I'll merge it just before noon today, after I've chatted to @svillafe if I can.

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -18,13 +18,15 @@ object ContentAccess {
...
-                       CardExpirationYear: Option[Int] = None,
-                       ContributionFrequency: Option[String] = None) {
+  UserId: String,
+  Tier: Option[String] = None,
+  MembershipNumber: Option[String],
+  AdFree: Option[Boolean] = None,
+  CardExpirationMonth: Option[Int] = None,
+  CardExpirationYear: Option[Int] = None,
+  ContributionFrequency: Option[String] = None,
+  MembershipJoinDate: Option[LocalDate] // TODO startDate shouldn't be optional once we've backfilled it everywhere
```
svillafe: MembershipJoinDate, should be optional since a person can be only a Contributor but not a Member. I would add a default value of None.

paulbrown1982: I think that argument holds for MembershipNumber too.

svillafe:
After reading @paulbrown1982 last commit, it looks good! 👍

membership-attribute-service/app/models/Attributes.scala
johnduffell: thanks, good point, I didn't think about that as part of the merge.  Since some of the fields are optional together rather than individually, we should consider having an Option[Membership] and Option[Contribution] which have all the fields inside as required.

johnduffell:
thanks for merging @paulbrown1982  and thanks for the comments @svillafe

johnduffell:
interestingly, we used to store joinDate in the beginning, and it was removed temporarily with no explanation https://github.com/guardian/members-data-api/pull/6/commits/aa2c06755be5d418b05f7fcfeaf68f15f37bcd4f

PR #198, opened 2017-06-22
Support for storing card expiry dates for both membership and recurring contributions

paulbrown1982:
Updated the Attributes and Features models to contain a grouping of cards rather than just one, as an Attributes record now relates to multiple Zuora subscriptions/billing accounts.

I've also used this opportunity to tweak the /features endpoint to list which product has an expired card and what the expiry date was. This means we no longer output non-expired card expiry dates via that endpoint (a tiny bit safer).

There are a few TODOs dotted around which explain how we'll merge the Salesforce hook with the /update-attributes endpoint, so that there's only one code path updating Dynamo.

cc @svillafe @pvighi @lmath @johnduffell @rupertbates @rtyley

### DIFF
--- a/membership-attribute-service/app/controllers/SalesforceHookController.scala
+++ b/membership-attribute-service/app/controllers/SalesforceHookController.scala
@@ -6,7 +6,7 @@ import com.gu.memsub.subsv2.SubscriptionPlan
 import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.typesafe.scalalogging.LazyLogging
-import models.{ApiErrors, Attributes}
+import models.{ApiErrors, Attributes, CardDetails, Wallet}
 import monitoring.Metrics
 import parsers.Salesforce._
 import parsers.{Salesforce => SFParser}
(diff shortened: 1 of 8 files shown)


membership-attribute-service/app/models/Attributes.scala
```diff
@@ -19,13 +17,35 @@ object ContentAccess {
...
+
 case class Attributes(
   UserId: String,
   Tier: Option[String] = None,
   MembershipNumber: Option[String] = None,
   AdFree: Option[Boolean] = None,
-  CardExpirationMonth: Option[Int] = None,
-  CardExpirationYear: Option[Int] = None,
+  Wallet: Option[Wallet] = None,
   ContributionPaymentPlan: Option[String] = None,
```
svillafe: In order to be consistent with names, probably we should rename this to `recurringContributionPayment Plan`.

membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -51,8 +49,7 @@ class ScanamoAttributeService(client: AmazonDynamoDBAsyncClient, table: String)
       scanamoSetOpt('Tier, attributes.Tier),
       scanamoSetOpt('MembershipNumber -> attributes.MembershipNumber),
       scanamoSetOpt('ContributionPaymentPlan -> attributes.ContributionPaymentPlan),
```
svillafe: Should we also change the name of the field in Dynamo?

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -19,14 +17,36 @@ object ContentAccess {
   implicit val jsWrite = Json.writes[ContentAccess]
 }
 
+case class CardDetails(last4: String, expirationMonth: Int, expirationYear: Int, forProduct: String) {
+  val asLocalDate = new LocalDate(expirationYear, expirationMonth, 1).plusMonths(1).minusDays(1)
```
davidfurey: I'd have a slight preference for this to be a `def` rather than a `val` (same on line 31)

rtyley: Oh! My preference runs the other way! I'd probably go with `val` by default - seeing as the value never changes no matter how many times you call it, so it _is_ constant and doesn't _need_ to be a `def`. If calculating the value was expensive (not the case here) I might go for a `lazy val`.

davidfurey: My main reason for preferring a def here is because this `case class` is used for json serialisation.  If the month/year do not represent a valid date then json deserialisation will fail with an exception and I'm not sure this will be surfaced in an informative way.

rtyley: Oh, json deserialisation's an interesting one, I guess that might be a good reason. I'm not sure how tolerant I'd want to be of invalid date formats tho', perhaps deserialisation _should_ fail if the date is invalid ¯\_(ツ)_/¯

membership-attribute-service/app/controllers/SalesforceHookController.scala
```diff
@@ -118,33 +118,27 @@ class SalesforceHookController extends LazyLogging {
...
+        val walletF = for {
           account <- OptionT(touchpoint.zuoraService.getAccount(membershipSubscription.accountId).map(Option(_)))
           paymentMethodId <- OptionT(Future.successful(account.defaultPaymentMethodId))
           paymentMethod <- OptionT(touchpoint.zuoraService.getPaymentMethod(paymentMethodId).map(Option(_)))
           customerToken <- OptionT(Future.successful(paymentMethod.secondTokenId))
           stripeCustomer <- OptionT(touchpoint.stripeService.Customer.read(customerToken).map(Option(_)))
         } yield {
-          (stripeCustomer.card.exp_month, stripeCustomer.card.exp_year)
-        }).run
+          Wallet(membershipCard = Some(CardDetails.fromStripeCard(stripeCustomer.card, Membership.id)))
```
johnduffell: forProduct should really be a Product for as long as possible i.e. `Membership` rather than `Membership.id`

PR #201, opened 2017-06-28
template for PRs

johnduffell:
so lots of repos have PR templates.  I'll come clean - I don't like them, and I just delete them and write whatever I want :face_with_head_bandage: 
However PR descriptions are an important form of communication that it's important to use effectively.
PRs are as varied as people, but we're doing them for a reason (OKRs hopefully!)
Maybe this would be a useful checklist/reminder so we know we're explaining ourselves properly.

@guardian/membership-and-subscriptions comments please? 💣

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+<!-- 
+The text you're about to write is documentation for why the change is needed.
+Think about OKRs!
+-->
+# Existing situation
+
+# Description of the change
+
+# Benefit towards KR or other metric
+
+# Misc (trello card, screenshot, json, related PRs)
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,11 +1,11 @@
 <!-- 
-The text you're about to write is documentation for why the change is needed.
+The text you're about to write will advocate why the change is needed.
 Think about OKRs!
 -->
-# Existing situation
+### Existing situation
 
-# Description of the change
+### Description of the change
 
-# Benefit towards KR or other metric
+### Benefit towards KR or other metric
 
-# Misc (trello card, screenshot, json, related PRs)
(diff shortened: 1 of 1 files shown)


Ap0c:
Nice! I like this idea 🐼, we've been [using one](https://github.com/guardian/support-frontend/blob/master/.github/PULL_REQUEST_TEMPLATE.md) on support-frontend for a while and it's been helpful.

In terms of comments, only thing that occurs to me: is it possible 'Existing situation' and 'Description of the change' might often overlap?

jacobwinch:
👍 for the idea. 

I also agree that 'Existing situation' and 'Description of the change' could probably just be merged into a single category. Keeping the template as brief as possible will discourage people from following your lead and deleting it completely :stuck_out_tongue_winking_eye:

johnduffell:
thanks, I chatted with Jamie in person, the description was supposed to be more of a technical description wheras existing situation was supposed to be more of an okr level.
I'm going to change the second field to be more like the membership-frontend one

rtyley:
The thing I want to see most of all in a PR is an answer to 'Why are you doing this?' (often the 'why' gets forgotten, and the PR only details the 'what are you doing', duplicating the information you can already see in the diff!) which is why we put that question at the top of the PR template for [membership-frontend](https://github.com/guardian/membership-frontend/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

You've covered several possible types of 'why?' in this PR template, so I guess you're achieving a similar thing, tho' sometimes questions like `What problem are people seeing?` won't be entirely appropriate, eg if you're adding a new feature - also I'm not sure if asking people to write about their stuff in relation to KRs meet lead to unnecessary verbiage.

johnduffell:
thanks for the comments, I guess I'm trying to be more explicit that most things have a higher purpose, and I've noticed the "why" question sometimes yields a low level explanation.  I'll give it another tweak as suggested and move things to the comment

johnduffell:
see how it goes for a while, and please feel free to do more PRs on this template if you don't think it's working for the team (if it's not working for yourself, then do what I do 😆  )

PR #202, opened 2017-06-29
Added a new error type in the APIErros Object.

svillafe:
## Why do we need this? 
Currently in the case of an invalid API key for the sync attributes endpoint, we are returning a plain text saying `Invalid API key`. This PR change that by returning a JSON. We keep returning `HTTP Status code 403`.

### Old response:
**HTTP Status Code**  403
``` 
Invalid API key
```

### New response
**HTTP Status Code**  403

``` JSON
{
    "message": "Forbidden",
    "details": "Invalid API key",
    "statusCode": 403
}
```

## The changes 
* New API Error
* Changed the returning error for `/user-attributes/:user_id`

## Related PRs:
https://github.com/guardian/members-data-api/pull/191
https://github.com/guardian/members-data-api/pull/200

### DIFF
--- a/membership-attribute-service/app/models/ApiErrors.scala
+++ b/membership-attribute-service/app/models/ApiErrors.scala
@@ -34,4 +34,10 @@ object ApiErrors {
     details = "Failed to authenticate",
     statusCode = 401
   )
+
+  val forbidden = ApiError(
+    message = "Forbidden",
+    details = "Invalid API key",
+    statusCode = 403
+  )
 }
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/models/ApiErrors.scala
```diff
@@ -34,4 +34,10 @@ object ApiErrors {
     details = "Failed to authenticate",
     statusCode = 401
   )
+
+  val forbidden = ApiError(
```
johnduffell: should be `val invalidApiKey` as forbidden is just the code used not the meaning

johnduffell:
looks good, small comment about the naming.  I'm not sure why we require a message and a detail and what the difference is!
👍

PR #203, opened 2017-07-04
log much more detail when things fail in mma

johnduffell:
### Why do we need this? 
At present, when users complain that the profile tab doesn't work, we can only find out from the logs that they made a request at a certain time, and whether it worked or not.  We can't find the detail of what went wrong in any case other than payment details wrong (credit card didn't validate etc).  There are a couple of places where it would spit out a separate message saying things failed.
This means we have to poke around in zuora etc to find some wrong looking data when we get a report.
This PR changes it so that any line in the for comprehension will return a proper message if it fails, so we know where things went wrong.

I introduced an artificial failure on a test user and this appeared in the logs
`WARN: Unable to retrieve payment details result for identity user Some(30000594) due to couldn't read sub from zuora for crmId 001g000001h0T50AAE (TODO check the separate WARN to find out why)`

@jacobwinch @paulbrown1982 @pvighi @lmath @michaelwmcnamara @AWare

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -1,26 +1,28 @@
 package controllers
+import actions._
 import play.api.libs.concurrent.Execution.Implicits._
 import services.{AuthenticationService, IdentityAuthService}
 import com.gu.memsub._
-import json.PaymentCardUpdateResultWriters._
+import com.gu.memsub.subsv2.{Subscription, SubscriptionPlan}
+import com.gu.memsub.subsv2.reads.ChargeListReads._
+import com.gu.memsub.subsv2.reads.SubPlanReads
+import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.services.model.PaymentDetails
+import com.typesafe.scalalogging.LazyLogging
 import configuration.Config
+import json.PaymentCardUpdateResultWriters._
+import models.AccountDetails._
 import models.ApiErrors._
 import play.api.data.Form
 import play.api.data.Forms._
 import play.api.libs.json.Json
 import play.api.mvc.Results._
 import play.filters.cors.CORSActionBuilder
-import com.gu.memsub.subsv2.reads.SubPlanReads._
-import com.gu.memsub.subsv2.reads.ChargeListReads._
-import com.gu.memsub.subsv2.SubscriptionPlan
-import scalaz.std.scalaFuture._
+
 import scala.concurrent.Future
-import models.AccountDetails._
-import scalaz.OptionT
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -35,34 +37,38 @@ class AccountController extends LazyLogging {
...
+      user <- EitherT(Future.successful(maybeUserId \/> "no identity cookie for user"))
+      contact <- EitherT(request.touchpoint.contactRepo.get(user).map(_ \/> s"couldn't read contact from SF for $user (TODO check the separate ERROR to find out why)"))
+      sub <- EitherT(request.touchpoint.subService.either[F, P](contact).map(_ \/> s"couldn't read sub from zuora for crmId ${contact.salesforceAccountId} (TODO check the separate WARN to find out why)"))
+      details <- EitherT(request.touchpoint.paymentService.paymentDetails(sub).map[\/[String, PaymentDetails]](\/.right))
+    } yield (contact, details).toResult).run.map {
+      case \/-(result) => {
+        logger.info(s"Successfully retrieved payment details result for identity user: $maybeUserId")
+        result
+      }
+      case -\/(message) => {
```
johnduffell: this also covers the case is they somehow turn up at the manage page without being logged in.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -35,34 +37,38 @@ class AccountController extends LazyLogging {
     val tp = request.touchpoint
 
     (for {
-      user <- OptionT(Future.successful(authenticationService.userId))
-      sfUser <- OptionT(tp.contactRepo.get(user))
-      subscription <- OptionT(tp.subService.current[P](sfUser).map(_.headOption))
-      stripeCardToken <- OptionT(Future.successful(updateForm.bindFromRequest().value))
-      updateResult <- OptionT(tp.paymentService.setPaymentCardWithStripeToken(subscription.accountId, stripeCardToken))
+      user <- EitherT(Future.successful(authenticationService.userId \/> "no identity cookie for user"))
+      sfUser <- EitherT(tp.contactRepo.get(user).map(_ \/> s"couldn't read contact from SF for $user (TODO check the separate ERROR to find out why)"))
```
johnduffell: in membership common, we actually log some errors and then return a None.  It would make more sense to let the caller (i.e. us) deal with the error and log everything in one message.  I'd [already put "todos" there](https://github.com/guardian/membership-common/blob/8d579378aa1449040338553fd84a52cbe2825520/src/main/scala/com/gu/salesforce/ContactRepository.scala#L52-L58), as I had realised that was not right.

PR #205, opened 2017-07-13
Endpoint to lookup attributes based on calls to Zuora

lmath:
### Why do we need this? 
We'd like to explore the viability of doing lookups via Zuora (without accessing the Membership Attributes dynamo table). I prefer to split this into two PRs: this one to introduce the endpoint and another to establish how it will be used to test the viability of doing lookups via zuora at all (i.e. calling the endpoint, and being able to control/modify how much it gets called). 

### The changes 
- `GET /user-attributes/zuora-lookup` should match the response of `/user-attributes/me` except minus the membership number (this comes from salesforce)
- This is done by querying zuora to get accountids, then looking up subscriptions for each account id returned. With that information, we can build up the attributes. 

Relies on [this PR in membership common](https://github.com/guardian/membership-common/pull/505/)

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -1,9 +1,11 @@
 package controllers
 import _root_.services.{AuthenticationService, IdentityAuthService}
 import actions._
+import com.gu.memsub.Subscription.AccountId
 import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.memsub.subsv2.{Subscription, SubscriptionPlan}
+import com.gu.zuora.ZuoraRestService.{QueryResponse, RestSubscriptions}
 import com.typesafe.scalalogging.LazyLogging
 import configuration.Config
 import configuration.Config.authentication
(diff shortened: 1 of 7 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -63,6 +69,42 @@ class AttributeController extends Controller with LazyLogging {
...
+    attributes.run.map(_.toOption).map(_.getOrElse(None))
+  }
+
+  private def zuoraLookup(endpointDescription: String) =
+    backendAction.async { request =>
+      authenticationService.userId(request) match {
+        case Some(id) =>
+          attributesFromZuora(id, request.touchpoint.zuoraRestService, request.touchpoint.subService).map {
+            case Some(attrs) =>
+              logger.info(s"$id is a contributor - $endpointDescription - $attrs")
```
jacobwinch: As the user could be a member or a contributor (or both!), it might be better to have something like: 

s"Successfully retrieved attributes from Zuora for user $id: $attrs"

jacobwinch:
Definitely a good choice to split this into 2 PRs. I've made one small comment about logging, but overall this looks great. 👍 (pending membership-common release)

membership-attribute-service/app/controllers/AttributeController.scala
lmath: Thanks; I will fix that.

PR #206, opened 2017-07-17
Minor fixes (proposal)

kant:
### Why do we need this? 
just for the sake of semantics


johnduffell:
I'd merge it, but the build seems to be stuck so the merging is blocked, and it's not going to move the dials on anything, thanks anyway

PR #209, opened 2017-07-27
Send some /features traffic to lookup via Zuora

lmath:
Returns user attributes based on calls to Zuora on calls to /features some percentage of the time (that percentage is stored in a new dynamo table) 

([PR 205](https://github.com/guardian/members-data-api/pull/205) added a way to lookup a user's attributes by identityid just based on calls to Zuora)

### Why do we need this? 
By maintaining membership attribute data in dynamo, we 
* Need to maintain the data and ensure it is up to date (and staying up to date is currently dependent on the Salesforce hook)
* We will need to ensure that storing this data is compliant in the GDPR future

We are going to run a test where on calls to /features we gradually increase traffic that gets a user's attributes based on calls to Zuora. Once we know what percentage that is, we can have a better idea of the caching strategy to use. Perhaps we just use the fastly cache and when to invalidate it, or we'll need to decide on a new caching strategy that's easier to keep up to date. 

### The changes 
* There is a zuora lookup traffic percentage updated on a schedule (every 30 seconds) based on a value stored in a new MembershipFeatureToggles dynamo table. In this case, it stores the percentage of how much traffic should currently get attributes via calls to Zuora. If we will keep this table after the test, I will update the cloudformation.
* /features now determines where to get attributes from (MembershipAttributes table or Zuora) based on that percentage in the dynamo feature toggles table. 


paulbrown1982:
I'd consider just replacing the old lookup() (in the same position of the code to make the diff nice!) with the new lookupWithTest() implementation and provide a flag which forces the test to return false and use the touchpoint.attrService. That'll save the duplication of the `case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _))`code.

johnduffell:
in the "Why do we need this" section you wrote what the change actually is.  It would be useful to explain briefly why we're pursuing this course of action (i.e. not having to maintain the code to read/write dynamo, not having to control the data in gdpr future, easier invalidation, possibly just needing to purge the http cache instead of rewriting dynamo)  This will also help anyone looking at this in a couple of years when they're trying to understand the change.

membership-attribute-service/app/prodtest/Allocator.scala
```diff
@@ -0,0 +1,14 @@
+package prodtest
+
+class Allocator {
```
johnduffell: this doesn't need to be a class as it has no state, it can be an object

membership-attribute-service/app/prodtest/Allocator.scala
```diff
@@ -0,0 +1,14 @@
...
+
+  def isInTest(identityId: String, percentageInTest: Double): Boolean = {
+    if(percentageInTest > 0) {
```
johnduffell: I wouldn't bother checking that as it will work out later in the comparison...

membership-attribute-service/app/prodtest/Allocator.scala
```diff
@@ -0,0 +1,14 @@
...
+      val cleaned = identityId.replaceFirst("^0+", "").toInt
+      val index = cleaned % 100
+
+      index <= percentageInTest
```
johnduffell: ...however this should be < rather than <= (as `index` will be between 0-99 includive)

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -49,71 +48,89 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
   lazy val authenticationService: AuthenticationService = IdentityAuthService
   lazy val metrics = Metrics("AttributesController")
 
+  private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessMemberAndOrContributor: Attributes => Result, onNotFound: Result, doExperiment: Boolean) =
+  {
+    def pickAttributes(identityId: String, prodTest: Boolean)(implicit request: BackendRequest[AnyContent]) = {
```
johnduffell: since this is inside `lookup` we still have access to the `doExperiment` parameter, so no need to pass it in.  Although it doesn't apply, even if we needed to pass it it, I'd recommend keeping the name the same to avoid confusion.

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -49,71 +48,89 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
+      prodTest match {
```
johnduffell: pattern match on a boolean is neater as an if (...)else statement
two nested pattern match on a boolean with the same false expression is an if (... && ...) else

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -49,71 +48,89 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
+        case true =>
+          val percentageInTest = request.touchpoint.scheduledUpdateVariables.getPercentageTrafficForZuoraLookup
+          request.touchpoint.testAllocator.isInTest(identityId, percentageInTest) match {
+            case true =>
+              attributesFromZuora(identityId, request.touchpoint.zuoraRestService, request.touchpoint.subService)
```
johnduffell: we should probably have some logging to tell kibana which path we are following - if there's not already enough

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -49,71 +48,89 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
+      import loghandling.StopWatch
+      val stopWatch = new StopWatch
 
-        maybeSubs map { subs =>
-          if (subs.isEmpty) \/.left(s"Error getting subscriptions for identityId $identityId") else \/.right(subs)
+      result.map { res: Disjunction[String, R] =>
+        val latency = stopWatch.elapsed
+        val customFields: List[LogField] = List("zuora_latency_millis" -> latency.toInt, "zuora_call" -> whichCall, "identityId" -> identityId)
+        res match {
+          case -\/(a) => logErrorWithCustomFields(s"$whichCall failed. Msg: ${a}", customFields)
```
johnduffell: look good with the logging, just be careful of logging error level as we can log PII to sentry by mistake (warn level only goes to kibana)

johnduffell: also I'm not sure if we need to deal with failed Future here or not?

membership-attribute-service/app/prodtest/VariablesUpdatedOnSchedule.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.gu.memsub.util.ScheduledTask
+import com.typesafe.scalalogging.LazyLogging
+import services.ScanamoFeatureToggleService
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.DurationInt
+
+class VariablesUpdatedOnSchedule(featureToggleService: ScanamoFeatureToggleService)(implicit ec: ExecutionContext, system: ActorSystem) extends LazyLogging {
+
+  private var percentageTrafficForZuoraLookup = 0
```
johnduffell: generally it's safe to assume that `var` should never be used in scala

membership-attribute-service/app/prodtest/VariablesUpdatedOnSchedule.scala
```diff
@@ -0,0 +1,33 @@
...
+        case None =>
+          logger.info("Tried to update the percentage of traffic to do lookups via zuora, but that feature was not " +
+            "found in the table. Setting traffic to 0%")
+          0
+      }
+
+    }
+  updateZuoraTrafficPercentageTask.start()
+
+  def getPercentageTrafficForZuoraLookup = percentageTrafficForZuoraLookup
```
johnduffell: there's an agent in the super class that keeps the value - an agent is a threadsafe way of maintaining a value (ebven though in reality there should be only one thread, and there's no dependency of the written value on the read/previous value. and int is fundementally atomic for reads and writes.

membership-attribute-service/app/prodtest/VariablesUpdatedOnSchedule.scala
johnduffell: that's a roundabout way of saying that there's usually another way to do it that doesn't involve mutable state.  But better to do your own reading about var in scala to make your own mind up!

lmath: It just wasn't obvious to me how to avoid a mutable variable once I needed a globally accessible value that changes on a schedule. 😅

membership-attribute-service/app/services/ScanamoFeatureToggleService.scala
```diff
@@ -0,0 +1,25 @@
...
+
+class ScanamoFeatureToggleService(client: AmazonDynamoDBAsyncClient, table: String) extends LazyLogging {
+
+  private val scanamo = Table[FeatureToggle](table)
+  def run[T] = ScanamoAsync.exec[T](client) _
+
+  def get(featureName: String): Future[Option[FeatureToggle]] =
+    run(scanamo.get('FeatureName -> featureName).map(_.flatMap {
+      _
+        .left.map(e => logger.warn(s"Scanamo error in get: ${DynamoReadError.describe(e)}", e))
```
johnduffell: I'd rather return the error in a left and let someone upstream log it, as this pattern is the root of a few problems we had in membership-common.  also we separate the error from its context, so we don't know what it was trying to do at the time, we have to find two warnings at a similar time in the logs

membership-attribute-service/conf/touchpoint.DEV.conf
```diff
@@ -2,6 +2,7 @@ touchpoint.backend.environments {
   DEV {
     dynamodb.table = MembershipAttributes-DEV
     behaviour.dynamodb.table = MembershipBehaviour-DEV
+    featureToggles.dynamodb.table = MembershipFeatureToggles-DEV
```
johnduffell: dynamodb is a nice way to do this as it's easy to edit in AWS console compared with S3.

membership-attribute-service/app/prodtest/VariablesUpdatedOnSchedule.scala
lmath: A rather fundamental piece of the puzzle I was entirely missing. Thank you!

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -49,71 +49,89 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
   lazy val authenticationService: AuthenticationService = IdentityAuthService
   lazy val metrics = Metrics("AttributesController")
 
+  private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessMemberAndOrContributor: Attributes => Result, onNotFound: Result, doExperiment: Boolean) =
+  {
+    def pickAttributes(identityId: String) (implicit request: BackendRequest[AnyContent]) = {
+      if(doExperiment){
```
johnduffell: change doExperiment to endpointEligibleForTest or similar

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -49,71 +49,89 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
+      import loghandling.StopWatch
+      val stopWatch = new StopWatch
 
-        result.map { res: Disjunction[String, R] =>
-          val latency = stopWatch.elapsed
-          val customFields: List[LogField] = List("zuora_latency_millis" -> latency.toInt, "zuora_call" -> whichCall, "identityId" -> identityId)
-          res match {
-            case -\/(a) => logErrorWithCustomFields(s"$whichCall failed. Msg: ${a}", customFields)
-            case \/-(_) => logInfoWithCustomFields(s"$whichCall took ${latency}ms", customFields)
+      result.map { res: Disjunction[String, R] =>
```
johnduffell: I think this is existing code but normally I'd avoid using two synonyms in the same code for different variables.
in this case it could be 
```scala
futureResult.map { result: Disjunction[String, R] =>
```

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -49,71 +49,89 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
+        val latency = stopWatch.elapsed
+        val customFields: List[LogField] = List("zuora_latency_millis" -> latency.toInt, "zuora_call" -> whichCall, "identityId" -> identityId)
+        res match {
+          case -\/(a) => {
```
johnduffell: `a` can be `message` or `error`

membership-attribute-service/app/prodtest/Allocator.scala
```diff
@@ -0,0 +1,12 @@
+package prodtest
+
+object Allocator {
+
+  def isInTest(identityId: String, percentageInTest: Double): Boolean = {
+    val cleaned = identityId.replaceFirst("^0+", "").toInt
```
johnduffell: can identityId ever be an invalid int? and if so will we end up with the correct response?

membership-attribute-service/app/prodtest/FeatureToggleDataUpdatedOnSchedule.scala
```diff
@@ -0,0 +1,38 @@
...
+           s"found in the table. Setting traffic to 0%. Error: $e")
+         updateZuoraTrafficPercentageTask.agent.alter(0)
+         0
+    }
+
+  }
+
+  updateZuoraTrafficPercentageTask.start()
+
+  def getPercentageTrafficForZuoraLookup = updateZuoraTrafficPercentageTask.agent.get()
```
johnduffell: I think scheduledtask itself has a .get method rather than having to get from the agent directly.  In fact FeatureToggleDataUpdatedOnSchedule doesn't really have any of its own state, so you could make it an object instead of a class, and just return the ScheduledTask to the caller rather than hiding it.  Then they can call get on the scheduled task themselves which will save on code a bit.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -63,6 +65,7 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   lazy val contactRepo = new SimpleContactRepository(tpConfig.salesforce, system.scheduler, Config.applicationName)
   lazy val attrService: AttributeService = new ScanamoAttributeService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoAttributesTable)
   lazy val behaviourService: BehaviourService = new ScanamoBehaviourService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoBehaviourTable)
+  lazy val featureToggleService: ScanamoFeatureToggleService = new ScanamoFeatureToggleService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoFeatureToggleTable)
```
johnduffell: really, the TouchpointComponents class is some kind of epic class of all knowledge about the world.  unfortunately this kind of omniscient class is a bit of an antipattern in java, and more so in scala, so ideally I'd like to avoid adding to it.  Basically everything depends on it, and it depends on everything.

In this specific case, the only thing that might have state (some kind of cached connection?) is the AmazonDynamoDBAsyncClient, so that's the only thing we need to maintain a reference to.  The rest can all be objects containing functions but no state, which are referenced directly when they are needed in the code.

johnduffell:
👍 go for it! looking forward to getting some results for this one! 🚀

PR #210, opened 2017-07-31
Add policy to access MembershipFeaturesToggles dynamo table

lmath:
### Why do we need this? 
In [PR 209 - send some traffic through Zuora](https://github.com/guardian/members-data-api/pull/209) we need to be able to get a percentage stored in the MembershipFeaturesToggles dynamo table to determine how much traffic to send via the Zuora lookup. But I did not include the cloudformation change to allow get from the MembershipFeaturesToggles table and so I'm doing that here. 
 
### The changes 

- Adds MembershipFeaturesToggles dynamo table + policy to the cloudformation

[PR 209 - send some traffic through Zuora](https://github.com/guardian/members-data-api/pull/209)


lmath:
I updated the cloudformation on the stack and will merge this first thing tomorrow.

PR #212, opened 2017-08-01
Better attributes controller logging

svillafe:
### Why do we need this? 

Currently, the update operation in dynamo fails we are not logging why. This PR introduces some logging.

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -27,8 +27,11 @@ import services.{AttributesMaker, AuthenticationService, IdentityAuthService}
 import scala.concurrent.Future
 import scalaz.std.scalaFuture._
 import scalaz.syntax.std.option._
+
 import prodtest.Allocator._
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -182,7 +185,7 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
           RecurringContributionPaymentPlan = conSub.map(_.plan.name),
           MembershipJoinDate = memSub.map(_.startDate)
         )
-        res <- EitherT(tp.attrService.update(attributes).map(\/.right))
+        res <- EitherT(tp.attrService.update(attributes).map(_.disjunction)).leftMap(e => s"Dynamo failed to update the user attributes: $e")
```
lmath: If you get a left as a DynamoReadError, I think you want `DynamoReadError.describe(e)` ?

svillafe: Good catch @lmath, I will include this in a future PR.

PR #213, opened 2017-08-01
Using DynamoReadError.describe

svillafe:
### Why do we need this? 

In a previous PR: https://github.com/guardian/members-data-api/pull/212 @lmath suggested using `DynamoReadError.describe` instead of `toString`. This PR adds it.


salesforce-test.xml
```diff
@@ -0,0 +1,20 @@
+<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
AWare: Not sure what this file is doing here.

svillafe: Good catch @AWare, deleted in the final version of the PR.

johnduffell:
👍 great!

PR #214, opened 2017-08-02
Better logging when getting subs from Zuora

lmath:
### Why do we need this? 
[PR 205](https://github.com/guardian/members-data-api/pull/205) added code to do lookups via Zuora. However, when getting subscriptions both an error when calling Zuora and the user literally having no subscriptions resulted in an empty list. 

Now that we want to [actually send some traffic](https://github.com/guardian/members-data-api/pull/209) through this lookup, it has become apparent that I need to be able to identify when calls to Zuora are failing, and log an error (so that I know to reduce the traffic). Since a user may really not have any subs, then I should just log with info in this case. I initially mistakenly logged both cases as an error.

### The changes 
* log an error when calls to get subscriptions from Zuora fail
* log info when a user has no subs

[PR 205 - lookups via Zuora](https://github.com/guardian/members-data-api/pull/205)
[PR 507 requires the client to handle errors when getting subs from Zuora](https://github.com/guardian/membership-common/pull/507)
[PR 209 - send some traffic to zuora lookup](https://github.com/guardian/members-data-api/pull/209)

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -32,6 +32,7 @@ import prodtest.Allocator._
 
 import scalaz.{-\/, Disjunction, DisjunctionT, EitherT, \/, \/-}
 import scalaz.syntax.std.either._
+import scalaz._, std.list._, syntax.traverse._
 
 
 class AttributeController extends Controller with LoggingWithLogstashFields {
@@ -101,7 +102,6 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
         val customFields: List[LogField] = List("zuora_latency_millis" -> latency.toInt, "zuora_call" -> whichCall, "identityId" -> identityId)
         result match {
           case -\/(messageOrError) => {
-            log.error(s"Attempted $whichCall but then: ${messageOrError}")
             logWarnWithCustomFields(s"$whichCall failed with ${messageOrError}", customFields)
           }
           case \/-(_) => logInfoWithCustomFields(s"$whichCall took ${latency}ms", customFields)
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -115,13 +115,20 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
-        if (subs.isEmpty) \/.left(s"Error getting subscriptions for identityId $identityId") else \/.right(subs)
+      val maybeSubs: Future[Disjunction[String, List[Subscription[AnyPlan]]]] = accountIds.traverseU(id => sub(id)).map(_.sequenceU.map(_.flatten))
+
+      maybeSubs.map {
+        _.leftMap { errorMsg =>
+          log.logger.error(s"We tried getting a subscription for a user, but then ${errorMsg}")
+          errorMsg
+        } map { subs =>
+          if(subs.isEmpty)
+            log.logger.info(s"We tried getting subs for user with identityId ${identityId} but there weren't any")
```
jacobwinch: Should these just be log.error / log.info?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -115,13 +115,20 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
-      }
-      val maybeSubs: Future[List[Subscription[AnyPlan]]] = Future.traverse(accountIds)(id => sub(id)).map(_.flatten)
 
-      maybeSubs map { subs =>
-        if (subs.isEmpty) \/.left(s"Error getting subscriptions for identityId $identityId") else \/.right(subs)
+      val maybeSubs: Future[Disjunction[String, List[Subscription[AnyPlan]]]] = accountIds.traverseU(id => sub(id)).map(_.sequenceU.map(_.flatten))
+
+      maybeSubs.map {
+        _.leftMap { errorMsg =>
+          log.logger.error(s"We tried getting a subscription for a user, but then ${errorMsg}")
```
jacobwinch: We should log the identityId in this line too

jacobwinch:
A couple of small comments, but overall this looks good 👍

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: what's special about the empty case? why not log the number of subs returned (zero or more..)

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -115,13 +115,20 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
     def getSubscriptions(accountIds: List[AccountId]): Future[Disjunction[String, List[Subscription[AnyPlan]]]] = {
-      def sub(accountId: AccountId): Future[List[Subscription[AnyPlan]]] = {
+      def sub(accountId: AccountId): Future[Disjunction[String, List[Subscription[AnyPlan]]]] =
         subscriptionService.subscriptionsForAccountId[AnyPlan](accountId)(anyPlanReads)
-      }
-      val maybeSubs: Future[List[Subscription[AnyPlan]]] = Future.traverse(accountIds)(id => sub(id)).map(_.flatten)
 
-      maybeSubs map { subs =>
-        if (subs.isEmpty) \/.left(s"Error getting subscriptions for identityId $identityId") else \/.right(subs)
+      val maybeSubs: Future[Disjunction[String, List[Subscription[AnyPlan]]]] = accountIds.traverseU(id => sub(id)).map(_.sequenceU.map(_.flatten))
```
johnduffell: so we will only end up with any data at all if they all worked (assuming the sequenceU turns any Left in the list into a Left?)

johnduffell: I think previously we would have flattened all the empty lists together then just lost any "erroring" ones.  This may not be a problem we actually have, but it's exactly what problem I had with [members-data-api](https://trello.com/c/6tfTRjJl/245-error-message-viewing-digital-pack-or-membership-tabs-in-mma)

lmath: Thanks for highlighting this. As I increase the traffic through the zuora lookup, I will watch out to see if this becomes an issue.

membership-attribute-service/app/controllers/AttributeController.scala
lmath: Good point @jacobwinch @johnduffell. I will take your advice.

membership-attribute-service/app/controllers/AttributeController.scala
lmath: Good idea.

johnduffell:
👍 great, thanks!

PR #215, opened 2017-08-04
Fix logging for timed GetSubscriptions calls to Zuora

lmath:
### Why do we need this? 
image
To stop logging that a call to zuora takes 0ms when that call did not in fact happen
### The changes 
- Remove responsibility for error logs from 'withTimer' as we only want to error log if the call to Zuora has failed. Error log right after a call to Zuora if the call returns a left instead. 
- If getSubscriptions gets an empty list of accounts, return a left (this ensures that the for comprehension will not log timed subscriptions calls to Zuora that don't happen)

[PR 207 - log latency on calls to Zuora](https://github.com/guardian/members-data-api/pull/207)
[PR 209 - send some /features traffic to lookup via zuora](https://github.com/guardian/members-data-api/pull/209)

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -152,7 +152,7 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
       _.leftMap { errorMsg =>
         log.info(s"Tried to get Attributes for ${identityId} but failed with $errorMsg")
         errorMsg
-      }.fold(_ => None, identity) //.getOrElse(None)
+      }.fold(_ => None, identity)
     }
   }
 
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -114,23 +114,21 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
     def queryToAccountIds(response: QueryResponse): List[AccountId] =  response.records.map(_.Id)
 
     def getSubscriptions(accountIds: List[AccountId]): Future[Disjunction[String, List[Subscription[AnyPlan]]]] = {
-      if(accountIds.isEmpty) Future.successful(\/.left(s"Can't get subscriptions for $identityId because they have no accountIds"))
-      else {
-        def sub(accountId: AccountId): Future[Disjunction[String, List[Subscription[AnyPlan]]]] =
-          subscriptionService.subscriptionsForAccountId[AnyPlan](accountId)(anyPlanReads)
 
-        val maybeSubs: Future[Disjunction[String, List[Subscription[AnyPlan]]]] = accountIds.traverseU(id => sub(id)).map(_.sequenceU.map(_.flatten))
+      def sub(accountId: AccountId): Future[Disjunction[String, List[Subscription[AnyPlan]]]] =
+        subscriptionService.subscriptionsForAccountId[AnyPlan](accountId)(anyPlanReads)
 
-        maybeSubs.map {
-          _.leftMap { errorMsg =>
-            log.error(s"We tried getting a subscription for a user with identityId ${identityId}, but then ${errorMsg}")
-            errorMsg
-          } map { subs =>
-            log.info(s"We got subs for identityId ${identityId} from Zuora and there were ${subs.length}")
-            subs
-          }
+      val maybeSubs: Future[Disjunction[String, List[Subscription[AnyPlan]]]] = accountIds.traverseU(id => sub(id)).map(_.sequenceU.map(_.flatten))
+      maybeSubs.map {
+        _.leftMap { errorMsg =>
+          log.error(s"We tried getting a subscription for a user with identityId ${identityId}, but then ${errorMsg}")
+          errorMsg
+        } map { subs =>
+          log.info(s"We got subs for identityId ${identityId} from Zuora and there were ${subs.length}")
+          subs
         }
       }
+
     }
 
     def zuoraAccountsQuery(identityId: String): Future[Disjunction[String, QueryResponse]] = zuoraRestService.getAccounts(identityId).map {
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -115,24 +114,34 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
+      _.leftMap { error =>
+        log.error(s"Calling ZuoraAccountIdsFromIdentityId failed for identityId ${identityId}. Error: ${error}")
+        error
+      }
+    }
+
+    val attributes = for {
+      accounts <- EitherT[Future, String, QueryResponse](withTimer(s"ZuoraAccountIdsFromIdentityId", zuoraAccountsQuery(identityId)))
       accountIds = queryToAccountIds(accounts)
       subscriptions <- EitherT[Future, String, List[Subscription[AnyPlan]]](withTimer(s"ZuoraGetSubscriptions", getSubscriptions(accountIds)))
```
johnduffell: could have done 
```scala
subscriptions <- EitherT[Future, String, List[Subscription[AnyPlan]]](
  if (accountIds.nonEmpty) Future.successful(\/.right(Nil))
  else withTimer(s"ZuoraGetSubscriptions", getSubscriptions(accountIds)))
```

also it won't let me comment on the code, but ideally on 151 we could log the error
```scala
attributes.run.map(_.toOption).map(_.getOrElse(None))
```
could be
```scala
attributes.run.leftMap(err => logger.error(s"failed with: $error")).fold(_ => None, _)
```

lmath: Good to know on checking accountIds.nonEmpty in the for comprehension. I was trying to figure that out earlier but I think I like having this check in `getSubscriptions` now. 

Good point about logging the error when going from String \/ Option[Attributes] to Option[Attributes]! I shall do that.

johnduffell: if an empty list is not acceptable for getSubscriptions, perhaps it would be better to change the type of the parameter to a scalaz.NonEmptyList just so it's clearer to anyone reusing it.  I'm not sure I'd expect this behaviour otherwise.

if on the other hand, getsubscriptions is using an either to signal back to the calling code whether it should be timing it, I think it's not really down to getsubscriptions to make that decision, it's better up to the thing doing the timing to work what it wants to time.
also it would need to be an either of either, as we want to know if something has failed, so we still need the "$whichCall failed with <error>"

all the extra 0 or 1 ms in the total may not be affecting the figures too much in reality, when the actual calls are adding ~250ms each time, unless we're getting (say) 95+% of calls  with nothing in the list?

lmath: OK, I buy your argument about checking for an empty list in the for comprehension instead. 🙂  

The idea of getting subscriptions from an empty list seemed nonsense to me (timers aside) but I think handling that when it is called seems sensible.

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -97,14 +97,13 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
-        val latency = stopWatch.elapsed
-        val customFields: List[LogField] = List("zuora_latency_millis" -> latency.toInt, "zuora_call" -> whichCall, "identityId" -> identityId)
-        result match {
-          case -\/(messageOrError) => {
-            logWarnWithCustomFields(s"$whichCall failed with ${messageOrError}", customFields)
-          }
-          case \/-(_) => logInfoWithCustomFields(s"$whichCall took ${latency}ms", customFields)
+      futureResult.map { disjunction: Disjunction[String, R] =>
+        disjunction match {
+          case -\/(message) => log.info(message)
```
johnduffell: perhaps better to make this warn with some further explanation attached
`log.warn(s"$whichCall failed with ${message}")`

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -129,17 +128,34 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
         }
       }
+
     }
 
-    val attributes: DisjunctionT[Future, String, Option[Attributes]] = for {
-      accounts <- EitherT(withTimer(s"ZuoraAccountIdsFromIdentityId", zuoraRestService.getAccounts(identityId)))
+    def zuoraAccountsQuery(identityId: String): Future[Disjunction[String, QueryResponse]] = zuoraRestService.getAccounts(identityId).map {
+      _.leftMap { error =>
+        log.error(s"Calling ZuoraAccountIdsFromIdentityId failed for identityId ${identityId}. Error: ${error}")
```
johnduffell: error goes to sentry, so we just have to be careful in case some PII comes through in the error message (warn will only go to kibana and cloudwatch which is ok)
in this case I wouldn't log anything until it gets back to the Controller, just change to
```scala
_.leftMap { error =>
   s"Calling ZuoraAccountIdsFromIdentityId failed for identityId ${identityId}. Error: ${error}")
}
```

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -129,17 +128,34 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
+      )
+
     } yield {
       AttributesMaker.attributes(identityId, subscriptions, LocalDate.now())
     }
 
-    attributes.run.map(_.toOption).map(_.getOrElse(None))
+    attributes.run.map {
+      _.leftMap { errorMsg =>
+        log.info(s"Tried to get Attributes for ${identityId} but failed with $errorMsg")
```
johnduffell: same for here, I'd use the pattern above to add to the error message and pass it on to the caller to decide what to do

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: it will be logged on line 155 anyway

paulbrown1982:
👍 LGTM nice work in making the logging happen intelligently/appropriately.

johnduffell:
agreed, good work getting this through! 👍

PR #216, opened 2017-08-09
In lookups via Zuora, get the ad free flag from dynamo anyway

lmath:
### Why do we need this? 
[There is a bug in the test to send a percentage of traffic through Zuora](https://github.com/guardian/members-data-api/pull/209) where all users have adFree = false if their /features call gets its data via Zuora. Currently the adfree flag is not kept in Zuora so we need to get it from the Dynamo table. If we don't get the AdFree state from Dynamo, then some users (currently staff) will see ads when they are not expecting to. 

### The changes 

- Fetch attributes from Dynamo anyway, but only use the state of the adFree flag when returning attributes from the Zuora lookup

[PR 209 -- send some /features traffic through Zuora](https://github.com/guardian/members-data-api/pull/209)


AWare:
The sad saga of members data api :( 👍

PR #217, opened 2017-08-09
Fix MMA to work even if you have another product.

johnduffell:
### Why do we need this? 
This is a health PR to fix a problem where mma endpoint for digipack returned an error if the user had a membership etc.  And vice versa.  The call is indeed failing, but fail to read a sub should just be treated as it is the wrong type.

### The changes 
Use the new type which can report error or optionality.
https://github.com/guardian/membership-common/pull/509

@paulbrown1982 @lmath @pvighi @jacobwinch

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -20,7 +20,7 @@ object Dependencies {
   val awsDynamo = "com.amazonaws" % "aws-java-sdk-dynamodb" % awsClientVersion
   val awsSNS = "com.amazonaws" % "aws-java-sdk-sns" % awsClientVersion
   val awsCloudWatch = "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsClientVersion
-  val membershipCommon = "com.gu" %% "membership-common" % "0.433"
+  val membershipCommon = "com.gu" %% "membership-common" % "0.435-SNAPSHOT"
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -20,7 +20,7 @@ object Dependencies {
   val awsDynamo = "com.amazonaws" % "aws-java-sdk-dynamodb" % awsClientVersion
   val awsSNS = "com.amazonaws" % "aws-java-sdk-sns" % awsClientVersion
   val awsCloudWatch = "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsClientVersion
-  val membershipCommon = "com.gu" %% "membership-common" % "0.433"
+  val membershipCommon = "com.gu" %% "membership-common" % "0.435-SNAPSHOT"
```
johnduffell: https://github.com/guardian/membership-common/pull/509

AWare:
Looks good :+1: 
This should make things easier in the long run too,

johnduffell:
> check that it returns the right subs when user has digipack and membership

https://github.com/guardian/membership-common/blob/766c977fe5feb95065f757d152e3f15c91b4dd2a/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionService.scala#L44

turns out that's already in place!  good thought though.

paulbrown1982:
👍 My accounts work now!

PR #218, opened 2017-08-10
Run sbt dependency vulnerability check on teamcity builds

johnduffell:
### Why do we need this? 
We are very good at keeping our AMIs up to date for known problems/vulnerabilities there (amigo/amiable).   However libraries that we use in prod also may have vulnerabilities, that could be targeted more specifically.  This becomes more dangerous the more backend systems that system has access to, and that danger could apply until the next time credentials are rotated.

There is a [common list of vulnerabilities](https://cve.mitre.org/data/downloads/index.html) known in various libraries, which could be used by attackers and developers alike.  Ideally we want to be ahead of the attackers so should use it to prompt us to keep things up to date.

The downside is this just tells you when libraries are vulnerable, but doesn't prove that any of the affected functions are actually exposed in such a way to cause a risk in our case.

The plugin produces a nice [html report](https://teamcity-aws.gutools.co.uk/repository/download/memsub_membership_membersdataapi/77023:id/membership-attribute-service-dependency-check-report.html) but you can also fail the build on severity.  Unfortunately we have a 9.3 so we can't fail the build yet as that's the highest rating.

The first time I ran this it took 2 minutes as it was downloading the vulnerability files.  I reran it and it took under 20 seconds.  there maybe a way to run it in parallel.

### The changes 
Basically add the plugin to SBT and tell it to copy the files into the build output.

https://github.com/guardian/janus/pull/796

In the end, it's only useful to have this if we're going to keep things clean (i.e. keep deps up to date).  And ideally run it on all our projects. Any thoughts? @paulbrown1982 @adamnfish  @guardian/membership-and-subscriptions

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,8 +64,6 @@ val buildDebSettings = Seq(
   )
 )
 
-dependencyCheckFailBuildOnCVSS := 10 // we have a 9.3, decrese this when possible
-
 def lib(name: String) = Project(name, file(name))
   .enablePlugins(PlayScala, BuildInfoPlugin, RiffRaffArtifact, JDebPackaging).settings(commonSettings)
 
@@ -80,7 +78,7 @@ val api = app("membership-attribute-service")
   .settings(
     addCommandAlias("devrun", "run 9400"),
     addCommandAlias("batch-load", "runMain BatchLoader"),
-    addCommandAlias("play-artifact", ";dependencyCheck;riffRaffNotifyTeamcity")
+    addCommandAlias("play-artifact", "riffRaffNotifyTeamcity")
   )
 
 val root = project.in(file(".")).aggregate(api)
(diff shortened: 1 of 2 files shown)


adamnfish:
Probably better to "warn on failure" for now, to raise awareness of the problems you have. If you're going to leave it at `10` and not direct people's attention to the result then all you've done is extend your build times for now.
Also seems a bit odd to add dependency check to the play artifact command. Would it be better included in `test`? Alternatively, you could just add it from TeamCity.

This is a great start, nice one!

johnduffell:
the visibility piece is still missing, adam mentioned a central dashboard, but more simply we just need some kind of plugin for teamcity or sbt to push a comment onto the originating PR stating the vulnerability level and/or pointing to the report.  Until that's present, I don't see the benefit of shipping this (as per adam's awareness point above)

johnduffell:
other things are going on at present to address this requirement in general, so I'm closing this as it's not needed in its current form

PR #219, opened 2017-08-14
Add a how many concurrent calls to zuora counter + logging this value

lmath:
### Why do we need this? 
[PR 209 in members-data-api ](https://github.com/guardian/members-data-api/pull/209) sends some traffic through /features to do a lookup via calls to Zuora. Once we switch the percentage higher, it should be useful to know some information about how many concurrent calls we are making to Zuora.

Originally, I wanted to add this in membership-common (see: [PR 512- add a how many current calls to Zuora counter)](https://github.com/guardian/membership-common/pull/512) but if during a call to Zuora a future failed, the counter may be incremented and then never decremented. 

### The changes 
- Adds a counter for how many calls we have made to Zuora that decrements on response or failed future

### DIFF
--- a/membership-attribute-service/app/loghandling/ZuoraRequestCounter.scala
+++ b/membership-attribute-service/app/loghandling/ZuoraRequestCounter.scala
@@ -1,0 +1,15 @@
+package loghandling
+
+import java.util.concurrent.atomic.AtomicInteger
+
+import com.typesafe.scalalogging.LazyLogging
+
+object ZuoraRequestCounter extends LazyLogging{
+  private val counter = new AtomicInteger()
+
+  def increment: Int = counter.incrementAndGet()
+
+  def decrement: Int = counter.decrementAndGet()
+
+  def get = counter.get()
+}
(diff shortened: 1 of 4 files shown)


membership-attribute-service/app/loghandling/ZuoraRequestCounter.scala
```diff
@@ -0,0 +1,15 @@
+package loghandling
+
+import java.util.concurrent.atomic.AtomicInteger
+
+import com.typesafe.scalalogging.LazyLogging
+
+object ZuoraRequestCounter extends LazyLogging{
+  private val counter = new AtomicInteger()
+
+  def increment: Int = counter.incrementAndGet()
```
johnduffell: for neatness, maybe just get rid of these convenience methods, and move the `withZuoraRequestCounter` into this class instead.

johnduffell:
great, looks good, just a small bit of potential simplification but I think this will work well! 🎉 🚀 👍

PR #221, opened 2017-08-16
Add /me and /me/membership to the Zuora lookup test

lmath:
### Why do we need this? 
We are exploring moving away from maintaining subscription info in dynamo. So, we are running a test where we send an increasing percentage of traffic for lookups via Zuora (instead of dynamo).

We did this test for /me/features but this looks to make up only about a quarter of the requests. So we will dial up the traffic for all the lookups and not just /me/features

### The changes 

- Adds /me and /me/membership to the Zuora lookup test (they will direct a percentage of traffic to lookup via zuora -- the same percentage for all the endpoints)

[pr 205 - endpoint to lookup via zuora](https://github.com/guardian/members-data-api/pull/205)
[pr 209 - send some /features traffic to lookup via zuora](https://github.com/guardian/members-data-api/pull/209)

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -196,8 +196,8 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
        .getOrElse(notFound)
   }
 
-  def membership = lookup("membership", onSuccessMember = membershipAttributesFromAttributes, onSuccessMemberAndOrContributor = _ => notAMember, onNotFound = notFound, endpointEligibleForTest = false)
-  def attributes = lookup("attributes", onSuccessMember = identity[Attributes], onSuccessMemberAndOrContributor = identity[Attributes], onNotFound = notFound, endpointEligibleForTest = false)
+  def membership = lookup("membership", onSuccessMember = membershipAttributesFromAttributes, onSuccessMemberAndOrContributor = _ => notAMember, onNotFound = notFound, endpointEligibleForTest = true)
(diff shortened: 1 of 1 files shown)


paulbrown1982:
👍 Just remember to scale down the % to something low before deploying then ramp up again slowly, as this change may route 200% more traffic down the new code path.

johnduffell:
this was a one hour block (although an app alert went out in that hour which causes an android spike)

url | count | platform
/user-attributes/me/membership | 26747 | android
/user-attributes/me | 10604 | ios
/user-attributes/me/features | 8477 | dotcom

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -196,8 +196,8 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
        .getOrElse(notFound)
   }
 
-  def membership = lookup("membership", onSuccessMember = membershipAttributesFromAttributes, onSuccessMemberAndOrContributor = _ => notAMember, onNotFound = notFound, endpointEligibleForTest = false)
-  def attributes = lookup("attributes", onSuccessMember = identity[Attributes], onSuccessMemberAndOrContributor = identity[Attributes], onNotFound = notFound, endpointEligibleForTest = false)
+  def membership = lookup("membership", onSuccessMember = membershipAttributesFromAttributes, onSuccessMemberAndOrContributor = _ => notAMember, onNotFound = notFound, endpointEligibleForTest = true)
```
johnduffell: technically we don't need the boolean at all, because false isn't used

johnduffell:
great, thanks! 👍

PR #222, opened 2017-08-22
Use a longer timeout for Zuora queries to get account ids

lmath:
### Why do we need this? 
We're looking at moving away from maintaining membership attribute info in Dynamo and getting it from Zuora. But we might overload Zuora so we're dialing up a percentage of traffic that does lookups this way.

At 10% traffic we started seeing socket timeout exceptions during the daily Zuora bill run.

There is a call in particular that tends to timeout during the bill run, the identityId -> accountIds query. We are willing to wait longer so [pr 517 in membership common](https://github.com/guardian/membership-common/pull/517) moves this call into its own rest service that can have a longer timeout here in members-data-api.

### The changes 
- When we do lookups via Zuora, we need to make a query to get accountIds from identityId. This change increases the timeout for that call to 30 seconds. 

[pr 517 - move zuora calls that can have longer timeouts to a PatientZuoraRestService](https://github.com/guardian/membership-common/pull/517)
[pr221 members-data-api: Zuora lookup test for all the lookup endpoints](https://github.com/guardian/members-data-api/pull/221)

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -16,11 +16,10 @@ import com.gu.stripe.StripeService
 import com.gu.touchpoint.TouchpointBackendConfig
 import com.gu.zuora.rest.SimpleClient
 import com.gu.zuora.soap.ClientWithFeatureSupplier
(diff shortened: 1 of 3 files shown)


johnduffell:
might be simpler if we follow https://github.com/guardian/membership-common/pull/517/files#r134511027

PR #223, opened 2017-08-25
Yet more logging: did attributes come from zuora or dynamo? + what stage for feature toggle logging

lmath:
### Why do we need this? 
We're testing out sending lookup (/me, /me/features, /me/membership) to get attributes information from Zuora. As I was increasing the traffic looking up via Zuora, I noticed that if I looked at a particular instance then the percentage of traffic seemed to be toggling between what I set it to and 0. This is probably just the percentage for UAT, but to be sure, more logging is required!

This also seemed like a sensible time to add a log on each lookup call for if the information came via Zuora or via Dynamo.
 
### The changes 

- When the percentage for the test is updated, log which stage that update is for
- On {/me, /me/features, /me/membership} calls, log if the attributes came from Zuora or Dynamo
 

johnduffell:
looks great! go for it! 👍 🚀

paulbrown1982:
👍 LGTM

PR #224, opened 2017-08-25
Use dynamo client builder instead of deprecated constructor in touchpoint backend

lmath:
I just looked at that deprecation warning enough times to crack.

### Why do we need this? 
Moving away from deprecated constructors must be a good thing. 

### The changes 
- In the TouchpointBackend, create a dynamo db client using the builder instead of the constructor, which is deprecated

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -2,7 +2,7 @@ package components
 
 import akka.actor.ActorSystem
 import com.amazonaws.regions.Regions
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClient
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClientBuilder
 
 import scalaz.std.scalaFuture._
 import com.gu.config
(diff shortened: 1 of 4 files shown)


membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -63,9 +63,16 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
     RequestRunners.loggingRunner(metrics("zuora-soap"))
   )
   lazy val contactRepo = new SimpleContactRepository(tpConfig.salesforce, system.scheduler, Config.applicationName)
-  lazy val attrService: AttributeService = new ScanamoAttributeService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoAttributesTable)
-  lazy val behaviourService: BehaviourService = new ScanamoBehaviourService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoBehaviourTable)
-  lazy val featureToggleService: ScanamoFeatureToggleService = new ScanamoFeatureToggleService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoFeatureToggleTable)
```
lmath: Is there some reason why we need a new instance of the client for each service?

lmath:
We also need to create a dynamo client using the builder elsewhere in members-data-api, so might as well do these together another time.

PR #225, opened 2017-08-30
Preserve the value of the adfree flag on calls to /features for contributors

lmath:
### Why do we need this? 
If a logged in user is a contributor, then for the moment, we should return the adFree flag we have 
stored for them in dynamo on calls to /features.

I initially suggested just populating the fields we have available, but since /features is deprecated, better to keep to its existing behaviour as much as possible.

### The changes 
- Preserve value of the adfree flag on calls to /features even for contributors

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -57,7 +57,7 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
   lazy val authenticationService: AuthenticationService = IdentityAuthService
   lazy val metrics = Metrics("AttributesController")
 
-  private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessMemberAndOrContributor: Attributes => Result, onNotFound: Result, endpointEligibleForTest: Boolean) =
+  private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessMemberAndOrContributor: Attributes => Result, onNotFound: Result, endpointEligibleForTest: Boolean): Action[AnyContent] =
   {
     def pickAttributes(identityId: String) (implicit request: BackendRequest[AnyContent]): (String, Future[Option[Attributes]]) = {
       if(endpointEligibleForTest){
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/models/Features.scala
```diff
@@ -33,6 +33,11 @@ object Features {
   }
 
   val unauthenticated = Features(None, adFree = false, adblockMessage = true, None, None, None)
+
+  def notAMember(attributes: Attributes) = {
+    val adFree = attributes.AdFree.getOrElse(false)
+    Features(None, adFree = adFree, adblockMessage = true, None, None, None)
```
johnduffell: not a big deal but you can do `unauthenticated.copy(adFree = adFree)` to save defining out the other fields again

johnduffell:
good debugging!  with 3 tests/changeovers on the go (ad free, direct-to-zuora and fulfiling contributions from here) it's getting difficult to keep out of trouble 😨 
🚀 👍

membership-attribute-service/app/models/Features.scala
lmath: Oh right! I'll do that. Thanks.

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -199,7 +199,7 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
 
   def membership = lookup("membership", onSuccessMember = membershipAttributesFromAttributes, onSuccessMemberAndOrContributor = _ => notAMember, onNotFound = notFound, endpointEligibleForTest = true)
   def attributes = lookup("attributes", onSuccessMember = identity[Attributes], onSuccessMemberAndOrContributor = identity[Attributes], onNotFound = notFound, endpointEligibleForTest = true)
-  def features = lookup("features", onSuccessMember = Features.fromAttributes, onSuccessMemberAndOrContributor = _ => Features.unauthenticated, onNotFound = Features.unauthenticated, endpointEligibleForTest = true)
+  def features = lookup("features", onSuccessMember = Features.fromAttributes, onSuccessMemberAndOrContributor = Features.notAMember, onNotFound = Features.unauthenticated, endpointEligibleForTest = true)
```
svillafe: My only concern is if we double check with everybody who uses this endpoint can handle this new response.

JustinPinner: I'd hope they wouldn't fall apart if a new element is present in the unauthenticated response :-/

JustinPinner:
Great work, thanks @lmath. The sooner we can wire this up properly the better, but now we can invite outsiders in to try it out when we're ready.

membership-attribute-service/app/controllers/AttributeController.scala
lmath: Before, the /features response for a contributor was always:
```
{
  "adFree": false,
  "adblockMessage": true
}
```

And now it's just possible for adFree to be true

paulbrown1982:
👍 Good spot of this bug.

PR #226, opened 2017-09-04
On lookups via Zuora, compare with the attributes from Dynamo and log if different

lmath:
### Why do we need this? 
We're running a test to do membership attribute lookups by calling Zuora for subscription info rather than maintaining the data in the membership attributes dynamo table. A big part of why we were interested getting this information from Zuora in the first place was to ensure we are returning more up to date data. 

We are doing a dynamo lookup to get the adfree flag anyway, so I think it would be useful to compare the attributes in Dynamo with the attributes created with Zuora lookups. If they are different, then this is likely an instance where Dynamo is out of date and we are returning more accurate data by finding the data via Zuora. 

I would like to be able to count when this occurs so I'm logging it.

### The changes 
- On lookups via Zuora, since we query Dynamo anyway to get the adfree anyway, we now compare the attributes from Dynamo with the attributes determined from calling Zuora. 

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -142,6 +142,16 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
       }
     }
 
+    def compareThenLogAttributes(attributesFromDynamo: Future[Option[Attributes]], attributesFromZuora: Future[Option[Attributes]]): Unit = {
+      attributesFromDynamo map { dynamoAttributes =>
+        attributesFromZuora map { zuoraAttributes =>
+          val zuoraAttributesWithAdfree = zuoraAttributes map { _.copy(AdFree = dynamoAttributes flatMap(_.AdFree))}
+          if (zuoraAttributesWithAdfree != dynamoAttributes)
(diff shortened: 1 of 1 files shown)


jacobwinch:
Nice idea - I'm expecting Dynamo to be wrong a significant number of times 👍

paulbrown1982:
👍 awesome

johnduffell:
good idea! consider loging when it's right as well as wrong, then you can express the result as a percentage (although it probably looks better to say "10000 incorrect responses blocked" rather than "0.01% of requests would have been wrong" so it depends how you want to pitch it!)

lmath:
Thanks @johnduffell.

My thinking was:
1) lookups completed via Zuora = count of log lines with "... $attributes found via Zuora"  
2) lookups where Dynamo was probably outdated = count of log lines with "Zuora and Dynamo disagreed"

lookups where dynamo and dynamo agreed/dynamo was right = 1) - 2)

But that's just because I imagined I just cared about the overall count. I can imagine it might be useful to be able to filter on successful lookups and examine them further so I think I'll try putting that in and see what I can do with it.

johnduffell:
fair enough, sounds good as it is!

PR #227, opened 2017-09-05
Allow Members Data API to authenticate and authorise iPad Daily Edition content

paulbrown1982:
### Why do we need this? 

The iPad Daily Edition app is changing to use Identity as its authentication, and as a result it can switch to use Members Data API rather than CAS to authorise that the app can download today's daily edition.

### The changes 
- Added support for digital subscriptions in the /user-attributes/me endpoint.
- The JSON now returns the latest expiry date of whatever digital plans the user is signed up to, and a boolean in the contentAccess JSON as to whether they should be able to access digital content (Daily Edition / Premium Tier).
- I've done some D.R.Y refactoring of AttributesMaker after adding the third search/filter concern.
- Added some more detailed logging in AttributesController as to what principal subscription type has been identified.

Example response to `https://members-data-api.theguardian.com/user-attributes/me`:

```
{
  "userId": "123456",
  "digitalSubscriptionExpiryDate":"2100-01-01",
  "contentAccess": {
    "member": false,
    "paidMember": false,
    "recurringContributor": false,
    "digitalPack": true
  }
}
```

https://trello.com/c/PyjUgof8/266-members-data-api-me-endpoint-to-show-whether-the-identity-user-has-a-digital-subscription

cc @johnduffell @pvighi @lmath @AWare @aodhol

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -80,17 +80,17 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
                 "X-Gu-Membership-Tier" -> tier,
                 "X-Gu-Membership-Is-Paid-Tier" -> attrs.isPaidTier.toString
               )
-            case Some(attrs @ Attributes(_, _, _, _, _, _, _, Some(date))) =>
-              log.info(s"$identityId is a digital subscriber expiring $date - $endpointDescription - found via $fromWhere")
-              onSuccessSupporter(attrs)
-            case Some(attrs @ Attributes(_, _, _, _, _, Some(paymentPlan), _, _)) =>
-              log.info(s"$identityId is a regular $paymentPlan contributor - $endpointDescription - found via $fromWhere")
-              onSuccessSupporter(attrs)
-            case Some(attrs @ Attributes(_, _, _, Some(true), _, _, _, _)) =>
-              log.info(s"$identityId is an ad-free reader - $endpointDescription - found via $fromWhere")
-              onSuccessSupporter(attrs)
             case Some(attrs) =>
-              log.info(s"$identityId is some unknown kind of supporter - $endpointDescription - $attrs found via $fromWhere")
+              attrs.DigitalSubscriptionExpiryDate.foreach { date =>
+                log.info(s"$identityId is a digital subscriber expiring $date")
+              }
+              attrs.RecurringContributionPaymentPlan.foreach { paymentPlan =>
+                log.info(s"$identityId is a regular $paymentPlan contributor")
+              }
+              attrs.AdFree.foreach { _ =>
+                log.info(s"$identityId is an ad-free reader")
+              }
+              log.info(s"$identityId supports the guardian - $attrs - found via $fromWhere")
               onSuccessSupporter(attrs)
             case _ =>
               onNotFound
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -57,7 +57,7 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
   lazy val authenticationService: AuthenticationService = IdentityAuthService
   lazy val metrics = Metrics("AttributesController")
 
-  private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessMemberAndOrContributor: Attributes => Result, onNotFound: Result, endpointEligibleForTest: Boolean) =
+  private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessSupporter: Attributes => Result, onNotFound: Result, endpointEligibleForTest: Boolean) =
   {
     def pickAttributes(identityId: String) (implicit request: BackendRequest[AnyContent]): (String, Future[Option[Attributes]]) = {
       if(endpointEligibleForTest){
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -74,15 +74,24 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
-            case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _)) =>
-              log.info(s"$identityId is a member - $endpointDescription - $attrs found via $fromWhere")
+            case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _, _)) =>
+              log.info(s"$identityId is a $tier member - $endpointDescription - $attrs found via $fromWhere")
               onSuccessMember(attrs).withHeaders(
                 "X-Gu-Membership-Tier" -> tier,
                 "X-Gu-Membership-Is-Paid-Tier" -> attrs.isPaidTier.toString
               )
+            case Some(attrs @ Attributes(_, _, _, _, _, _, _, Some(date))) =>
+              log.info(s"$identityId is a digital subscriber expiring $date - $endpointDescription - found via $fromWhere")
```
lmath: Might it be worth logging $attrs in these cases still since someone could match more than one case?

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -60,8 +61,9 @@ case class Attributes(
   lazy val isPaidTier = isSupporterTier || isPartnerTier || isPatronTier || isStaffTier
   lazy val isAdFree = AdFree.exists(identity)
   lazy val isContributor = RecurringContributionPaymentPlan.isDefined
+  lazy val digitalSubscriberHasActivePlan = DigitalSubscriptionExpiryDate.exists(_.isAfter(now))
 
-  lazy val contentAccess = ContentAccess(member = isPaidTier || isFriendTier, paidMember = isPaidTier, recurringContributor = isContributor) // we want to include staff!
+  lazy val contentAccess = ContentAccess(member = isPaidTier || isFriendTier, paidMember = isPaidTier, recurringContributor = isContributor, digitalPack = digitalSubscriberHasActivePlan || isStaffTier)
```
lmath: Why does isStaffTier = true imply digitalPack should be true?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -74,15 +74,24 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
+              onSuccessSupporter(attrs)
```
johnduffell: just put the logging in the pattern match, and don't repeat the onSuccessSupporter(attrs) bit

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -1,30 +1,41 @@
...
 
 class AttributesMaker extends LazyLogging {
 
   def attributes(identityId: String, subs: List[Subscription[AnyPlan]], forDate: LocalDate): Option[Attributes] = {
 
-    val groupedSubs: Map[Option[Product], List[Subscription[AnyPlan]]] = subs.groupBy(subscription => GetCurrentPlans(subscription, forDate).toOption.map(_.head.product))
-    val membershipSub = groupedSubs.getOrElse(Some(Product.Membership), Nil)
-    val contributionSub = groupedSubs.getOrElse(Some(Product.Contribution), Nil)
+    def getCurrentPlans(subscription: Subscription[AnyPlan]) = GetCurrentPlans(subscription, forDate).toOption
+    def getTopProduct(subscription: Subscription[AnyPlan]) = getCurrentPlans(subscription).map(_.head.product)
```
johnduffell: not down to your change, but it's interesting that we're turning an error into an option (.toOption on the either) and an option into an exception (.head on a list).  Maybe subs with no plans are already filtered out, so .head won't fail.

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -1,30 +1,41 @@
...
+    def getTopBenefitId(subscription: Subscription[AnyPlan]) = getCurrentPlans(subscription).map(_.head.charges.benefits.head.id)
+    def getTopPlanName(subscription: Subscription[AnyPlan]) = getCurrentPlans(subscription).map(_.head.name)
+    def getAllBenefits(subscription: Subscription[AnyPlan]) = getCurrentPlans(subscription).toList.flatMap(_.map(_.charges.benefits.list).list.flatten).toSet
 
-    val tier = membershipSub.headOption.map(sub => GetCurrentPlans(sub, forDate).toOption.map(_.head.charges.benefits.head.id)).flatten
-    val recurringContributionPaymentPlan = contributionSub.headOption.map(sub => GetCurrentPlans(sub, forDate).toOption.map(_.head.name)).flatten
-    val membershipJoinDate = membershipSub.map(_.startDate).headOption
+    val groupedSubs: Map[Option[Product], List[Subscription[AnyPlan]]] = subs.groupBy(getTopProduct)
```
johnduffell: what does a key `None` mean? looks like subs that had an error while trying to read them.  does that make sense?

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -1,30 +1,41 @@
...
 
-    if(!membershipSub.isEmpty || !contributionSub.isEmpty)
+    val membershipSub = groupedSubs.getOrElse(Some(Product.Membership), Nil).headOption         // Assumes only 1 membership per Identity customer
+    val contributionSub = groupedSubs.getOrElse(Some(Product.Contribution), Nil).headOption     // Assumes only 1 regular contribution per Identity customer
+    val subsWhichIncludeDigitalPack = subs.filter(getAllBenefits(_).contains(Benefit.Digipack))
+
+    val tier = membershipSub.flatMap(getTopBenefitId)
+    val recurringContributionPaymentPlan = contributionSub.flatMap(getTopPlanName)
+    val membershipJoinDate = membershipSub.map(_.startDate)
+    val latestDigitalPackExpiryDate = subsWhichIncludeDigitalPack.map(_.termEndDate).sorted.reverse.headOption
```
johnduffell: fair enough.  for a longer list, max would have been more efficient

johnduffell:
yes looks great to me from a code point of view!  just a few various comments inline.

membership-attribute-service/app/controllers/AttributeController.scala
paulbrown1982: I'd meant to make this only digital subscriber, so I'll fix that and let the user with multiple products scenario fall down to the last case.

membership-attribute-service/app/models/Attributes.scala
paulbrown1982: I think it'd be nice if all staff got access to the Daily Edition, it saves us generating urgent override tokens. Also we can get more feedback to make the product even better! I'll get this signed off before the new app gets published on the Apple App Store.

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -31,22 +33,20 @@ class AttributesMaker extends LazyLogging {
...
+    val hasAttributableProduct = membershipSub.nonEmpty || contributionSub.nonEmpty || subsWhichIncludeDigitalPack.nonEmpty
+
+    hasAttributableProduct.option {
       val tier: Option[String] = membershipSub.flatMap(getCurrentPlans(_).headOption.map(_.charges.benefits.head.id))
       val recurringContributionPaymentPlan: Option[String] = contributionSub.flatMap(getTopPlanName)
       val membershipJoinDate: Option[LocalDate] = membershipSub.map(_.startDate)
-      val latestDigitalPackExpiryDate: Option[LocalDate] = subsWhichIncludeDigitalPack.map(_.termEndDate).sorted.reverse.headOption
-
-      Some(Attributes(
+      val latestDigitalPackExpiryDate: Option[LocalDate] = Some(subsWhichIncludeDigitalPack.map(_.termEndDate)).filter(_.nonEmpty).map(_.max)
```
johnduffell: clever!

paulbrown1982: Google!

membership-attribute-service/app/models/Attributes.scala
johnduffell: it'd be sensible if staff get everything that doesn't cost us to fulfil, but I don't know the policy on that.

johnduffell:
go for it 👍 🏇

lmath:
lgtm 👍

paulbrown1982:
I've tested in prod using an account with no other product and everything seems to be working OK. I've registered for a digital pack free trial and the endpoint now returns: 

```
{
  "userId": "123456789",
  "digitalSubscriptionExpiryDate": "2018-09-06",
  "contentAccess": {
    "member": false,
    "paidMember": false,
    "recurringContributor": false,
    "digitalPack": true
  }
}
```

PR #228, opened 2017-09-05
In Attributes comparison, just compare values that can be found via Zuora

lmath:
### Why do we need this? 
We're running a test to do membership attribute lookups by calling Zuora for subscription info rather than maintaining the data in the membership attributes dynamo table.

In [PR 226](https://github.com/guardian/members-data-api/pull/226) I compared attributes found via Zuora and attributes from Dynamo. 

However, there are some fields that will not be populated by the Zuora lookup:
1) AdFree - we query dynamo just to get this field in the Zuora lookup as Dynamo is currently the source of truth for that field 
2) Wallet - Not currently used yet, and relies on a call to Stripe
3) Membership number - Not actually required for the lookup endpoints and requires a call to Salesforce.

So, we will need to ignore these when logging the comparison for the comparison to be meaningful. 

### The changes 
- When logging the comparison between lookups via Zuora and what's in the Dynamo table, ignore the fields that are not populated by the Dynamo lookup. 

 [PR 226 - in lookups via Zuora log a comparison of the attributes](https://github.com/guardian/members-data-api/pull/226) 

cc @paulbrown1982 @johnduffell @jacobwinch @pvighi @AWare

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -143,11 +143,20 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
     }
 
     def compareThenLogAttributes(attributesFromDynamo: Future[Option[Attributes]], attributesFromZuora: Future[Option[Attributes]]): Unit = {
-      attributesFromDynamo map { dynamoAttributes =>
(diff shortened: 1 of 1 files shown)


paulbrown1982:
👍 LGTM

PR #231, opened 2017-09-11
Fixed how staff memberships enable the Digital Pack subscription

paulbrown1982:
### Why do we need this? 
Fixed how staff memberships enable the Digital Pack subscription. 

### The changes 
It's now impossible for contentAccess.digitalPack to be true if the digitalSubscriptionExpiryDate is absent. This was leading to a confusing ambiguity, see: https://github.com/guardian/members-data-api/pull/230#issuecomment-328131062

cc @aodhol @JustinPinner @johnduffell @pvighi @lmath @AWare

### DIFF
--- a/membership-attribute-service/app/services/AttributesMaker.scala
+++ b/membership-attribute-service/app/services/AttributesMaker.scala
@@ -7,6 +7,7 @@ import com.gu.memsub.{Benefit, Product}
 import com.typesafe.scalalogging.LazyLogging
 import models.Attributes
 import org.joda.time.LocalDate
+import org.joda.time.LocalDate.now
 
 import scalaz.syntax.std.boolean._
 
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -40,12 +41,13 @@ class AttributesMaker extends LazyLogging {
       val recurringContributionPaymentPlan: Option[String] = contributionSub.flatMap(getTopPlanName)
       val membershipJoinDate: Option[LocalDate] = membershipSub.map(_.startDate)
       val latestDigitalPackExpiryDate: Option[LocalDate] = Some(subsWhichIncludeDigitalPack.map(_.termEndDate)).filter(_.nonEmpty).map(_.max)
+      val staffExpiryDate: Option[LocalDate] = tier.exists(_.equalsIgnoreCase("staff")).option(now.plusDays(1))
```
johnduffell: so I guess we're telling it to call back in every day to check the expiry, I'm not expecting that to overload the servers though!

johnduffell: does tier really come in mixed case? I thought it was the product name which is "Staff"

johnduffell:
looks good to me! 👍

membership-attribute-service/app/services/AttributesMaker.scala
paulbrown1982: That expiry date is not actually cacheable anyway because the product can be cancelled and refunded before its term or paid-in-advance period ends.

PR #233, opened 2017-09-20
Fix staff tier giving digipack from dynamo

paulbrown1982:
### Why do we need this? 
For the Daily Edition - allow staff access

### The changes 

Moved the logic to override the DigitalSubscriptionExpiryDate to with… ……in the case class itself so that it works when the class is constructed from Zuora or from a serialised Staff membership or Digital Supporter representation.

cc @johnduffell @aodhol @lmath @pvighi


johnduffell:
fine, I still think we shouldn't put logic in the writers as it means we can't unit test it but I agree it's an existing place where both code paths happen to converge so it allows minimal refactoring.
👍

PR #234, opened 2017-09-20
Move Zuora lookup code out of the controller

lmath:
### Why do we need this? 
We initially did lookups directly from Zuora as a test, but now we are determining how to keep doing lookups via Zuora but cope with spikes of traffic. So, I think it's time to move the monster attributesFromZuora function out of the controller.

There's more refactoring left to do, but I think moving the code out of the controller is enough for a PR. 

### The changes 
- Move zuora lookup code out of the controller into a service
- Add some tests for identityId -> subscriptions

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -64,6 +64,7 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   )
   lazy val contactRepo = new SimpleContactRepository(tpConfig.salesforce, system.scheduler, Config.applicationName)
   lazy val attrService: AttributeService = new ScanamoAttributeService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoAttributesTable)
+  lazy val zuoraAttrService: ZuoraAttributeService = new ZuoraAttributeService(zuoraRestService, subService, attrService)
   lazy val behaviourService: BehaviourService = new ScanamoBehaviourService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoBehaviourTable)
   lazy val featureToggleService: ScanamoFeatureToggleService = new ScanamoFeatureToggleService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoFeatureToggleTable)
   lazy val snsGiraffeService: SNSGiraffeService = SNSGiraffeService(giraffeSns)
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -7,6 +7,7 @@ import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClient
 import scalaz.std.scalaFuture._
 import com.gu.config
 import com.gu.memsub.services.PaymentService
+import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
 import com.gu.memsub.subsv2.services.SubscriptionService.CatalogMap
 import com.gu.memsub.subsv2.services._
 import com.gu.monitoring.ServiceMetrics
@@ -64,7 +65,7 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   )
   lazy val contactRepo = new SimpleContactRepository(tpConfig.salesforce, system.scheduler, Config.applicationName)
   lazy val attrService: AttributeService = new ScanamoAttributeService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoAttributesTable)
-  lazy val zuoraAttrService: ZuoraAttributeService = new ZuoraAttributeService(zuoraRestService, subService, attrService)
+  lazy val zuoraAttrService: ZuoraAttributeService = new ZuoraAttributeService(accountId => zuoraRestService.getAccounts(accountId), accountId => reads => subService.subscriptionsForAccountId[AnyPlan](accountId)(reads), attrService)
   lazy val behaviourService: BehaviourService = new ScanamoBehaviourService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoBehaviourTable)
   lazy val featureToggleService: ScanamoFeatureToggleService = new ScanamoFeatureToggleService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoFeatureToggleTable)
   lazy val snsGiraffeService: SNSGiraffeService = SNSGiraffeService(giraffeSns)
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -1,15 +1,9 @@
 package controllers
 import actions._
-import com.gu.memsub.Subscription.AccountId
-import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
 import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads._
-import com.gu.memsub.subsv2.services.SubscriptionService
 import com.gu.memsub.subsv2.{Subscription, SubscriptionPlan}
 import com.gu.scanamo.error.DynamoReadError
-import com.gu.zuora.ZuoraRestService
-import com.gu.zuora.ZuoraRestService.QueryResponse
-import loghandling.ZuoraRequestCounter
 import configuration.Config
 import configuration.Config.authentication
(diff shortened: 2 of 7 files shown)


membership-attribute-service/app/services/ZuoraAttributeService.scala
```diff
@@ -0,0 +1,141 @@
...
+import scalaz.std.scalaFuture._
+import scalaz.syntax.traverse._
+import scalaz.{-\/, Disjunction, EitherT, \/, \/-, _}
+
+
+class ZuoraAttributeService(identityIdToAccountIds: String => Future[String \/ QueryResponse],
+                            subscriptionsForAccountId: AccountId => SubPlanReads[AnyPlan] => Future[Disjunction[String, List[Subscription[AnyPlan]]]],
+                            dynamoAttributeGetter: String => Future[Option[Attributes]]) extends LoggingWithLogstashFields {
+
+  def get(userId: String): Future[Option[Attributes]] = attributesFromZuora(userId)
```
johnduffell: get and attributesFromZuora are the same function, no need to have both?

johnduffell: regarding naming, since (from a dependencies point of view) we want to keep this code only doing the one thing, it might be worth removing the word "service" and calling it AttributesFromZuora.  Then people won't be tempted to add any other functions in the same file.
Ideally we could make this into an object instead of a class and avoid adding more to touchpointbackend, but that would need more refactoring.

membership-attribute-service/app/services/ZuoraAttributeService.scala
```diff
@@ -0,0 +1,141 @@
...
+    val attributes = attributesDisjunction.run.map {
+      _.leftMap { errorMsg =>
+        log.error(s"Tried to get Attributes for $identityId but failed with $errorMsg")
+        errorMsg
+      }.fold(_ => None, identity)
+    }
+
+    val attributesFromDynamo: Future[Option[Attributes]] = dynamoAttributeGetter(identityId)
+
+    dynamoAndZuoraAgree(attributesFromDynamo, attributes, identityId)
```
johnduffell: we get a Future back here, but we just lose the reference.  So if the future failed or hung, we wouldn't do anything about it.

membership-attribute-service/app/services/ZuoraAttributeService.scala
```diff
@@ -0,0 +1,141 @@
...
+      accountIds = queryToAccountIds(accounts)
+      subscriptions <- EitherT[Future, String, List[Subscription[AnyPlan]]](
+        if(accountIds.nonEmpty) withTimer(s"ZuoraGetSubscriptions", getSubscriptions(accountIds, identityId), identityId)
+        else Future.successful(\/.right {
+          log.info(s"User with identityId $identityId has no accountIds and thus no subscriptions.")
+          Nil
+        })
+      )
+    } yield {
+      AttributesMaker.attributes(identityId, subscriptions, LocalDate.now())
```
johnduffell: LocalDate.now() is another side effect that would need passing in before it's testable.  The amount of times we've had a test failing due to passing a date is quite high!

membership-attribute-service/app/services/ZuoraAttributeService.scala
```diff
@@ -0,0 +1,141 @@
...
+            zuoraAttributes.copy(
+              AdFree = dynamoAttributes.AdFree, //fetched from Dynamo in the Zuora lookup anyway (dynamo is the source of truth)
+              Wallet = dynamoAttributes.Wallet, //can't be found based on Zuora lookups, and not currently used
+              MembershipNumber = dynamoAttributes.MembershipNumber, //I don't think membership number is needed and it comes from Salesforce
+              MembershipJoinDate = dynamoAttributes.MembershipJoinDate.flatMap(_ => zuoraAttributes.MembershipJoinDate), //only compare if dynamo has value
+              DigitalSubscriptionExpiryDate = None
+            )
+          }
+        }
+        val dynamoAndZuoraAgree = zuoraAttributesWithIgnoredFields == maybeDynamoAttributes
```
johnduffell: interesting approach, I'd have probably done it by pattern matching the relevant fields from both and then just comparing them either in a tuple or just individually, otherwise you're comparing fields that you just set to the same.  However since it's hopefully temporary code it's not worth going into detail.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -64,6 +65,9 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   )
   lazy val contactRepo = new SimpleContactRepository(tpConfig.salesforce, system.scheduler, Config.applicationName)
   lazy val attrService: AttributeService = new ScanamoAttributeService(new AmazonDynamoDBAsyncClient(com.gu.aws.CredentialsProvider).withRegion(Regions.EU_WEST_1), dynamoAttributesTable)
+  lazy val zuoraAttrService: ZuoraAttributeService = new ZuoraAttributeService(accountId => zuoraRestService.getAccounts(accountId),
+    accountId => reads => subService.subscriptionsForAccountId[AnyPlan](accountId)(reads),
+    attrService.get)
```
pvighi: maybe we could use named params here to make it clearer

membership-attribute-service/app/services/ZuoraAttributeService.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  def get(userId: String): Future[Option[Attributes]] = attributesFromZuora(userId)
+
+  def attributesFromZuora(identityId: String): Future[Option[Attributes]] = {
+
+    val attributesDisjunction = for {
+      accounts <- EitherT[Future, String, QueryResponse](withTimer(s"ZuoraAccountIdsFromIdentityId", zuoraAccountsQuery(identityId, identityIdToAccountIds), identityId))
+      accountIds = queryToAccountIds(accounts)
+      subscriptions <- EitherT[Future, String, List[Subscription[AnyPlan]]](
+        if(accountIds.nonEmpty) withTimer(s"ZuoraGetSubscriptions", getSubscriptions(accountIds, identityId), identityId)
```
pvighi: is there any reason why we could not just call getSubscriptions and let it return empty if accountIds is empty?

lmath: I'm checking accountIds.nonEmpty because I've opted to wrap calls to Zuora in a timer and I wouldn't like to time and then log a call that doesn't happen

membership-attribute-service/app/components/TouchpointComponents.scala
lmath: Good idea. I will do this.

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -107,114 +105,11 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
...
-      }
-    }
-  }
-
   private def zuoraLookup(endpointDescription: String) =
     backendAction.async { implicit request =>
       authenticationService.userId(request) match {
         case Some(identityId) =>
-          attributesFromZuora(identityId, request.touchpoint.patientZuoraRestService, request.touchpoint.subService, request.touchpoint.attrService).map {
+          getAttributes(identityId, request.touchpoint.zuoraRestService.getAccounts, accountId => reads => request.touchpoint.subService.subscriptionsForAccountId[AnyPlan](accountId)(reads), request.touchpoint.attrService.get).map {
```
pvighi: This looks a lot like the getAttributes call in line 58, could we extract this to a shared def that takes identity id and request as parameter ? It might clean up this long line as well

lmath: I need to just totally remove that endpoint and that function but I couldn't decide if this PR was when I wanted to do it.

membership-attribute-service/app/services/ZuoraAttributeService.scala
lmath: Ah, right! I will log a message when the future fails then.

membership-attribute-service/app/services/ZuoraAttributeService.scala
lmath: re: get and attributesFromZuora
Agreed. I didn't mean to leave get behind so thanks for spotting that. 

re: an object AttributesFromZuora
I like this suggestion and I've tried to take it. Can you take another look?

membership-attribute-service/app/services/ZuoraAttributeService.scala
lmath: Well spotted. I'm on it; thanks.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -0,0 +1,144 @@
...
+import scalaz.syntax.traverse._
+import scalaz.{-\/, Disjunction, EitherT, \/, \/-, _}
+
+
+object AttributesFromZuora extends LoggingWithLogstashFields {
+
+  def getAttributes(identityId: String,
+                    identityIdToAccountIds: String => Future[String \/ QueryResponse],
+                    subscriptionsForAccountId: AccountId => SubPlanReads[AnyPlan] => Future[Disjunction[String, List[Subscription[AnyPlan]]]],
+                    dynamoAttributeGetter: String => Future[Option[Attributes]]): Future[Option[Attributes]] = {
```
johnduffell: great! this feels very self contained we could even reuse this in a new project without a touchpoint backend!
there are a fair few parameters but hopefully we won't need the dynamoAttributeGetter in the long run.

johnduffell:
good work this makes the code easier to maintain 👍 🚀

PR #235, opened 2017-09-22
Updated to use the latest membership-common

paulbrown1982:
### Why do we need this? 
Part of the Australia international charging change. See: https://github.com/guardian/membership-frontend/pull/1725

### The changes 
- Updated to use the latest [membership-common](https://github.com/guardian/membership-common/pull/534), which requires services to choose which Stripe account to connect with. Also made some possible deprecation notes and TODOs.
- Switched to using the touchpoint.paymentService in the Salesforce hook (this is kinda moot as this is likely going away). However the change is really there to support the AU cards in the AccountDetails page in Identity.

cc @johnduffell @desbo @lmath

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -56,8 +56,11 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   implicit lazy val _bt = tpConfig
   lazy val metrics = new ServiceMetrics(tpConfig.zuoraRest.envName, Config.applicationName,_: String)
 
-  lazy val stripeService = new StripeService(tpConfig.stripe, RequestRunners.loggingRunner(metrics("stripe")))
-  lazy val giraffeStripeService = new StripeService(tpConfig.giraffe, RequestRunners.loggingRunner(metrics("stripe")))
+  lazy val ukStripeService = new StripeService(tpConfig.stripeUKMembership, RequestRunners.loggingRunner(metrics("stripe")))
+  lazy val auStripeService = new StripeService(tpConfig.stripeAUMembership, RequestRunners.loggingRunner(metrics("stripe")))
+
+  lazy val ukContributionsStripeService = new StripeService(tpConfig.stripeUKContributions, RequestRunners.loggingRunner(metrics("stripe")))
+
   lazy val soapClient = new ClientWithFeatureSupplier(Set.empty, tpConfig.zuoraSoap,
     RequestRunners.loggingRunner(metrics("zuora-soap")),
     RequestRunners.loggingRunner(metrics("zuora-soap"))
@@ -76,7 +79,7 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   lazy val futureCatalog: Future[CatalogMap] = catalogService.catalog.map(_.fold[CatalogMap](error => {println(s"error: ${error.list.mkString}"); Map()}, _.map))
 
   lazy val subService = new SubscriptionService[Future](productIds, futureCatalog, simpleClient, zuoraService.getAccountIds)
-  lazy val paymentService = new PaymentService(stripeService, zuoraService, catalogService.unsafeCatalog.productMap)
+  lazy val paymentService = new PaymentService(zuoraService, catalogService.unsafeCatalog.productMap)
   lazy val featureToggleData = new FeatureToggleDataUpdatedOnSchedule(featureToggleService, stage)
 
 }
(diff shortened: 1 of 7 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -33,6 +33,9 @@ class AccountController extends LazyLogging {
   lazy val mmaCardAction = NoCacheAction andThen corsCardFilter andThen BackendFromCookieAction
 
   def updateCard[P <: SubscriptionPlan.AnyPlan : SubPlanReads] = mmaCardAction.async { implicit request =>
+
+    // TODO - refactor to use the Zuora-only based lookup, like in AttributeController.pickAttributes - https://trello.com/c/RlESb8jG
```
johnduffell: if we really want to do these salesforce removals, maybe we can have a cloudwatch metric and dashboard it! ;)

johnduffell: I suppose the metric is "number of high priority trello cards"

paulbrown1982: It's one more piece of functionality that'll work if Salesforce goes down. Which it rarely does!

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -41,8 +44,10 @@ class AccountController extends LazyLogging {
       user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
       sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
       subscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
+      account <- EitherT(tp.zuoraService.getAccount(subscription.accountId).map(x => Some(x)).map(_ \/> s"no account for subscription: ${subscription.name} with account id ${subscription.accountId}"))
+      stripeService = if (account.paymentGateway.contains(tp.auStripeService.paymentGateway.gatewayName)) tp.auStripeService else tp.ukStripeService
```
johnduffell: should we make the reader do this? I.e. turn the String into a sealed trait ADT to represent which.  Then we can pattern match on that.  Just a thought.

johnduffell:
yep a few todos which you're on, but I think overall looks good once it all builds 👍

membership-attribute-service/app/controllers/AccountController.scala
paulbrown1982: OK let's have a chat on Monday on how to do that. I did think about an implicit class over Account.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -76,7 +79,7 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   lazy val futureCatalog: Future[CatalogMap] = catalogService.catalog.map(_.fold[CatalogMap](error => {println(s"error: ${error.list.mkString}"); Map()}, _.map))
 
   lazy val subService = new SubscriptionService[Future](productIds, futureCatalog, simpleClient, zuoraService.getAccountIds)
-  lazy val paymentService = new PaymentService(stripeService, zuoraService, catalogService.unsafeCatalog.productMap)
+  lazy val paymentService = new PaymentService(zuoraService, catalogService.unsafeCatalog.productMap)
```
johnduffell: great, dependencies--

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -41,8 +44,10 @@ class AccountController extends LazyLogging {
       user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
       sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
       subscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
+      account <- EitherT(tp.zuoraService.getAccount(subscription.accountId).map(x => Some(x)).map(_ \/> s"no account for subscription: ${subscription.name} with account id ${subscription.accountId}"))
```
johnduffell: need a recover here not a map (the Some will never be turned into a None so the \/> will never be an error)

johnduffell: the second map needs to be a recover with a left, and the first one needs to map to a right

johnduffell:
👍 great!

PR #236, opened 2017-09-22
Minor fixes (proposal)

kant:
### Why do we need this? 
For the sake of clarity
### The changes 
Minor fixes on behalf of proper explanation


johnduffell:
thanks for the changes, unfortunately we can't merge things until the build passes, and we can only build branches in the same repo for security reasons, so we're in a catch 22 situation!

PR #238, opened 2017-09-25
Provide the Stripe public key to channels which allow card updating

paulbrown1982:
### Why do we need this? 
Credit cards used to pay for membership, Digipack and regular contributions can be updated on the Identity profile.theguardian.com pages. Right now the Stripe public key is hardcoded in that app. This change provides the public key used for updating the card along with the card details used for display. This is need for cards stored in Australia Stripe.

### The changes 
Adding the relevant Stripe public key to the payment details feed so that client channels know what public key they need to use to do a card update.

cc @johnduffell @rupertbates @AWare @lmath

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -42,12 +42,12 @@ class AccountController extends LazyLogging {
     logger.info(s"Attempting to update card for $maybeUserId")
     (for {
       user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
+      stripeCardToken <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no card token submitted with request"))
       sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
       subscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -42,12 +42,12 @@ class AccountController extends LazyLogging {
...
       sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
       subscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
       account <- EitherT(tp.zuoraService.getAccount(subscription.accountId).map(\/.right).recover { case x => \/.left(s"error receiving account for subscription: ${subscription.name} with account id ${subscription.accountId}. Reason: $x") })
-      stripeService = if (account.paymentGateway.contains(tp.auStripeService.paymentGateway)) tp.auStripeService else tp.ukStripeService
-      stripeCardToken <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no card token submitted with request"))
-      updateResult <- EitherT(tp.paymentService.setPaymentCardWithStripeToken(subscription.accountId, stripeCardToken, stripeService).map(_ \/> "something missing when try to zuora payment card"))
+      stripeService <- EitherT(Future.successful(account.paymentGateway.flatMap(tp.stripeServicesByPaymentGateway.get) \/> s"No Stripe service available for account: ${account.id}"))
```
jacobwinch: @paulbrown1982 FYI I just found a case where the user pays by Stripe but the paymentGateway was null!

PR #239, opened 2017-09-25
On zuora lookups, write updated information to the dynamo table

lmath:
Also, I was sick of updating `/me/zuora-lookup` as it's not actually being used so I'm removing it. 

### Why do we need this? 
We are moving towards doing member/supporter attributes lookups by getting subscription information directly from Zuora because then the information will be more up to date. 

However, during spikes of traffic such as push notifications we will make too many requests to Zuora. So this PR updates the the MembershipAttributes table when we do a Zuora lookup. That way, during a spike of traffic or a Zuora outage we have the option to serve from the Dynamo table, though it may be stale. 

There will be a PR to follow to detect when to serve from the dynamo table instead. 

### The changes 
- When we get attributes to Zuora,  we now use this information to update the `MembershipAttributes` dynamo table.
- Removed `/me/zuora-lookup` endpoint because it is not actually in use anymore.
- `update` in the ScanamoAttributeService should also update DigitalSubscriptionExpiryDate, if it's available. 

cc @paulbrown1982 @johnduffell @pvighi @AWare @jacobwinch

### DIFF
--- a/membership-attribute-service/test/repositories/DynamoAttributeServiceTest.scala
+++ b/membership-attribute-service/test/repositories/DynamoAttributeServiceTest.scala
@@ -6,7 +6,7 @@ import com.amazonaws.auth.BasicAWSCredentials
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClient
 import com.amazonaws.services.dynamodbv2.model.CreateTableRequest
 import com.github.dwhjames.awswrap.dynamodb.{AmazonDynamoDBScalaClient, Schema}
-import models.Attributes
+import models.{Attributes, CardDetails, Wallet}
 import org.joda.time.LocalDate
 import org.specs2.concurrent.ExecutionEnv
 import org.specs2.mutable.Specification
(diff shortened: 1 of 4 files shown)


membership-attribute-service/test/repositories/DynamoAttributeServiceTest.scala
```diff
@@ -87,4 +87,57 @@ class DynamoAttributeServiceTest extends Specification {
...
+
+      val result = for {
+        _ <- repo.update(newAttributes)
+        retrieved <- repo.get("6789")
+      } yield retrieved
+
+      result must be_==(Some(newAttributes)).await
+    }
+
+    "update a user who has bought a digital subscription" in {
```
lmath: @paulbrown1982 an update should update the DigitalSubscriptionExpiryDate right?

paulbrown1982: Yup it's fine. When reading back / upon construction it takes a max of the stored and generated date when serialising it back out for Staff subs.

paulbrown1982:
👍 Looks great!

PR #240, opened 2017-09-26
update dynamo only if we get some updated values from zuora

lmath:
### Why do we need this? 
To reduce the number of dynamo write requests needed (and so ££ spent on write capacity)

### The changes 
- Move the updating of dynamo from the controller to the attributesFromZuora function. We already query dynamo to get the adFree flag here and do a comparison between the two, so then we just update if the comparison says Zuora is more up to date.

[PR 239 - On zuora lookups, write updated information to the dynamo table](https://github.com/guardian/members-data-api/pull/239)

cc @paulbrown1982 @johnduffell @pvighi @AWare @jacobwinch 

Write capacity graph of MembershipAttributes-PROD before this change
writecapacitygraph_annotated

### DIFF
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -70,9 +70,15 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
         val attributesAfterUpdate = for {
             attributes <- OptionT(zuoraAttributesWithAdfree)
           _ = dynamoAttributeUpdater(attributes).map { result =>
-            result.left.map { error: DynamoReadError => log.error(s"Tried updating attributes for $identityId but then ${DynamoReadError.describe(error)}") }
+            result.left.map { error: DynamoReadError =>
+              log.warn(s"Tried updating attributes for $identityId but then ${DynamoReadError.describe(error)}")
+              log.error("Tried updating attributes with updated values from Zuora but there was a dynamo error.")
+            }
           }.onFailure {
-            case error => log.warn(s"Tried updating attributes for $identityId but then ${error.getMessage}", error)
+            case error => {
+              log.warn(s"Tried updating attributes for $identityId but then ${error.getMessage}", error)
+              log.error("Future failed when attempting to update dynamo when there were updated values from Zuora.")
+            }
           }
         } yield attributes
         attributesAfterUpdate.run
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -68,11 +68,13 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
         zuoraAttributesWithAdfree
       else {
         val attributesAfterUpdate = for {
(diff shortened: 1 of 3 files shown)


jacobwinch:
Good idea 👍

paulbrown1982:
👍 Looks great. Let's ramp up the throttle to 100% for 30 mins or so, then merge this change, so we can see a Dynamo metric graph showing the effect of this change.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -49,13 +51,32 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
   }
 
+  private def attributesAfterDynamoUpdate(identityId: String, skipUpdate: Future[Boolean], zuoraAttributesWithAdfree: Future[Option[Attributes]], dynamoAttributeUpdater: Attributes => Future[Either[DynamoReadError, Attributes]]) = {
+    skipUpdate flatMap { shouldSkipUpdate =>
+      if (shouldSkipUpdate)
+        zuoraAttributesWithAdfree
+      else {
+        val attributesAfterUpdate = for {
+            endAttributes <- OptionT(zuoraAttributesWithAdfree)
+          _ = dynamoAttributeUpdater(endAttributes).onFailure {
```
johnduffell: this onFailure will log a failed future, but a left of error will be silently dropped

johnduffell: what will happen if updates fail silently?  will we find out via sentry/cloudwatch that a high percentage of disagrees are happening?

johnduffell: I'm sure there are other ways, we have quite a few places in the code where we ignore errors or don't wait for Futures!

I was just thinking whether there was a more wholistic way to realise a problem, rather than relying on spotting the failure itself.  However it's probably a lot of work and it depends what the bad stuff is if it starts failing (we would serve out of date info if zuora is down anyway?) and how likely (probably not so likely to fail silently?)  So overall not a big press to do that.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -70,7 +70,7 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
         val attributesAfterUpdate = for {
             attributes <- OptionT(zuoraAttributesWithAdfree)
           _ = dynamoAttributeUpdater(attributes).map { result =>
-            result.left.map { error: DynamoReadError => log.warn(s"Tried updating attributes for $identityId but then ${DynamoReadError.describe(error)}") }
+            result.left.map { error: DynamoReadError => log.error(s"Tried updating attributes for $identityId but then ${DynamoReadError.describe(error)}") }
           }.onFailure {
             case error => log.warn(s"Tried updating attributes for $identityId but then ${error.getMessage}", error)
```
johnduffell: should be an error here really too

johnduffell: do we need to be careful of PII though? we can error log (sentry) one vaguer thing and warn log the full details (kibana/cloudwatch)

PR #243, opened 2017-09-29
Serve from cache if concurrency count over a threshold

lmath:
### Why do we need this? 
Now that we have a way to [lookup attributes via Zuora](https://github.com/guardian/members-data-api/pull/205) and [update the membership-attributes dynamo table](https://github.com/guardian/members-data-api/pull/239) on these lookups we want to be able to send 100% of the traffic to lookup attributes via Zuora.

However, during push notifications, this generates too many calls to Zuora. This is the simplest possible way to address that: just checking on each lookup call if the number of concurrent calls to Zuora is over a threshold.

### The changes 
- On each lookup call to Zuora, check to see if the number of concurrent calls is over a threshold, and if it is, serve the response from Dynamo

[lookup attributes via Zuora](https://github.com/guardian/members-data-api/pull/205)
[update the membership-attributes dynamo table on lookups via Zuora](https://github.com/guardian/members-data-api/pull/239)

There are default values in the agent that updates current percentage of lookup traffic via zuora and the Zuora and the threshold for concurrent calls to Zuora on that instance. I put in a default value of 10 concurrent calls per instance by looking at the pattern of concurrent calls for when the traffic was at 100% 

concurrency-count-annotated
cc @jacobwinch @paulbrown1982 @pvighi  @AWare

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -8,7 +8,7 @@ import com.gu.scanamo.error.DynamoReadError
 import configuration.Config
 import configuration.Config.authentication
 import loghandling.LoggingField.{LogField, LogFieldString}
-import loghandling.LoggingWithLogstashFields
+import loghandling.{LoggingWithLogstashFields, ZuoraRequestCounter}
 import models.ApiError._
 import models.ApiErrors._
 import models.Features._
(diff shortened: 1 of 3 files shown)


jacobwinch:
Looks good to me 👍

paulbrown1982:
👍 LGTM too!

PR #246, opened 2017-10-03
On lookups, fallback to Dynamo cache if getting subscriptions from Zuora fails

lmath:
### Why do we need this? 
So that we can return a response on calls to {/me, /me/features, /me/membership} even if there is a failure in getting subscriptions for a user.

### The changes 
- If we fail to get subscriptions for a user from Zuora, we return the response from the cache if we have something stored in the cache. 

[PR 239 - update the cache on zuora lookups](https://github.com/guardian/members-data-api/pull/239)
[PR 243 - serve from the cache if concurrent calls to zuora is over a threshold](https://github.com/guardian/members-data-api/pull/243)

cc @paulbrown1982 @jacobwinch @pvighi @AWare

### DIFF
--- a/membership-attribute-service/test/services/AttributesFromZuoraTest.scala
+++ b/membership-attribute-service/test/services/AttributesFromZuoraTest.scala
@@ -56,6 +56,25 @@ class AttributesFromZuoraTest(implicit ee: ExecutionEnv) extends Specification w
         val attributes: Future[Option[Attributes]] = AttributesFromZuora.getAttributes(testId, identityIdToAccountIds, subscriptionFromAccountId, dynamoAttributeGetter, dynamoAttributeUpdater, referenceDate)
         attributes must be_==(None).await
       }
+
+      "return the attributes from Dynamo if Zuora query for account ids fails" in new errorWhenGettingAccounts {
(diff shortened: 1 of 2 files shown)


membership-attribute-service/test/services/AttributesFromZuoraTest.scala
```diff
@@ -56,6 +56,33 @@ class AttributesFromZuoraTest(implicit ee: ExecutionEnv) extends Specification w
         val attributes: Future[Option[Attributes]] = AttributesFromZuora.getAttributes(testId, identityIdToAccountIds, subscriptionFromAccountId, dynamoAttributeGetter, dynamoAttributeUpdater, referenceDate)
         attributes must be_==(None).await
       }
+
+      "return the attributes from Dynamo if Zuora query for account ids fails" in new errorWhenGettingAccounts {
```
jacobwinch: nice work getting test coverage for all of the different cases 👍

jacobwinch:
Now that we have this additional layer of complexity, it will be helpful if we can start being explicit about whether we are serving from cache or from source to help with debugging (feel free to save this for a future PR though).

Good work 👍

lmath:
Good point @jacobwinch. I shall add that in a future PR.

PR #247, opened 2017-10-06
On Zuora lookups, update the cache when all subs are expired or cancelled too.

lmath:
### Why do we need this? 
Now when we send 100% of lookup traffic through Zuora, we will still serve from the dynamo table (as a cache) during spikes of traffic or if to calls to Zuora to get subscription information fail. 

When we do a lookup via Zuora, if there are no active subscriptions from Zuora/the attributes are None, then the update to the cache should mean that the entry for that user is removed from the table. Then, when we need to serve from the cache later on we report that the user has no active subscriptions.

### The changes 
* When we do a lookup via Zuora, if there are no active subscriptions from Zuora/the attributes are None, then the update to the cache should mean that the entry for that user is removed from the table.

[on trello](https://trello.com/c/IFuUjZnO/224-supporters-who-cancel-downgrade-are-never-removed-from-dynamo)
[PR 239 - on zuora lookups, update the dynamo cache](https://github.com/guardian/members-data-api/pull/239)
[PR 243 - serve from cache if concurrency over a threshold](https://github.com/guardian/members-data-api/pull/243)
[PR 246 - fallback to cache if a zuora lookup fails](https://github.com/guardian/members-data-api/pull/246)

### DIFF
--- a/membership-attribute-service/app/prodtest/FeatureToggleDataUpdatedOnSchedule.scala
+++ b/membership-attribute-service/app/prodtest/FeatureToggleDataUpdatedOnSchedule.scala
@@ -13,11 +13,11 @@ import scalaz.{-\/, \/-}
 class FeatureToggleDataUpdatedOnSchedule(featureToggleService: ScanamoFeatureToggleService, stage: String)(implicit ec: ExecutionContext, system: ActorSystem) extends LazyLogging {
 
   private val updateZuoraLookupFeatureDataTask: ScheduledTask[ZuoraLookupFeatureData] = {
-    val featureName = "UpdateAttributesFromZuoraLookup"
+    val featureName = "AttributesFromZuoraLookup"
     val defaultFeatureValues = ZuoraLookupFeatureData(TrafficPercentage = 0, ConcurrentZuoraCallThreshold = 10)
 
     ScheduledTask[ZuoraLookupFeatureData](featureName, defaultFeatureValues, 0.seconds, 30.seconds) {
-      featureToggleService.get("AttributesFromZuoraLookup") map { result =>
+      featureToggleService.get(featureName) map { result =>
         result match {
           case \/-(feature) =>
             val percentage = feature.TrafficPercentage.getOrElse(defaultFeatureValues.TrafficPercentage)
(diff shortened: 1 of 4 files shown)


membership-attribute-service/app/prodtest/FeatureToggleDataUpdatedOnSchedule.scala
```diff
@@ -13,11 +13,11 @@ import scalaz.{-\/, \/-}
 class FeatureToggleDataUpdatedOnSchedule(featureToggleService: ScanamoFeatureToggleService, stage: String)(implicit ec: ExecutionContext, system: ActorSystem) extends LazyLogging {
 
   private val updateZuoraLookupFeatureDataTask: ScheduledTask[ZuoraLookupFeatureData] = {
-    val featureName = "UpdateAttributesFromZuoraLookup"
```
lmath: I didn't mean for this featureName to diverge from what is is the FeatureToggles dynamo table in the first place. This affects the logging.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -43,64 +42,65 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
+
+    attributesFromZuoraOrDynamoFallback flatMap { attributesFromSomewhere =>
+      val (fromWhere: String, attributes: Option[Attributes]) = attributesFromSomewhere
 
-    attributesAfterDynamoUpdate(identityId, shouldSkipUpdate, zuoraAttributesWithAdfree, dynamoAttributeUpdater)
+      if(fromWhere == "Zuora") {
+        attributesFromDynamo flatMap { (dynamoAttributes: Option[Attributes]) =>
+          val shouldSkipUpdate: Boolean = dynamoAndZuoraAgree(dynamoAttributes, attributes, identityId)
+          val zuoraAttributesWithAdfree: Option[Attributes] = attributesWithFlagFromDynamo(attributes, dynamoAttributes)
+          attributesAfterDynamoUpdate(identityId, shouldSkipUpdate, zuoraAttributesWithAdfree, dynamoAttributeService)
```
jacobwinch: We could consider running this cache update in a separate thread, and then just returning zuoraAttributesWithAdFree regardless of the result of the cache update?

jacobwinch: If we did that, we'd probably want to rename attributesAfterDynamoUpdate to something like 'updateDynamo' or 'updateCache'...

PR #252, opened 2017-10-13
Don't make returning zuora attributes depend on the cache update

lmath:
### Why do we need this? 
We should just update the cache as a best effort, and return attributes from Zuora regardless, if we have them. 

### The changes 
- Start the cache update and log if it fails but return attributes from Zuora regardless

[pr 247 - update the cache for cancelled or expired subs too](https://github.com/guardian/members-data-api/pull/247)

### DIFF
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -143,16 +143,18 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
   }
 
   def dynamoAndZuoraAgree(maybeDynamoAttributes: Option[Attributes], maybeZuoraAttributes: Option[Attributes], identityId: String): Boolean = {
-     val zuoraAttributesWithIgnoredFields = maybeZuoraAttributes flatMap { zuoraAttributes =>
-       maybeDynamoAttributes map { dynamoAttributes =>
-         zuoraAttributes.copy(
-           AdFree = dynamoAttributes.AdFree, //fetched from Dynamo in the Zuora lookup anyway (dynamo is the source of truth)
-           Wallet = dynamoAttributes.Wallet, //can't be found based on Zuora lookups, and not currently used
-           MembershipNumber = dynamoAttributes.MembershipNumber, //I don't think membership number is needed and it comes from Salesforce
-           MembershipJoinDate = dynamoAttributes.MembershipJoinDate.flatMap(_ => zuoraAttributes.MembershipJoinDate), //only compare if dynamo has value
-           DigitalSubscriptionExpiryDate = None
-         )
-       }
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -60,40 +60,41 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
-          val zuoraAttributesWithAdfree: Option[Attributes] = attributesWithFlagFromDynamo(attributes, dynamoAttributes)
-          attributesAfterDynamoUpdate(identityId, shouldSkipUpdate, zuoraAttributesWithAdfree, dynamoAttributeService) map { (fromWhere, _) }
+        attributesFromDynamo map { (dynamoAttributes: Option[Attributes]) =>
+          val canSkipCacheUpdate: Boolean = dynamoAndZuoraAgree(dynamoAttributes, attributes, identityId)
+          val zuoraAttributesWithAdfree: Option[Attributes] = attributesWithAdFreeFlagFromDynamo(attributes, dynamoAttributes)
+
+          if(!canSkipCacheUpdate) {
+            updateCache(identityId, zuoraAttributesWithAdfree, dynamoAttributeService).onFailure {
+              case e: Throwable =>
+                log.error(s"Future failed when attempting to update cache. Update was needed? $canSkipCacheUpdate", e)
```
jacobwinch: I know you haven't introduced this here, but do we need two separate log lines for this?

jacobwinch:
Nice work - 👍 for when the tests are green!

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -60,40 +60,41 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
-            dynamoAttributeService.delete(identityId) map { result =>
-              log.info(s"Deleting the dynamo entry for $identityId. $result")
-              None
-            }
+  private def updateCache(identityId: String, zuoraAttributesWithAdfree: Option[Attributes], dynamoAttributeService: AttributeService): Future[Unit] = {
+    zuoraAttributesWithAdfree match {
+      case Some(attributes) =>
+        dynamoAttributeService.update(attributes).map { result =>
+          result.left.map { error: DynamoReadError =>
+            log.warn(s"Tried updating attributes for $identityId but then ${DynamoReadError.describe(error)}")
```
jacobwinch: Same comment here about the logging

membership-attribute-service/app/services/AttributesFromZuora.scala
lmath: I wanted to be able to see the attributes from Zuora in Kibana in case it's needed for debugging, but an error log to get picked up by Sentry if updates are failing.

PR #253, opened 2017-10-17
Remove dead provisioning code

jacobwinch:
### Why do we need this? 
General health/housekeeping.

### The changes 
* Delete packer/AMI config and README sections, since we are now using AMIgo.
* Delete ssh keys script and invocation - since this is taken care of by the [S3 SSH keys role](https://amigo.gutools.co.uk/roles#s3-ssh-keys), which is part of our [AMIgo bake](https://amigo.gutools.co.uk/recipes/xenial-membership).
* Delete old deploy.json - I think this must have been used to copy static provisioning files to S3, and it now seems to be redundant.

* The changeover to automatic AMI updates (using AMIgo), happened here: https://github.com/guardian/members-data-api/pull/179

### Before merge:

- [x] Cloudform CODE stack, then deploy using riff-raff
- [x] Confirm that it's still possible to ssh onto a box
- [x] Cloudform PROD stack (after +1)

cc @pvighi @lmath


lmath:
👍  lgtm!

PR #254, opened 2017-10-17
Serve from the cache if the future fails on calls to Zuora

lmath:
### Why do we need this? 
We should return a value from the cache if there is one when the future fails on calls to Zuora because it's better than returning a 500. 

### The changes 
- If an individual call to {/me, /me/membership, /me/features} fails because the future fails on a call to Zuora, we return the value from the cache, if there is one.

[PR 246](https://github.com/guardian/members-data-api/pull/246) handles lefts on calls to Zuora by serving from the cache, but this happens rarely.  

cc @jacobwinch @pvighi @paulbrown1982

### DIFF
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -42,7 +42,7 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
     }
 
 
-    val attributesFromDynamo = dynamoAttributeService.get(identityId)
+    val attributesFromDynamo: Future[(String, Option[Attributes])] = dynamoAttributeService.get(identityId) map { attributes => ("Dynamo", attributes)}
 
     val attributesFromZuora: Future[Disjunction[String, Option[Attributes]]] = attributesDisjunction.run.map {
         _.leftMap { errorMsg =>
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -54,13 +54,16 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
     val attributesFromZuoraOrDynamoFallback: Future[(String, Option[Attributes])] = for {
       dynamoAttributes <- attributesFromDynamo
       attributesFromZuora <- attributesFromZuora
```
pvighi: not introduced in this PR but it's a little confusing having the same name in both sides of the arrow here

lmath: Agreed. I was surprised that even worked.

Yeah, that was an accident and I'm going to change it.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -54,13 +54,16 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
-    attributesFromZuoraOrDynamoFallback flatMap { attributesFromSomewhere =>
+    val zuoraOrCachedAttributes = attributesFromZuoraOrDynamoFallback fallbackTo attributesFromDynamo
+
+    zuoraOrCachedAttributes flatMap { attributesFromSomewhere =>
       val (fromWhere: String, attributes: Option[Attributes]) = attributesFromSomewhere
 
       if(fromWhere == "Zuora") {
-        attributesFromDynamo map { (dynamoAttributes: Option[Attributes]) =>
+        attributesFromDynamo map { sourceAndAttributes =>
+          val (_, dynamoAttributes) = sourceAndAttributes
```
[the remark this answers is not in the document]
lmath: Oh nice. I will take your suggestion.

PR #256, opened 2017-10-20
Remove Salesforce message receiver

jacobwinch:
### Why do we need this?
* Removes dead code as part of the work to make digital fulfilment more reliable. All of this code will be redundant once the corresponding Salesforce changes are made, but we can't merge this until the outbound message is switched off.

### The changes 
* Delete unused controller, SF parser and actions
* Tidy up routes
* Remove tests for redundant code
* Remove whitespace
* Remove references from README

Corresponding PR in membership: https://github.com/guardian/membership-frontend/pull/1729

### Before merge
- [x] Disable outbound message

### After merge
- [ ] Delete org 'secrets' from private conf files


jacobwinch:
cc @davidpepper223 - this is the reason that I'd like to disable an outbound message in SF - will talk you through the history tomorrow!

PR #259, opened 2017-10-30
Migrate au customers to au stripe

paulbrown1982:
### Why do we need this? 
We want customers in Australia to transition over to the new AU Stripe account both organically as they update their payment details, and upon advice from the contact centre should they phone up and complain about international bank charges.

### The changes 
This change determines which Stripe account is used to generate the card token by receiving the Stripe public key back in the AJAX call from the MMA page. This means Members Data API now knows, reliably, which Stripe account to create the Customer object in. It also lets us save a call to retrieve the account object.

### Depends on: ###
https://github.com/guardian/frontend/pull/18089 and https://github.com/guardian/membership-common/pull/543

### Tested in code ###
Test script: 
1) Create a new UK Supporter in CODE
2) Change Billing Country to AU in Zuora. Do card update. 
3) Change Billing country to UK in Zuora. Do card update.

### Screenshots ###
Before:
picture 225

After:

picture 226

cc @jacobwinch @johnduffell

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -3,8 +3,11 @@ package components
 import akka.actor.ActorSystem
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClient
+
 import scalaz.std.scalaFuture._
 import com.gu.config
+import com.gu.i18n.Country
+import com.gu.zuora.api.InvoiceTemplate
 import com.gu.memsub.services.PaymentService
 import com.gu.memsub.subsv2.services.SubscriptionService.CatalogMap
 import com.gu.memsub.subsv2.services._
@@ -13,7 +16,7 @@ import com.gu.okhttp.RequestRunners
 import com.gu.salesforce.SimpleContactRepository
 import com.gu.stripe.StripeService
 import com.gu.touchpoint.TouchpointBackendConfig
-import com.gu.zuora.api.PaymentGateway
+import com.gu.zuora.api.{InvoiceTemplate, InvoiceTemplates, PaymentGateway}
 import com.gu.zuora.rest.SimpleClient
 import com.gu.zuora.soap.ClientWithFeatureSupplier
 import com.gu.zuora.{ZuoraRestService, ZuoraService}
@@ -42,6 +45,7 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   lazy val dynamoAttributesTable = environmentConf.getString("dynamodb.table")
   lazy val dynamoBehaviourTable = environmentConf.getString("behaviour.dynamodb.table")
   lazy val dynamoFeatureToggleTable = environmentConf.getString("featureToggles.dynamodb.table")
+  lazy val invoiceTemplatesConf = environmentConf.getConfig(s"zuora.invoiceTemplateIds")
 
   lazy val digitalPackPlans = config.DigitalPackRatePlanIds.fromConfig(digitalPackConf)
   lazy val productIds = config.SubsV2ProductIds(environmentConf.getConfig("zuora.productIds"))
@@ -56,6 +60,7 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   lazy val auStripeService = new StripeService(tpConfig.stripeAUMembership, RequestRunners.loggingRunner(metrics("stripe")))
   lazy val allStripeServices = Seq(ukStripeService, auStripeService)
   lazy val stripeServicesByPaymentGateway: Map[PaymentGateway, StripeService] = allStripeServices.map(s => s.paymentGateway -> s).toMap
+  lazy val invoiceTemplateIdsByCountry: Map[Country, InvoiceTemplate] = InvoiceTemplates.fromConfig(invoiceTemplatesConf).map(it => (it.country, it)).toMap
 
   lazy val soapClient = new ClientWithFeatureSupplier(Set.empty, tpConfig.zuoraSoap,
     RequestRunners.loggingRunner(metrics("zuora-soap")),
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -54,7 +54,8 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
 
   lazy val ukStripeService = new StripeService(tpConfig.stripeUKMembership, RequestRunners.loggingRunner(metrics("stripe")))
   lazy val auStripeService = new StripeService(tpConfig.stripeAUMembership, RequestRunners.loggingRunner(metrics("stripe")))
-  lazy val stripeServicesByPaymentGateway: Map[PaymentGateway, StripeService] = Seq(ukStripeService, auStripeService).map(s => s.paymentGateway -> s).toMap
+  lazy val allStripeServices = Seq(ukStripeService, auStripeService)
+  lazy val stripeServicesByPaymentGateway: Map[PaymentGateway, StripeService] = allStripeServices.map(s => s.paymentGateway -> s).toMap
 
   lazy val soapClient = new ClientWithFeatureSupplier(Set.empty, tpConfig.zuoraSoap,
     RequestRunners.loggingRunner(metrics("zuora-soap")),
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -60,6 +60,7 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
   lazy val auStripeService = new StripeService(tpConfig.stripeAUMembership, RequestRunners.loggingRunner(metrics("stripe")))
   lazy val allStripeServices = Seq(ukStripeService, auStripeService)
   lazy val stripeServicesByPaymentGateway: Map[PaymentGateway, StripeService] = allStripeServices.map(s => s.paymentGateway -> s).toMap
+  lazy val stripeServicesByPublicKey: Map[String, StripeService] = allStripeServices.map(s => s.publicKey -> s).toMap
   lazy val invoiceTemplateIdsByCountry: Map[Country, InvoiceTemplate] = InvoiceTemplates.fromConfig(invoiceTemplatesConf).map(it => (it.country, it)).toMap
 
   lazy val soapClient = new ClientWithFeatureSupplier(Set.empty, tpConfig.zuoraSoap,
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -8,6 +8,8 @@ import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.services.model.PaymentDetails
+import com.gu.stripe.StripeService
+import com.gu.zuora.api.RegionalStripeGateways
 import com.typesafe.scalalogging.LazyLogging
 import components.TouchpointComponents
 import configuration.Config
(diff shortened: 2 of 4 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -105,11 +110,12 @@ class AccountController extends LazyLogging {
       freeOrPaidSub <- OptionEither(tp.subService.either[F, P](contact).map(_.leftMap(message => s"couldn't read sub from zuora for crmId ${contact.salesforceAccountId} due to $message")))
       details <- OptionEither.liftOption(tp.paymentService.paymentDetails(freeOrPaidSub).map(\/.right))
       sub = freeOrPaidSub.fold(identity, identity)
-      account <- OptionEither.liftOption(tp.zuoraService.getAccount(sub.accountId).map(\/.right).recover { case x => \/.left(s"error receiving account for subscription: ${sub.name} with account id ${sub.accountId}. Reason: $x") })
-      publicKey = account.paymentGateway.flatMap(tp.stripeServicesByPaymentGateway.get).map(_.publicKey)
+      account <- OptionEither.liftOption(tp.zuoraService.getAccount(sub.accountId).map(\/.right).recover { case x => \/.left(s"error retrieving account for subscription: ${sub.name} with account id ${sub.accountId}. Reason: $x") })
+      billToContact <- OptionEither.liftOption(tp.zuoraService.getContact(account.billToId).map(\/.right).recover { case x => \/.left(s"error retrieving bill to contact id (${account.billToId}) for account id ${account.id}. Reason $x")})
```
jacobwinch: Is this an extra network call? It's a shame that we can't just get what we need from the account...

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,18 +42,18 @@ class AccountController extends LazyLogging {
...
     logger.info(s"Attempting to update card for $maybeUserId")
     (for {
-      user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
-      stripeCardToken <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no card token submitted with request"))
+      user <- EitherT(Future.successful(maybeUserId \/> "no identity cookie for user"))
+      stripeDetails <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no card token submitted with request"))
+      (stripeCardToken, stripePublicKey) = stripeDetails
       sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
       subscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
-      account <- EitherT(tp.zuoraService.getAccount(subscription.accountId).map(\/.right).recover { case x => \/.left(s"error receiving account for subscription: ${subscription.name} with account id ${subscription.accountId}. Reason: $x") })
```
jacobwinch: 👍

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -69,18 +71,21 @@ class AccountController extends LazyLogging {
...
           paymentMethodId <- OptionT(Future.successful(defaultPaymentMethodId.map(_.trim).filter(_.nonEmpty)))
           zuoraPaymentMethod <- tp.zuoraService.getPaymentMethod(paymentMethodId).liftM[OptionT]
           customerId <- OptionT(Future.successful(zuoraPaymentMethod.secondTokenId.map(_.trim).filter(_.startsWith("cus_"))))
-          maybeUkStripeCustomer <- tp.ukStripeService.Customer.read(customerId).map(Option(_)).recover { case _ => None }.liftM[OptionT]
-          maybeStripeCustomer <-
-            // Performance optimisation not to go to AU Stripe if UK is one gotten above
-            if (maybeUkStripeCustomer.nonEmpty) Future.successful(maybeUkStripeCustomer).liftM[OptionT]
-            else tp.auStripeService.Customer.read(customerId).map(Option(_)).recover { case _ => None }.liftM[OptionT]
+          maybeAStripeCustomer <- stripeService.Customer.read(customerId).map(Option(_)).recover { case _ => None }.liftM[OptionT]
+          maybeStripeCustomer <- {
```
jacobwinch: This is getting pretty tough to follow within the for comprehension - could we pull it out into a separate function for read/test-ability?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,18 +42,18 @@ class AccountController extends LazyLogging {
 
     // TODO - refactor to use the Zuora-only based lookup, like in AttributeController.pickAttributes - https://trello.com/c/RlESb8jG
```
jacobwinch: As an aside, we should be able to clean up this code significantly once we've done this re-factor. Given that these updates are happening at the account level, we should be able to stop caring about the subscription/plan types altogether, and have a single endpoint which will update a users card for any account with an active sub.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,18 +42,18 @@ class AccountController extends LazyLogging {
...
 
-    val updateForm = Form { single("stripeToken" -> nonEmptyText) }
+    val updateForm = Form { tuple("stripeToken" -> nonEmptyText, "publicKey" -> text) }
```
johnduffell: what does `tuple` do? is it just to make the types explicit?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,18 +42,18 @@ class AccountController extends LazyLogging {
...
     val tp = request.touchpoint
     val maybeUserId = authenticationService.userId
     logger.info(s"Attempting to update card for $maybeUserId")
     (for {
-      user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
-      stripeCardToken <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no card token submitted with request"))
+      user <- EitherT(Future.successful(maybeUserId \/> "no identity cookie for user"))
+      stripeDetails <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no card token submitted with request"))
+      (stripeCardToken, stripePublicKey) = stripeDetails
```
johnduffell: could this have been done on the line before? it doesn't look like we use the `stripeDetails` val anywhere as it is

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,18 +42,18 @@ class AccountController extends LazyLogging {
...
       sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
       subscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
-      account <- EitherT(tp.zuoraService.getAccount(subscription.accountId).map(\/.right).recover { case x => \/.left(s"error receiving account for subscription: ${subscription.name} with account id ${subscription.accountId}. Reason: $x") })
-      stripeService <- EitherT(Future.successful(account.paymentGateway.flatMap(tp.stripeServicesByPaymentGateway.get) \/> s"No Stripe service available for account: ${account.id}"))
-      updateResult <- EitherT(tp.paymentService.setPaymentCardWithStripeToken(subscription.accountId, stripeCardToken, stripeService, Some(user)).map(_ \/> "something missing when try to zuora payment card"))
+      stripeService <- EitherT(Future.successful(tp.allStripeServices.find(_.publicKey == stripePublicKey)).map(_ \/> s"No Stripe service for public key: $stripePublicKey"))
```
johnduffell: we could use a Map rather than doing a find, for clarity?

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -56,7 +60,9 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem) {
 
   lazy val ukStripeService = new StripeService(tpConfig.stripeUKMembership, RequestRunners.loggingRunner(metrics("stripe")))
   lazy val auStripeService = new StripeService(tpConfig.stripeAUMembership, RequestRunners.loggingRunner(metrics("stripe")))
-  lazy val stripeServicesByPaymentGateway: Map[PaymentGateway, StripeService] = Seq(ukStripeService, auStripeService).map(s => s.paymentGateway -> s).toMap
+  lazy val allStripeServices = Seq(ukStripeService, auStripeService)
```
johnduffell: really, it's the stripe config that changes, not the whole stripe service, but I guess that's a bigger symptom of the touchpoint backend bloat

johnduffell:
👍 agreed with the complexity about the for comprehension, I think with it being only two services at the moment, running through a list might be overkill and we should just be explicit that we have UK and AU perhaps?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,18 +42,18 @@ class AccountController extends LazyLogging {
...
-    val updateForm = Form { single("stripeToken" -> nonEmptyText) }
+    val updateForm = Form { tuple("stripeToken" -> nonEmptyText, "publicKey" -> text) }
     val tp = request.touchpoint
     val maybeUserId = authenticationService.userId
     logger.info(s"Attempting to update card for $maybeUserId")
     (for {
-      user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
-      stripeCardToken <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no card token submitted with request"))
+      user <- EitherT(Future.successful(maybeUserId \/> "no identity cookie for user"))
+      stripeDetails <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no card token submitted with request"))
```
jacobwinch: We should tweak the error message here too, as it could be a missing token **or** public key

membership-attribute-service/app/controllers/AccountController.scala
paulbrown1982: It's just saying you're looking to process 2 form-data fields

membership-attribute-service/app/controllers/AccountController.scala
paulbrown1982: Maybe you can help me with this? https://stackoverflow.com/questions/32651337/scala-for-comprehension-with-tuple-decomposition

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -80,14 +80,16 @@ class AccountController extends LazyLogging {
...
-  private def getUpToDatePaymentDetailsFromStripe(accountId: Subscription.AccountId, paymentDetails: PaymentDetails, stripeService: StripeService)(implicit tp: TouchpointComponents): Future[PaymentDetails] = {
+  private def getUpToDatePaymentDetailsFromStripe(accountId: Subscription.AccountId, paymentDetails: PaymentDetails)(implicit tp: TouchpointComponents): Future[PaymentDetails] = {
     paymentDetails.paymentMethod.map {
       case card: PaymentCard =>
         (for {
           account <- tp.zuoraService.getAccount(accountId).liftM[OptionT]
           defaultPaymentMethodId <- OptionT(Future.successful(account.defaultPaymentMethodId.map(_.trim).filter(_.nonEmpty)))
           zuoraPaymentMethod <- tp.zuoraService.getPaymentMethod(defaultPaymentMethodId).liftM[OptionT]
           customerId <- OptionT(Future.successful(zuoraPaymentMethod.secondTokenId.map(_.trim).filter(_.startsWith("cus_"))))
+          paymentGateway <- OptionT(Future.successful(account.paymentGateway))
```
jacobwinch: 👍

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -69,28 +71,37 @@ class AccountController extends LazyLogging {
...
+          account <- tp.zuoraService.getAccount(accountId).liftM[OptionT]
+          defaultPaymentMethodId <- OptionT(Future.successful(account.defaultPaymentMethodId.map(_.trim).filter(_.nonEmpty)))
+          zuoraPaymentMethod <- tp.zuoraService.getPaymentMethod(defaultPaymentMethodId).liftM[OptionT]
           customerId <- OptionT(Future.successful(zuoraPaymentMethod.secondTokenId.map(_.trim).filter(_.startsWith("cus_"))))
-          maybeUkStripeCustomer <- tp.ukStripeService.Customer.read(customerId).map(Option(_)).recover { case _ => None }.liftM[OptionT]
-          maybeStripeCustomer <-
-            // Performance optimisation not to go to AU Stripe if UK is one gotten above
-            if (maybeUkStripeCustomer.nonEmpty) Future.successful(maybeUkStripeCustomer).liftM[OptionT]
-            else tp.auStripeService.Customer.read(customerId).map(Option(_)).recover { case _ => None }.liftM[OptionT]
+          paymentGateway <- OptionT(Future.successful(account.paymentGateway))
```
davidfurey: `OptionT(Future.successful(` and `EitherT(Future.successful(` come up quite a lot in this section of code.  I know it is necessary, but it is quite distracting when reading the code.  A helpful method with a shorter name might make it more readable.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -113,11 +114,10 @@ class AccountController extends LazyLogging {
       freeOrPaidSub <- OptionEither(tp.subService.either[F, P](contact).map(_.leftMap(message => s"couldn't read sub from zuora for crmId ${contact.salesforceAccountId} due to $message")))
       details <- OptionEither.liftOption(tp.paymentService.paymentDetails(freeOrPaidSub).map(\/.right))
       sub = freeOrPaidSub.fold(identity, identity)
```
jacobwinch: Can we retain the 'recover' here even though we've swapped the call?

jacobwinch:
Looks much better to me 👍 (pending membership-common release).

membership-attribute-service/app/controllers/AccountController.scala
paulbrown1982: Good spot

paulbrown1982:
Tested in prod with a Digital Pack subscription and all looks good to me.

PR #260, opened 2017-11-07
Adds mma-paper endpoint to allow paper subscribers to update their card details

AWare:
### Why do we need this? 
This allows paper subscribers to change their cards*. 
*(Accompanying subs-frontend and frontend prs incomong)
### The changes 
Uses existing types on the mma methods.

### DIFF
--- a/membership-attribute-service/conf/DEV.public.conf
+++ b/membership-attribute-service/conf/DEV.public.conf
@@ -31,7 +31,8 @@ authentication {
 
 mma.cors.allowedOrigins = [
   "https://profile.code.dev-theguardian.com",
-  "https://profile.thegulocal.com"
+  "https://profile.thegulocal.com",
+  "https://sub.thegulocal.com"
 ]
 
 ft.cors.allowedOrigins = [
--- a/membership-attribute-service/conf/routes
+++ b/membership-attribute-service/conf/routes
@@ -7,12 +7,16 @@ GET        /user-attributes/me/features                     controllers.Attribut
 GET        /user-attributes/me/mma-digitalpack              controllers.AccountController.digitalPackDetails
 GET        /user-attributes/me/mma-membership               controllers.AccountController.membershipDetails
 GET        /user-attributes/me/mma-monthlycontribution      controllers.AccountController.monthlyContributionDetails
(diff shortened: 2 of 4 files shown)


membership-attribute-service/conf/DEV.public.conf
```diff
@@ -31,7 +31,8 @@ authentication {
 
 mma.cors.allowedOrigins = [
   "https://profile.code.dev-theguardian.com",
-  "https://profile.thegulocal.com"
+  "https://profile.thegulocal.com",
```
jacobwinch: I think you'll need to make the corresponding change in the PROD config file

membership-attribute-service/conf/routes
```diff
@@ -7,13 +7,17 @@ GET        /user-attributes/me/features                     controllers.Attribut
...
 GET        /user-attributes/me/mma-monthlycontribution      controllers.AccountController.monthlyContributionDetails
+GET        /user-attributes/me/mma-paper                    controllers.AccountController.paperDetails
 
 POST       /user-attributes/me/membership-update-card       controllers.AccountController.membershipUpdateCard
 OPTIONS    /user-attributes/me/membership-update-card       controllers.AccountController.membershipUpdateCard
 
 POST       /user-attributes/me/digitalpack-update-card      controllers.AccountController.digitalPackUpdateCard
 OPTIONS    /user-attributes/me/digitalpack-update-card      controllers.AccountController.digitalPackUpdateCard
 
+POST       /user-attributes/me/paper-update-card      controllers.AccountController.paperUpdateCard
```
jacobwinch: I appreciate that this is following the existing pattern, but it would be good if we could look to consolidate all of these endpoints under /me/update-card in the future - as all this is happening at the account level, I don't think we should need a new function for every card update

jacobwinch:
👍 once the prod config file is updated

membership-attribute-service/conf/DEV.public.conf
AWare: Done, the PROD file has some code.dev urls in it. As subscriptions isn't really running a CODE env right now I've got mixed feelings about including them.

AWare: Being the subs code environment url too.

PR #261, opened 2017-11-10
Added cancel my account functionality

svillafe:
### Why do we need this? 
We need to provide to our user a way of cancelling their account. This PR adds a new endpoint to the API which will:

* Disable Autopay
* Cancel contributor subscription
* Update Reason for cancelling

The new url is:

`[POST] /user-attributes/me/cancel-regular-contribution`

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -48,17 +48,21 @@ class AccountController extends Controller with LazyLogging {
     val maybeUserId = authenticationService.userId
     logger.info(s"Attempting to cancel contribution for $maybeUserId")
 
+
+
     (for {
       user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
+      cancellationReason <- EitherT(Future.successful(cancelForm.bindFromRequest().value \/> "no reason for cancellation submitted with request"))
+      (reason) = cancellationReason
       sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
       zuoraSubscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
       _ <- EitherT(tp.zuoraRestService.disableAutoPay(zuoraSubscription.accountId)).leftMap(message => s"Error while trying to disable AutoPay: $message")
       _ <- EitherT(tp.zuoraRestService.cancelSubscription(zuoraSubscription.name)).leftMap(message => s"Error while cancelling subscription: $message")
-      updateReasonResult <- EitherT(tp.zuoraRestService.updateCancellationReason(zuoraSubscription.name, cancelForm.bindFromRequest().value.get)).leftMap(message => s"Error while updating cancellation reason: $message")
+      updateReasonResult <- EitherT(tp.zuoraRestService.updateCancellationReason(zuoraSubscription.name, reason)).leftMap(message => s"Error while updating cancellation reason: $message")
     } yield updateReasonResult).run.map {
       case -\/(message) =>
         logger.warn(s"Failed to cancel subscription for user $maybeUserId, due to $message")
-        NotFound()//notFound
+        notFound //notFound
       case \/-(result) => Ok("Subscription cancelled correctly")
     }
 
@@ -93,7 +97,7 @@ class AccountController extends Controller with LazyLogging {
     }).run.map {
       case -\/(message) =>
         logger.warn(s"Failed to update card for user $maybeUserId, due to $message")
-        notFound
+        InternalServerError(s"Failed to update card for user $maybeUserId")
       case \/-(result) => result
     }
   }
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -3,7 +3,7 @@ import actions._
 import play.api.libs.concurrent.Execution.Implicits._
 import services.{AuthenticationService, IdentityAuthService}
 import com.gu.memsub._
-import com.gu.memsub.subsv2.SubscriptionPlan
+import com.gu.memsub.subsv2.{Subscription, SubscriptionPlan}
 import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads
 import com.gu.memsub.subsv2.reads.SubPlanReads._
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -36,12 +36,37 @@ class AccountController extends LazyLogging {
   lazy val mmaAction = NoCacheAction andThen mmaCorsFilter andThen BackendFromCookieAction
   lazy val mmaCardAction = NoCacheAction andThen corsCardFilter andThen BackendFromCookieAction
 
+
+  def cancelAccount [P <: SubscriptionPlan.AnyPlan : SubPlanReads] = mmaCardAction.async { implicit request =>
```
pvighi: shouldn't this be cancelSubscription or something like that?

svillafe: Yeap

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -36,12 +36,37 @@ class AccountController extends LazyLogging {
   lazy val mmaAction = NoCacheAction andThen mmaCorsFilter andThen BackendFromCookieAction
   lazy val mmaCardAction = NoCacheAction andThen corsCardFilter andThen BackendFromCookieAction
 
+
+  def cancelContribution [P <: SubscriptionPlan.AnyPlan : SubPlanReads] = mmaCardAction.async { implicit request =>
+    val tp = request.touchpoint
+    val cancelForm = Form { single("reason" -> nonEmptyText) }
+
+    val maybeUserId = authenticationService.userId
+    logger.info(s"Attempting to cancel account for $maybeUserId")
```
pvighi: the log message should also say contribution instead of account here

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -36,12 +36,37 @@ class AccountController extends LazyLogging {
...
+    (for {
+      user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
+      sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
+      zuoraSubscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
+      _ <- EitherT(tp.zuoraRestService.disableAutoPay(zuoraSubscription.accountId)).leftMap(message => s"Error message: $message")
+      _ <- EitherT(tp.zuoraRestService.cancelSubscription(zuoraSubscription.name)).leftMap(message => s"Error message: $message")
+      updateReasonResult <- EitherT(tp.zuoraRestService.updateCancellationReason(zuoraSubscription.name, cancelForm.bindFromRequest().value.get)).leftMap(message => s"Error message: $message")
+    } yield updateReasonResult).run.map {
+      case -\/(message) =>
+        logger.warn(s"Failed to update card for user $maybeUserId, due to $message")
```
pvighi: the name 'updateReasonResult' and the log message seem to be left over from the code you based this on right?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -38,12 +38,37 @@ class AccountController extends LazyLogging {
...
+    val tp = request.touchpoint
+    val cancelForm = Form { single("reason" -> nonEmptyText) }
+
+    val maybeUserId = authenticationService.userId
+    logger.info(s"Attempting to cancel contribution for $maybeUserId")
+
+    (for {
+      user <- EitherT(Future.successful( maybeUserId \/> "no identity cookie for user"))
+      sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
+      zuoraSubscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
```
pvighi: Can we be sure this will always be a contributions subscription ?

svillafe: Yes, we went through the code offline and discussed it.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -38,12 +38,37 @@ class AccountController extends LazyLogging {
...
+      _ <- EitherT(tp.zuoraRestService.disableAutoPay(zuoraSubscription.accountId)).leftMap(message => s"Error while trying to disable AutoPay: $message")
+      _ <- EitherT(tp.zuoraRestService.cancelSubscription(zuoraSubscription.name)).leftMap(message => s"Error while cancelling subscription: $message")
+      updateReasonResult <- EitherT(tp.zuoraRestService.updateCancellationReason(zuoraSubscription.name, cancelForm.bindFromRequest().value.get)).leftMap(message => s"Error while updating cancellation reason: $message")
+    } yield updateReasonResult).run.map {
+      case -\/(message) =>
+        logger.warn(s"Failed to cancel subscription for user $maybeUserId, due to $message")
+        NotFound()//notFound
```
jacobwinch: Really there are two cases here, which I'd expect a client of this API to handle differently:

1) We tried to look up the contribution and couldn't find it (404 seems appropriate for this)
2) We found the contribution and tried to cancel it, but something went wrong (404 is not appropriate here)

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,12 +41,77 @@ class AccountController extends Controller with LazyLogging {
...
+        \/-(cancellationReason)
+      }.getOrElse {
+        logger.warn("No reason for cancellation was submitted with the request.")
+        -\/(badRequest("Malformed request. Expected a valid reason for cancellation."))
+      }
+    }
+
+    def retrieveZuoraSubscription(user: String): Future[ApiError \/ Subscription[P]] = {
+      val getSubscriptionData = for {
+        sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"No Salesforce user: $user")))
```
jacobwinch: We should change this to use the Zuora based lookup in a future PR.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,12 +41,77 @@ class AccountController extends Controller with LazyLogging {
...
+
+
+  def cancelSubscription [P <: SubscriptionPlan.AnyPlan : SubPlanReads] = mmaCardAction.async { implicit request =>
+
+    val tp = request.touchpoint
+    val cancelForm = Form { single("reason" -> nonEmptyText) }
+    val maybeUserId = authenticationService.userId
+
+    def handleInputBody(cancelForm: Form[String]): Future[ApiError \/ String] = Future.successful {
+      cancelForm.bindFromRequest().value.map { cancellationReason =>
```
pvighi: shouldn't this be a json post instead of a regular form ? how is this going to be called ?

svillafe: Good call, I will change the format of the post to be JSON only (if it is possible) in a future PR.

PR #262, opened 2017-11-13
Upgrade to play 2.5

svillafe:
Upgrade to play 2.5 

### Why do we need this? 
To be able to use the latest version of membership-common.

cc: @rupertbates


membership-attribute-service/app/filters/LogRequestsFilter.scala
```diff
@@ -1,28 +0,0 @@
-package filters
-
-import play.api.Logger
-import play.api.mvc._
-import play.api.libs.concurrent.Execution.Implicits.defaultContext
-import scala.concurrent.Future
-import Headers._
-
-/** Logs all requests, the time the app took to respond, and the status of the response */
-object LogRequestsFilter extends Filter {
```
jacobwinch: Was this unused? Presumably we record all of this information in the Fastly logs anyway...

svillafe: Yes, we ran a full-text search of this object in the codebase and we didn't find it.

membership-attribute-service/app/Global.scala
```diff
@@ -1,23 +1,17 @@
...
 import scala.concurrent.Future
 
 
-object Global extends WithFilters(
-  CheckCacheHeadersFilter,
-  CSRFFilter(),
-  AddEC2InstanceHeader,
-  AddGuIdentityHeaders
-) {
+object Global extends GlobalSettings {
```
jacobwinch: It looks like we should be aiming to remove this altogether in a future PR? https://www.playframework.com/documentation/2.6.x/Migration25#Deprecated-GlobalSettings

membership-attribute-service/app/filters/LogRequestsFilter.scala
jacobwinch: 👍

PR #265, opened 2017-11-13
Reapply play 2.5

rupertbates:
### Why do we need this?
Reapplies the changes introduced in https://github.com/guardian/members-data-api/pull/262
and fixes the error caused by that PR.

### The changes 
The previous PR switched over to Play 2.5 without taking account of the changed CSRF filter behaviour that introduces. Specifically, in Play 2.5 by default, requests of type `application/json` need a CSRF token to be valid. This PR removes this requirement to revert the behaviour to that of Play 2.4.


rupertbates:
Tested successfully on code so I'm going to merge

rupertbates:
All good on PROD

PR #269, opened 2017-11-24
Remove zuora healthcheck

paulbrown1982:
### Why do we need this? 
[Retrospective action](https://docs.google.com/document/d/1GYqi3VrfogEmODi2vh86WU6xRi8uP82IrI0W3FfHzqo/edit#heading=h.bx4ztaqu25hd): Zuora outages should only degrade certain services not take down this app. Its critical function now has a backup cache in Dynamo which kicks in if it can't talk to Zuora.

### The changes 
- Stopped a failing Zuora health check from taking Members Data API out of service, instead it just logs the failure, so we can track it. I'll raise another PR to add/change this to a CloudWatch Metric.
- Refactored how the health is checked on Service classes, applying a HealthCheckableService trait and overriding the checkHealth on each concrete class.
- Added healthchecks for the 3 Dynamo DB services and created a new SalesforceService wrapper of the SimpleContactRepository so that the HealthCheckController can be generalised.

https://trello.com/c/VEHrvjkj

cc @jacobwinch @johnduffell @davidfurey @joelochlann

### DIFF
--- a/membership-attribute-service/app/controllers/HealthCheckController.scala
+++ b/membership-attribute-service/app/controllers/HealthCheckController.scala
@@ -1,10 +1,11 @@
 package controllers
 
-import play.api.Logger
 import play.api.libs.json.Json
 import play.api.mvc.{Action, Results}
-import components.NormalTouchpointComponents
+import components.NormalTouchpointComponents._
 import com.github.nscala_time.time.Imports._
+import com.typesafe.scalalogging.LazyLogging
+import services.HealthCheckableService
 
 trait Test {
   def ok: Boolean
@@ -16,14 +17,25 @@ class BoolTest(name: String, exec: () => Boolean) extends Test {
   override def ok = exec()
 }
 
-class HealthCheckController extends Results {
-  lazy val zuora = NormalTouchpointComponents.zuoraService
-  lazy val salesforce = NormalTouchpointComponents.contactRepo.salesforce
+class HealthCheckController extends Results with LazyLogging {
 
-  val tests: Seq[Test] = Seq(
-    new BoolTest("ZuoraPing", () => zuora.lastPingTimeWithin(2.minutes)),
-    new BoolTest("Salesforce", () => salesforce.isAuthenticated)
-  )
+  private lazy val wrappedZuoraService = new HealthCheckableService {
+    private val duration = 2.minutes
+    override def serviceName = "ZuoraService"
+    override def checkHealth: Boolean =
+      if (zuoraService.lastPingTimeWithin(duration)) {
+        true
+      } else {
+        logger.error(s"${this.serviceName} has not responded within the last ${duration.getMinutes} minutes.")
+        // Return true rather than false, because we don't want Zuora to fail the healthcheck.
+        // However we do want to log outages because we have an SLA from Zuora where we can claim compensation.
+        true
+      }
+  }
+
+  private lazy val services = Set(wrappedZuoraService, salesforceService, featureToggleService, attrService)
+
+  private lazy val tests = services.map(service => new BoolTest(service.serviceName, () => service.checkHealth))
 
   def healthCheck() = Action {
     Cached(1) {
@@ -32,7 +44,7 @@ class HealthCheckController extends Results {
       if (failures.isEmpty) {
         Ok(Json.obj("status" -> "ok", "gitCommitId" -> app.BuildInfo.gitCommitId))
       } else {
-        failures.flatMap(_.messages).foreach(msg => Logger.warn(msg))
+        failures.flatMap(_.messages).foreach(msg => logger.error(msg))
         ServiceUnavailable("Service Unavailable")
       }
     }
--- a/membership-attribute-service/app/controllers/HealthCheckController.scala
+++ b/membership-attribute-service/app/controllers/HealthCheckController.scala
@@ -33,7 +33,7 @@ class HealthCheckController extends Results with LazyLogging {
       }
   }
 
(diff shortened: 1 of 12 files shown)


membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -1,6 +1,6 @@
 package services
 
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClient
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsync
```
paulbrown1982: I changed this to use the interface not the concrete class. This is in preparation for switching to use the builder as instantiating a AmazonWebServiceClient alone is now deprecated.

membership-attribute-service/app/controllers/HealthCheckController.scala
```diff
@@ -32,7 +44,7 @@ class HealthCheckController extends Results {
       if (failures.isEmpty) {
         Ok(Json.obj("status" -> "ok", "gitCommitId" -> app.BuildInfo.gitCommitId))
       } else {
-        failures.flatMap(_.messages).foreach(msg => Logger.warn(msg))
+        failures.flatMap(_.messages).foreach(msg => logger.error(msg))
```
jacobwinch: I am guessing that this was at warn level to stop us picking up errors in Sentry on every new deployment (this is just a theory though)

membership-attribute-service/app/controllers/HealthCheckController.scala
```diff
@@ -16,14 +17,25 @@ class BoolTest(name: String, exec: () => Boolean) extends Test {
...
 
-  val tests: Seq[Test] = Seq(
-    new BoolTest("ZuoraPing", () => zuora.lastPingTimeWithin(2.minutes)),
-    new BoolTest("Salesforce", () => salesforce.isAuthenticated)
-  )
+  private lazy val wrappedZuoraService = new HealthCheckableService {
+    private val duration = 2.minutes
+    override def serviceName = "ZuoraService"
+    override def checkHealth: Boolean =
+      if (zuoraService.lastPingTimeWithin(duration)) {
```
jacobwinch: I notice that the lastPingTimeWithin function (in the depths of membership-common) is making use of a ScheduledTask (https://github.com/guardian/membership-common/blob/master/src/main/scala/com/gu/zuora/soap/Client.scala#L91). 

Could we add this logging at that level (and pick it up for free in all the other apps which use this task), and take Zuora out of the healthcheck altogether?

membership-attribute-service/app/controllers/HealthCheckController.scala
```diff
@@ -16,14 +17,25 @@ class BoolTest(name: String, exec: () => Boolean) extends Test {
...
+        true
+      } else {
+        logger.error(s"${this.serviceName} has not responded within the last ${duration.getMinutes} minutes.")
+        // Return true rather than false, because we don't want Zuora to fail the healthcheck.
+        // However we do want to log outages because we have an SLA from Zuora where we can claim compensation.
+        true
+      }
+  }
+
+  private lazy val services = Set(wrappedZuoraService, salesforceService, featureToggleService, attrService, behaviourService)
```
jacobwinch: I know it's not specifically part of the retro action, but do we want Salesforce auth on the health check either?

jacobwinch: I would also question whether the behaviourService is really essential functionality for this app

membership-attribute-service/app/controllers/HealthCheckController.scala
```diff
@@ -16,14 +17,25 @@ class BoolTest(name: String, exec: () => Boolean) extends Test {
   override def ok = exec()
 }
 
-class HealthCheckController extends Results {
-  lazy val zuora = NormalTouchpointComponents.zuoraService
-  lazy val salesforce = NormalTouchpointComponents.contactRepo.salesforce
+class HealthCheckController extends Results with LazyLogging {
```
jacobwinch: Do we gain anything from using lazy vals here?

membership-attribute-service/app/controllers/HealthCheckController.scala
paulbrown1982: Ah yes, ok I'll change that.

membership-attribute-service/app/controllers/HealthCheckController.scala
```diff
@@ -16,14 +17,25 @@ class BoolTest(name: String, exec: () => Boolean) extends Test {
...
-    new BoolTest("ZuoraPing", () => zuora.lastPingTimeWithin(2.minutes)),
-    new BoolTest("Salesforce", () => salesforce.isAuthenticated)
-  )
+  private lazy val wrappedZuoraService = new HealthCheckableService {
+    private val duration = 2.minutes
+    override def serviceName = "ZuoraService"
+    override def checkHealth: Boolean =
+      if (zuoraService.lastPingTimeWithin(duration)) {
+        true
+      } else {
```
davidfurey: Might be clearer as
```
if (!zuoraService.lastPingTimeWithin(duration)) { 
  logger... 
}
true
```

membership-attribute-service/app/controllers/HealthCheckController.scala
johnduffell: 😨

membership-attribute-service/app/controllers/HealthCheckController.scala
paulbrown1982: See https://github.com/guardian/membership-common/pull/551

PR #270, opened 2017-11-30
The new healthcheck requires DescribeTable permissions

paulbrown1982:
### Why do we need this? 
The new healthcheck requires DescribeTable permissions.

This should allow https://github.com/guardian/members-data-api/pull/269 to be released.

cc @jacobwinch


paulbrown1982:
Tested and verified in CODE.

PR #272, opened 2017-12-05
Remove deprecated code messages on AWS library usages

paulbrown1982:
Removed most of the deprecated code messages on our use of an old AWS library.

### Why do we need this? 
GDPR / best practice

### The changes 
- Updated the AWS library to match the one brought in by membership-common (1.11.226)
- Removed deprecated message for using AmazonDynamoDBAsyncClient by using the AmazonDynamoDBAsyncClientBuilder instead.
- Removed the old BatchLoader code which isn't necessary anymore now that the Dynamo table is a cache rather than a data store.
- Moved the MembershipAttributesSerializer to a test area as nothing now uses that.
- Removed the second Dynamo client and unused SNS client from Config.scala, and attempted to refactor the AmazonSQSAsyncClient to use the builder, but it's not compatible with AmazonSQSScalaClient. I'll leave that for a new PR which may remove or recover the abandoned cart stuff.

cc @johnduffell @jacobwinch @AWare @JustinPinner @davidfurey


membership-attribute-service/app/BatchLoader.scala
```diff
@@ -1,45 +0,0 @@
...
-          .membersAttributes(file)
-          .filterNot { attrs => existingIds.contains(attrs.UserId) }
-          .map(writeRequest)
-        val loader = new SingleThreadedBatchWriter(table, com.gu.aws.CredentialsProvider)
-        loader.client.withRegion(Regions.EU_WEST_1)
-        loader.run(requests)
-      }
-    })
-  }
-}
```
jacobwinch: 🔪

membership-attribute-service/test/repositories/MembershipAttributesSerializer.scala
```diff
@@ -1,5 +1,6 @@
 package repositories
```
jacobwinch: If nothing uses this then should we just delete it?

paulbrown1982: The DynamoAttributeServiceTest uses it. I think we should keep that test.

PR #274, opened 2017-12-07
Add code subs to the CORS allowed list for mma.

AWare:
Add code subs to the CORS allowed list for mma.


johnduffell:
great!! make code better 👍 🚀

PR #275, opened 2017-12-08
Tighten up logging permissions and set a retention policy

jacobwinch:
### Why do we need this? 
GDPR

### The changes 
* New log group name
* Add the new log group to Cloudformation
* Remove duplicate logging permissions
* Restrict allowed actions (full list is [here](http://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/permissions-reference-cwl.html))
* Restrict creation of log streams and put log events to the new, Cloudformed log group, so that we don't accidentally log events to a group without a retention policy.

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -115,11 +115,6 @@ Resources:
             Resource: "*"
             Effect: Allow
           - Action:
-            - logs:*
-            Resource:
-            - arn:aws:logs:*:*:*
-            Effect: Allow
-          - Action:
             - dynamodb:PutItem
             - dynamodb:GetItem
             - dynamodb:UpdateItem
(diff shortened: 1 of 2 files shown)


lmath:
Good stuff 👍

cloudformation/membership-attribute-service.yaml
```diff
@@ -328,6 +328,11 @@ Resources:
         FromPort: '443'
         ToPort: '443'
         CidrIp: 0.0.0.0/0
+  MembersDataApiLogGroup:
+    Type: "AWS::Logs::LogGroup"
+    Properties:
+      LogGroupName: !Sub members-data-api-${Stage}
+      RetentionInDays: 30
```
paulbrown1982: Please could raise another ticket to tweak this to a value no more than 26 days?! David F or John D can explain why!

davidfurey: My default recommendation for logs is now 14 days

If you have a particular reason to keep it for longer that is fine, but come and chat with me so I can understand your use case.

PR #276, opened 2017-12-11
Slightly better logging

AWare:
Log slightly more so that we can figure out what's going wrong when two seemingly identical records are found to differ.


johnduffell:
good idea 👍

PR #280, opened 2018-01-22
Add a TTL attribute to records in the Membership Attributes Table

jacobwinch:
### Why do we need this? 
We don't want to keep multiple copies of supporter data indefinitely, but at the moment the membership attributes table (which is used to provide redundancy) does not have a TTL on records. 

This change allows us to start adding a TTL attribute to existing records, which means that when we [enable time to live](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-how-to.html), the Dynamo records for users who lapse, cancel or visit the site infrequently will be cleaned up.

### The changes
* Add a TTL attribute to a record if it doesn't already have one
* On subsequent lookups, only update the TTL on a record if it was last updated more than 1 day ago

Note: we shouldn't actually enable TTL on the PROD table until we have decided what to do about ad-free and wallet data.


membership-attribute-service/app/models/Attributes.scala
```diff
@@ -50,7 +50,8 @@ case class Attributes(
   Wallet: Option[Wallet] = None,
   RecurringContributionPaymentPlan: Option[String] = None,
   MembershipJoinDate: Option[LocalDate] = None,
-  DigitalSubscriptionExpiryDate: Option[LocalDate] = None
+  DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
+  TTLTimestamp: Option[Long] = None
```
AWare: Why is this a long?

johnduffell:
> membership attributes table (which is used to provide redundancy)

more like a failover cache rather than redundancy

membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -8,12 +8,12 @@ import com.gu.scanamo.syntax.{set => scanamoSet, _}
...
 import com.typesafe.scalalogging.LazyLogging
 import models.Attributes
-import org.joda.time.LocalDate
+import org.joda.time.{DateTime, LocalDate}
 import play.api.libs.concurrent.Execution.Implicits._
 
 import scala.concurrent.Future
 
-class ScanamoAttributeService(client: AmazonDynamoDBAsync, table: String)
+class ScanamoAttributeService(client: AmazonDynamoDBAsync, table: String, projectedExpiry: => DateTime = Timestamper.projectedExpiryDate)
```
johnduffell: the laziness of this is a bit surprising

jacobwinch:
@johnduffell - that's a good point. When we raise the second PR which [enables TTL for our table](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-table-timetolivespecification.html), we could also add a CloudWatch metric which gets written to if we detect 'old records' in Dynamo, and alert based on that?

membership-attribute-service/test/repositories/DynamoAttributeServiceTest.scala
```diff
@@ -113,8 +119,37 @@ class DynamoAttributeServiceTest(implicit ee: ExecutionEnv) extends Specificatio
...
+      val result = for {
+        _ <- repo.set(existingAttributes)
+        _ <- repo.update(existingAttributes)
+        retrieved <- repo.get("6789")
+      } yield retrieved
+
+      result must be_==(Some(attributesWithTimestamp)).await
+    }
+
+    "only update the TTLTimestamp of the attribute in the table if nothing has changed and the timestamp is old" in {
```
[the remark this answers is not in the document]
jacobwinch: We've moved this logic now - it's happening [here](https://github.com/guardian/members-data-api/blob/lm-jw-dynamo-ttl/membership-attribute-service/app/services/AttributesFromZuora.scala#L85)

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -63,14 +62,14 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
 
       if(fromWhere == "Zuora") {
         attributesFromDynamo map { case (_, dynamoAttributes) =>
-          val canSkipCacheUpdate: Boolean = dynamoAndZuoraAgree(dynamoAttributes, attributes, identityId)
           val zuoraAttributesWithAdfree: Option[Attributes] = attributesWithAdFreeFlagFromDynamo(attributes, dynamoAttributes)
 
-          if(!canSkipCacheUpdate) {
+          val updateRequired = dynamoUpdateRequired(dynamoAttributes, attributes, identityId)
```
johnduffell: I prefer this positive logic, easier to read 💯

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -63,14 +62,14 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
+          if(updateRequired) {
             updateCache(identityId, zuoraAttributesWithAdfree, dynamoAttributeService).onFailure {
```
lmath: I think this means we don't ever write the projected timestamp when we update the dynamo table. I'm looking at this right now.

johnduffell: updateCache returns a Future, but I can't see where we use that future.  This shouldn't really be left hanging even if scala's dubious concurrency model lets us.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -81,6 +80,39 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
     }
   }
 
+
+
+  def dynamoUpdateRequired(dynamoAttributes: Option[Attributes], zuoraAttributes: Option[Attributes], identityId: String, projectedExpiry: => DateTime = DateTime.now().plusDays(14)) = {
```
johnduffell: notsure about projectedExpiry as it doesn't explain what it is, maybe idealExpiry or latestExpiryLimit? not 100% sure

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -81,6 +80,39 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
+        log.info(s"TTL update required for user $identityId")
+        projectedExpiry
+      }  else {
+        log.info(s"No TTL update required for user $identityId")
+        expiry
+      }
+    }.getOrElse {
+      log.info(s"Record for user $identityId has no TTL so setting TTL to $projectedExpiry.")
+      projectedExpiry
+    }
```
johnduffell: not sure if this shows the cases better?
```scala
currentExpiry match { expiry =>
       case Some(expiry) => if (ttlUpdateRequired(expiry)) =>
         log.info(s"TTL update required for user $identityId with current expiry $expiry. Projected expiry is $projectedExpiry.")
         projectedExpiry
       case Some(expiry) =>
         log.info(s"No TTL update required for user $identityId with current expiry $expiry. Projected expiry is $projectedExpiry.")
         expiry
       case None =>
       log.info(s"Record for user $identityId has no TTL so setting TTL to $projectedExpiry.")
       projectedExpiry
     }
```

lmath:
```
aws dynamodb scan --table-name MembershipAttributes-PROD --select COUNT --filter-expression "attribute_not_exists (TTLTimestamp)" --profile membership --region eu-west-1
```

Now returns count = 0

Using this batch delete script
https://gist.github.com/lmath/85ad032d042aa4d1420726d3501b0df5

PR to make TTLTimestamp mandatory still to follow

PR #281, opened 2018-01-25
Add monitoring to ensure that Dynamo TTL cleanup keeps working in PROD

jacobwinch:
### Why do we need this? 

To ensure that we stay GDPR compliant, we want to make sure that the data retention rules that were put in place with https://github.com/guardian/members-data-api/pull/280 don't get accidentally lifted or invalidated. Thanks to @johnduffell for suggesting this clever idea.

### The changes 

* Add a TTL age check when we read attributes from Dynamo
* Publish metric data if the age check fails
* Add alert which listens to the metric to CloudFormation template.

https://trello.com/c/WIIQuQIn/187-data-retention-for-membership-attributes-dynamo-table-members-data-api

### DIFF
--- a/membership-attribute-service/app/services/ScanamoAttributeService.scala
+++ b/membership-attribute-service/app/services/ScanamoAttributeService.scala
@@ -86,7 +86,7 @@ class ScanamoAttributeService(client: AmazonDynamoDBAsync, table: String)
     } yield TtlConversions.secondsToDateTime(ttl).isBefore(oldestAcceptedAge)
     if (tooOld.contains(true)) {
       logger.error(s"Dynamo Attributes for $identityId have an old TTL. The oldest accepted age is: $oldestAcceptedAge - are we still cleaning the table correctly?")
-      metrics.put("Old Dynamo Item", 1)
+      metrics.put("Old Dynamo Item", 1) //referenced in CloudFormation
     }
   }
 
--- a/membership-attribute-service/app/services/ScanamoAttributeService.scala
+++ b/membership-attribute-service/app/services/ScanamoAttributeService.scala
@@ -8,13 +8,15 @@ import com.gu.scanamo.syntax.{set => scanamoSet, _}
 import com.gu.scanamo.update.UpdateExpression
 import com.typesafe.scalalogging.LazyLogging
 import models.Attributes
+import monitoring.Metrics
 import org.joda.time.{DateTime, LocalDate}
 import play.api.libs.concurrent.Execution.Implicits._
-
 import scala.concurrent.Future
 
 class ScanamoAttributeService(client: AmazonDynamoDBAsync, table: String)
     extends AttributeService with LazyLogging {
+
+  val metrics = Metrics("ScanamoAttributeService") //referenced in CloudFormation
 
   def checkHealth: Boolean = client.describeTable(table).getTable.getTableStatus == "ACTIVE"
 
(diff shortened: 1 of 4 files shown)


membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -8,14 +8,16 @@ import com.gu.scanamo.syntax.{set => scanamoSet, _}
...
+import monitoring.Metrics
 import org.joda.time.{DateTime, LocalDate}
 import play.api.libs.concurrent.Execution.Implicits._
-
 import scala.concurrent.Future
 
 class ScanamoAttributeService(client: AmazonDynamoDBAsync, table: String)
     extends AttributeService with LazyLogging {
 
+  val metrics = Metrics("ScanamoAttributeService") //referenced in CloudFormation
```
jacobwinch: Having to copy the string between here and Cloudformation doesn't feel like a great solution (despite this comment, which tries to warn people, I feel like this could get broken quite easily). Does anyone have a better idea?

membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -73,4 +75,25 @@ class ScanamoAttributeService(client: AmazonDynamoDBAsync, table: String)
...
+    // Dynamo typically cleans up within 48 hours of expiry. I'm allowing slightly older expiry dates to avoid false alarms
+    // https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/howitworks-ttl.html
+    val oldestAcceptedAge = currentDate.toDateTimeAtCurrentTime.minusDays(3)
+    val tooOld = for {
+      attributes <- dynamoAttributes
+      ttl <- attributes.TTLTimestamp
+    } yield TtlConversions.secondsToDateTime(ttl).isBefore(oldestAcceptedAge)
+    if (tooOld.contains(true)) {
+      logger.error(s"Dynamo Attributes for $identityId have an old TTL. The oldest accepted age is: $oldestAcceptedAge - are we still cleaning the table correctly?")
+      metrics.put("Old Dynamo Item", 1) //referenced in CloudFormation
```
jacobwinch: As above

lmath:
👍  nice!

membership-attribute-service/app/services/ScanamoAttributeService.scala
paulbrown1982: How about tagging the EC2 instance in the cloudformation, use the AWS API to read the tag from the local instance? But then you have the same problem with the tag name!

jacobwinch: Hmm interesting suggestion, I think I'll go with this approach for now and re-visit if we can think of something cleaner...

johnduffell: my only suggestion would be to overload the name and make it more "id' like rather than a comment, e.g. aws-scanamoAttributeService and metric-oldDynamoItem rather than looking like a generic string description

membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -73,4 +75,25 @@ class ScanamoAttributeService(client: AmazonDynamoDBAsync, table: String)
...
 
   override def delete(userId: String): Future[DeleteItemResult] = run(scanamo.delete('UserId -> userId))
+
+  override def ttlAgeCheck(dynamoAttributes: Option[Attributes], identityId: String, currentDate: LocalDate): Unit = {
+    // Dynamo typically cleans up within 48 hours of expiry. I'm allowing slightly older expiry dates to avoid false alarms
+    // https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/howitworks-ttl.html
+    val oldestAcceptedAge = currentDate.toDateTimeAtCurrentTime.minusDays(3)
+    val tooOld = for {
+      attributes <- dynamoAttributes
+      ttl <- attributes.TTLTimestamp
```
johnduffell: after a while we can change it to if the timestamp is None or if it's too old

johnduffell:
great idea! 👍 🏅

membership-attribute-service/app/services/ScanamoAttributeService.scala
jacobwinch: We still have an outstanding task to add a TTL to all items in the table (will probably wait a few days and see how much happens naturally first). Once this happens I think we'd like to remove the optionality for TTL.

membership-attribute-service/app/services/ScanamoAttributeService.scala
jacobwinch: Yeah that sounds sensible - I guess we could also store these strings in our config files. Then we could group all of the metrics we care about (which makes it easy to delete spurious custom metrics with confidence). It would also probably make it a bit more obvious that changing them might break stuff.

PR #282, opened 2018-01-25
Remove wallet from Attributes

lmath:
### Why do we need this? 
PR [182](https://github.com/guardian/members-data-api/pull/182) added card expiry information to lookups via /user-attributes/features. We will be doing further work on credit card expiry messaging in the future, but not via card expiry stored in the membership-attributes-table. Also, the wallet is no longer being populated/updated as of [PR 256](https://github.com/guardian/members-data-api/pull/256/files) to remove the Salesforce hook. So, this PR removes wallet from Attributes. 

### The changes 

- Remove wallet from Attributes. If there is an existing value for wallet, it remains: [the update function](https://github.com/guardian/members-data-api/blob/master/membership-attribute-service/app/services/ScanamoAttributeService.scala#L48). I think it's worth seeing if it's easy to add on removing the wallet when we add a TTL for each record that doesn't have one.

[On trello](https://trello.com/c/WIIQuQIn/187-data-retention-for-membership-attributes-dynamo-table-members-data-api)

cc @paulbrown1982 @jacobwinch @johnduffell @AWare

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -75,7 +75,7 @@ class AttributeController extends Controller with LoggingWithLogstashFields {
             def customFields(supporterType: String): List[LogField] = List(LogFieldString("lookup-endpoint-description", endpointDescription), LogFieldString("supporter-type", supporterType), LogFieldString("data-source", fromWhere))
 
             attributes match {
-              case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _, _, _)) =>
+              case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _, _)) =>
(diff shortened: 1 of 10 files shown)


jacobwinch:
@paulbrown1982 - what do you think we need to do about existing wallet data in Dynamo - should we be running a script or something to tidy this up?

johnduffell:
image
🎉 👍

paulbrown1982:
Just delete all records with a wallet?

lmath:
I've just run a script to remove any records with a wallet and

```
$ aws dynamodb scan --table-name MembershipAttributes-PROD --select COUNT --filter-expression "attribute_exists (Wallet)" --profile membership --region eu-west-1
```
gives count = 0

I got the ids that have a wallet with:
https://gist.github.com/lmath/a71a2d583c3177909075606264fbd9f2

And then ran this script to do the deletions:
https://gist.github.com/lmath/9036ea278e6a49278e183c37c8feb24d

PR #284, opened 2018-02-02
Delete unused code

jacobwinch:
### Why do we need this? 

This code is unused. The lambda function which used to call this endpoint has been [removed](https://github.com/guardian/support-workers/pull/87). (I also double checked that nothing else is calling it, by zgrepping the last 14 days of fastly logs).

### The changes 
* Delete dead code, which used to be called to insert new contributors into the MembershipAttributes Dynamo table.

https://trello.com/c/yLVBl0V1/1199-delete-unused-code-from-members-data-api


johnduffell:
great work! 🎉

PR #283, opened 2018-01-26
Update sample responses in the README

lmath:
### Why do we need this? 
Sample responses were out of date. They should take into account the digital pack and not include wallet. 

### The changes 

- Updated the sample responses to remove wallet
- Updated the sample responses to include the digital pack

### trello card/screenshot/json/related PRs etc
[on trello](https://trello.com/c/qQ3jljVu/190-tidy-up-members-data-api-github)

cc @paulbrown1982 @jacobwinch @johnduffell @AWare

PR #286, opened 2018-02-08
Make TTLTimestamp a mandatory field in the MembershipAttributes table

lmath:
### Why do we need this? 
As part of GDPR compliance, we need a data retention time limit on persistent data stores which contain personal data. This includes the Membership Attributes table. Currently the TTLTimestamp is an option, and that enabled us to add a TTL in the first place. However, now that every record has a TTLTimestamp, we would like to make this mandatory so that we do not accidentally retain data for longer than intended. 

### The changes 

- Make TTLTimestamp mandatory in the Membership Attributes table
- DynamoAttributes for attributes stored in Dynamo. ZuoraAttributes for attributes determined from calling Zuora. Attributes for the responses returned on lookup with the most up to date info possible (typically from Zuora) but also AdFree which is currently only stored in Dynamo. 

- [on trello](https://trello.com/c/WIIQuQIn/187-data-retention-for-membership-attributes-dynamo-table-members-data-api)
- [PR for new records to include a TTLTimestamp](https://github.com/guardian/members-data-api/pull/280)
- [Monitoring for old TTLTimestamps](https://github.com/guardian/members-data-api/pull/281)


membership-attribute-service/app/sources/SalesforceCSVExport.scala
```diff
@@ -1,26 +0,0 @@
-package sources
```
lmath: I should have removed the SalesforceCSVExport stuff in https://github.com/guardian/members-data-api/pull/283

But I'm doing it here because otherwise I'd be maintaining its reference to Attributes

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -12,60 +10,96 @@ import play.api.mvc.Results.Ok
...
+
+case class DynamoAttributes(UserId: String,
+                            Tier: Option[String] = None,
+                            RecurringContributionPaymentPlan: Option[String] = None,
+                            MembershipJoinDate: Option[LocalDate] = None,
+                            DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
+                            MembershipNumber: Option[String],
+                            AdFree: Option[Boolean],
+                            TTLTimestamp: Long) {
+  lazy val isFriendTier = Tier.exists(_.equalsIgnoreCase("friend"))
```
jacobwinch: All of this code is duplicated, so we could pull it out of the case classes and into a separate trait or object?

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -38,56 +39,53 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
-
-    val attributesFromDynamo: Future[(String, Option[Attributes])] = dynamoAttributeService.get(identityId) map { attributes =>
+    val attributesFromDynamo: Future[Option[DynamoAttributes]] = dynamoAttributeService.get(identityId) map { attributes =>
       dynamoAttributeService.ttlAgeCheck(attributes, identityId, forDate)
-      ("Dynamo", attributes)
+      attributes
     }
 
-    val attributesFromZuora: Future[Disjunction[String, Option[Attributes]]] = attributesDisjunction.run.map {
+    val defaultIfZuoraFails: Future[Option[ZuoraAttributes]] = attributesFromDynamo map { maybeDynamoAttributes => maybeDynamoAttributes map {DynamoAttributes.asZuoraAttributes(_)}}
```
jacobwinch: It feels a bit weird that we're converting DynamoAttributes to ZuoraAttributes...

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -38,56 +39,53 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
+
+    val attributesFromZuora: Future[Option[ZuoraAttributes]] = defaultIfZuoraFails flatMap  { default =>
```
jacobwinch: maybe we could call this fallback, rather than default?

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -38,56 +39,53 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
-              case e: Throwable =>
-                log.error(s"Future failed when attempting to update cache.", e)
-                log.warn(s"Future failed when attempting to update cache. Attributes from Zuora: $attributes")
-            }
-          }
-
-          (fromWhere, zuoraAttributesWithAdfree)
+    def updateIfNeeded(zuoraAttributes: Option[ZuoraAttributes], dynamoAttributes: Option[DynamoAttributes], attributes: Option[Attributes]) = {
+      val updateRequired = dynamoUpdateRequired(dynamoAttributes, zuoraAttributes, identityId, twoWeekExpiry)
+      if(updateRequired) {
```
jacobwinch: tiny point, but I think we're missing a space here

membership-attribute-service/app/services/ScanamoAttributeService.scala
```diff
@@ -58,7 +58,7 @@ class ScanamoAttributeService(client: AmazonDynamoDBAsync, table: String)
       scanamoSetOpt('RecurringContributionPaymentPlan -> attributes.RecurringContributionPaymentPlan),
       scanamoSetOpt('MembershipJoinDate -> attributes.MembershipJoinDate),
       scanamoSetOpt('DigitalSubscriptionExpiryDate -> attributes.DigitalSubscriptionExpiryDate),
-      scanamoSetOpt('TTLTimestamp -> attributes.TTLTimestamp)
+      scanamoSetOpt('TTLTimestamp -> Some(attributes.TTLTimestamp)) //todo stop needlessly option-ing me
```
jacobwinch: 😆

membership-attribute-service/app/models/Attributes.scala
lmath: Yes, I think there is definitely a better way to model this. Do you have some time to work with me on this point? I thought about it earlier but had some trouble with the Attributes writer.

membership-attribute-service/app/sources/SalesforceCSVExport.scala
jacobwinch: 🔪

membership-attribute-service/app/models/Attributes.scala
jacobwinch: Yes sure - we could always remove this duplication in a separate PR though - it shouldn't block this one as it's not really a GDPR requirement!

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -50,6 +50,32 @@ class AttributesMaker extends LazyLogging {
     }
   }
 
+  def attributesAndSource(zuoraAttributes: Option[ZuoraAttributes], dynamoAttributes: Option[DynamoAttributes]): (String, Option[Attributes]) = {
```
jacobwinch: It's a shame that we need this at all now that we have our ZuoraAttributes and DynamoAttributes classes... but I guess we can't get rid of the more generic Attributes until we remove MembershipNumber and AdFree from the responses, and that's one for another PR!

membership-attribute-service/app/services/AttributesFromZuora.scala
jacobwinch: Also I feel like the name 'attributesFromZuora' is a bit weird, given that this val may end up referencing the attributes from Dynamo!

jacobwinch:
This looks good so far! Let me know when you've finished working on it and I'll take another look

membership-attribute-service/app/services/AttributesFromZuora.scala
lmath: Good point. Now I'm just doing this in the dynamoAndZuoraAgree comparison, but I think it makes sense in that context. Well, and the test.

membership-attribute-service/app/services/AttributesFromZuora.scala
lmath: Yep, will do.

lmath: Agreed. I've tried to make this clearer.

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -12,60 +10,87 @@ import play.api.mvc.Results.Ok
...
-  AdFree: Option[Boolean] = None,
-  RecurringContributionPaymentPlan: Option[String] = None,
-  MembershipJoinDate: Option[LocalDate] = None,
-  DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
-  TTLTimestamp: Option[Long] = None
-) {
-
-  require(UserId.nonEmpty)
-
+case class Attributes(UserId: String,
```
johnduffell: generally I stick with the 2 space indentation (and start the fields on the next line), because otherwise the amount of indentation required would change if you renamed the class, also then we will be consistent with the `(...)` indentation as the `{..}` indentation

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -12,60 +10,87 @@ import play.api.mvc.Results.Ok
...
-    (__ \ "digitalSubscriptionExpiryDate").writeNullable[LocalDate] and
-      (__ \ "ttlTimestamp").writeNullable[Long]
-    )(unlift(Attributes.unapply)).addNullableField("digitalSubscriptionExpiryDate", _.latestDigitalSubscriptionExpiryDate).addField("contentAccess", _.contentAccess)
+      (__ \ "tier").writeNullable[String] and
+      (__ \ "recurringContributionPaymentPlan").writeNullable[String] and
+      (__ \ "membershipJoinDate").writeNullable[LocalDate] and
+      (__ \ "digitalSubscriptionExpiryDate").writeNullable[LocalDate] and
+      (__ \ "membershipNumber").writeNullable[String] and
+      (__ \ "adFree").writeNullable[Boolean]
+  )(unlift(Attributes.unapply)).addNullableField("digitalSubscriptionExpiryDate", _.latestDigitalSubscriptionExpiryDate).addField("contentAccess", _.contentAccess)
```
johnduffell: existing code I know, but if you are in anyway and fancied splitting this onto a few physical lines it would help some one else

membership-attribute-service/app/models/Attributes.scala
johnduffell: how about make it a `tier: MaybeTier` with subclasses NoTier,  Friend, Partner, Staff etc and sort out which it is in the readers.  Then the readers can also sort out the ignore caseing so we don't need all this IgnoreCase logic any more.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -174,25 +164,24 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
...
-        case Some(dynamoAttributes) => zuoraAttributes.copy(
-          AdFree = dynamoAttributes.AdFree, //fetched from Dynamo in the Zuora lookup anyway (dynamo is the source of truth)
-          MembershipNumber = dynamoAttributes.MembershipNumber, //I don't think membership number is needed and it comes from Salesforce
-          MembershipJoinDate = dynamoAttributes.MembershipJoinDate.flatMap(_ => zuoraAttributes.MembershipJoinDate), //only compare if dynamo has value
-          TTLTimestamp = dynamoAttributes.TTLTimestamp //TTL only in dynamo
-        )
-        case None => zuoraAttributes
-      }
+  def dynamoAndZuoraAgree(maybeDynamoAttributes: Option[DynamoAttributes], maybeZuoraAttributes: Option[ZuoraAttributes], identityId: String): Boolean = {
+    val dynamoAttributesForComparing: Option[ZuoraAttributes] = maybeDynamoAttributes map { dynamoAttributes =>
```
jacobwinch: pedantic review comment: dynamoAttributesForComparison?

membership-attribute-service/test/services/AttributesMakerTest.scala
```diff
@@ -1,111 +1,164 @@
...
+            UserId = testId,
+            Tier = None,
+            RecurringContributionPaymentPlan = Some("Monthly Contribution"),
+            MembershipJoinDate = None,
+            DigitalSubscriptionExpiryDate = None
+          )
+        )
+
+        val attributes = AttributesMaker.zuoraAttributesWithAddedDynamoFields(Some(zuoraAttributes), None)
+
```
jacobwinch: Could also check that the update cache function has been called exactly once in this scenario? (Or could add a separate test for this if it's not already covered elsewhere)?

lmath: I've tried to check that here:
https://github.com/guardian/members-data-api/blob/dc8b4d6ba96baabc46c69c9e922271ca0cedcc7a/membership-attribute-service/test/services/AttributesFromZuoraTest.scala#L157

Does that cover the scenario you're thinking of?

membership-attribute-service/app/services/AttributesFromZuora.scala
lmath: Sure, that is better

membership-attribute-service/test/services/AttributesMakerTest.scala
jacobwinch: Yep, looks good 👍

PR #287, opened 2018-02-09
remove some  more deprecation warnings

pvighi:
remove actor system related deprecation warning.
I did some basic tests in CODE but could use some help with testing the behaviour controller


johnduffell:
looks great, thanks! there are a few parameters with funny spacing, should be (myVal: String) rather than (myVal:String) but overall looks great
👍

PR #288, opened 2018-02-12
remove aws deprecation warning

pvighi:
remove sqs client deprecation warning and aws-wrap library.
aws deprecated client constructors and now provide a library that return instances of less specific interface, not the specific client class which is what the aws-wrap expects. I could cast it to the correct class and it would work but it would be a risk of future changes to the aws sdk.


membership-attribute-service/app/configuration/Config.scala
```diff
@@ -3,17 +3,14 @@ package configuration
 import java.time.Duration
 
 import com.amazonaws.regions.Regions
-import com.amazonaws.services.sqs.AmazonSQSAsyncClient
+import com.amazonaws.services.sqs.AmazonSQSAsyncClientBuilder
 import com.getsentry.raven.dsn.Dsn
-import com.github.dwhjames.awswrap.sqs.AmazonSQSScalaClient
```
pvighi: even though we are not using this library anymore here I could not remove it from the dependencies since it's still used in tests

paulbrown1982: That library stops us receiving another deprecation warning about using a deprecated SQS Scala client... now that we've moved to a fully DI app this may not be necessary anymore.

membership-attribute-service/test/repositories/MembershipAttributesSerializer.scala
```diff
@@ -1,42 +0,0 @@
...
-  object AttributeNames {
-    val userId = "UserId"
-    val membershipNumber = "MembershipNumber"
-    val tier = "Tier"
-    val membershipJoinDate = "MembershipJoinDate"
-  }
-}
-
-case class MembershipAttributesSerializer(tableName: String)
-  extends DynamoDBSerializer[Attributes] {
```
pvighi: I'm not sure where this serializer was being used.. I removed it and all the tests still passed

membership-attribute-service/app/configuration/Config.scala
pvighi: I removed the library and refactored all the test code that still needed it. That removed aws deprecation warning from tests as well

PR #289, opened 2018-02-13
Make sure ScanamoAttributeServiceTest doesn't need credentials in the default profile

lmath:
### Why do we need this? 
I was getting 
```
cannot create an instance for class repositories.ScanamoAttributeServiceTest
[error]   caused by com.amazonaws.SdkClientException: Unable to load AWS credentials from any provider in the chain
```
when running the tests but they worked if I put membership aws credentials in my default profile. I think we need to give credentials to the builder or else they will look for credentials in the default profile.
### The changes 
- I put the stub credentials back in so that the test still works on a dev box with named profile credentials only


lmath:
See: https://github.com/guardian/membership-common/pull/465

PR #291, opened 2018-02-15
update to Play 2 6

pvighi:
updated play to 2.6.
I will remove the new deprecation warnings (or at least some of them) in another pull request.


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -1,5 +1,6 @@
 package controllers
 import actions._
+import akka.stream.Materializer
```
paulbrown1982: You don't need this import now either.

PR #292, opened 2018-02-16
remove play 2.6 deprecations

pvighi:
i just opened one PR for all deprecations since they are somewhat related and it was mostly removing the deprecated execution contexts. To see the changes individually you can check the individual commits as they are grouped by deprecation warning:

1. stop using deprecated controllers
2. stop using deprecated execution contexts
3. stop using deprecated global Action

### DIFF
--- a/membership-attribute-service/app/components/NormalTouchpointComponents.scala
+++ b/membership-attribute-service/app/components/NormalTouchpointComponents.scala
@@ -1,7 +1,0 @@
-package components
-
-import akka.actor.ActorSystem
-import configuration.Config
-
-
-class NormalTouchpointComponents(system: ActorSystem) extends TouchpointComponents(Config.defaultTouchpointBackendStage)(system)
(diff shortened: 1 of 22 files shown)


membership-attribute-service/app/components/NormalTouchpointComponents.scala
```diff
@@ -1,7 +0,0 @@
-package components
```
pvighi: this file was just left over from the 2.6 migration, it was not being used

membership-attribute-service/app/components/TestTouchpointComponents.scala
```diff
@@ -1,6 +0,0 @@
-package components
-
-import akka.actor.ActorSystem
-import configuration.Config
```
pvighi: this file was just left over from the 2.6 migration, it was not being used

PR #294, opened 2018-02-16
Revert "remove play 2.6 deprecations"

pvighi:
Reverts guardian/members-data-api#292
we are seeing some issues in the log with the update card functionality so we are going back to play 2.5 while we investigate


johnduffell:
👍

PR #295, opened 2018-02-16
Revert "update to Play 2 6"

pvighi:
Reverts guardian/members-data-api#291
We are seeing some issues in the log with the update card functionality so we are going back to play 2.5 while we investigate


johnduffell:
👍

PR #296, opened 2018-02-20
Play 2 6 cors filter

pvighi:
put back play 2.6 and implement cors in a different way.
the first commit is just reverting the revert PR so it should not need review,  so it might be easier to look at the other commits individually


membership-attribute-service/app/wiring/AppLoader.scala
```diff
@@ -43,9 +45,36 @@ class MyComponents(context: Context)
...
+    "/user-attributes/me/mma-membership",
+    "/user-attributes/me/mma-paper")
+
+  val mmaUpdatePaths = Seq(
+    "/user-attributes/me/membership-update-card",
+    "/user-attributes/me/digitalpack-update-card",
+    "/user-attributes/me/paper-update-card",
+    "/user-attributes/me/contribution-update-card",
+    "/user-attributes/me/cancel-regular-contribution",
+    "/user-attributes/me/contribution-update-amount")
```
pvighi: I'm not 100% convinced with having to specify all the endpoints we want to enable CORS here individually, It makes it easier to forget to add new endpoints we might add.

If the endpoints were organized in a way we could just have a couple of general prefixes here it would be better, but sadly they are not

jacobwinch: Yes I'm not sure why we start almost everything with 'user-attributes/me'...

PR #299, opened 2018-02-22
Encrypted dynamo

pvighi:
added a cloudformation for the new encrypted dynamo tables and change the code and original cloudformation to refer to the new ones.
Tested in CODE and it seems to work.
As part of this test the CODE and DEV tables have been created. 

**Before merging**
Before merging this PR we should:

- Create the encrypted prod table running the tables cloudformation

- Copy all data from the non encrypted table into it
- Run the main Cloudformation to switch tables

### DIFF
--- a/cloudformation/membership-attribute-service-tables.yaml
+++ b/cloudformation/membership-attribute-service-tables.yaml
@@ -42,4 +42,4 @@ Outputs:
     Description: supporter attributes fallback table
     Value: !GetAtt [ SupporterAttributesFallbackTable, "Arn" ]
     Export:
-      Name: !Sub ${AWS::StackName}-SupporterAttributesFallbackA
+      Name: !Sub ${AWS::StackName}-SupporterAttributesFallback
--- a/cloudformation/membership-attribute-service-tables.yaml
+++ b/cloudformation/membership-attribute-service-tables.yaml
@@ -8,11 +8,11 @@ Parameters:
 Mappings:
   StageVariables:
     DEV:
-          DynamoWriteCapacity: 1
-          DynamoReadCapacity: 5
+      DynamoWriteCapacity: 1
+      DynamoReadCapacity: 5
     UAT:
-          DynamoWriteCapacity: 1
-          DynamoReadCapacity: 5
+      DynamoWriteCapacity: 1
+      DynamoReadCapacity: 5
     PROD:
       DynamoWriteCapacity: 10
       DynamoReadCapacity: 200
@@ -30,7 +30,7 @@ Resources:
         ReadCapacityUnits:
           Fn::FindInMap: [ StageVariables, { Ref: TouchpointBackend }, DynamoReadCapacity ]
         WriteCapacityUnits:
-          Fn::FindInMap: [ StageVariables, { Ref: TouchpointBackend }, DynamoReadCapacity ]
+          Fn::FindInMap: [ StageVariables, { Ref: TouchpointBackend }, DynamoWriteCapacity ]
       SSESpecification:
         SSEEnabled: true
       TableName: !Sub SupporterAttributesFallback-${TouchpointBackend}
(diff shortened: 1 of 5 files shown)


cloudformation/membership-attribute-service.yaml
```diff
@@ -71,8 +71,8 @@ Mappings:
     PROD:
       NotificationAlarmPeriod: 1200
       InstanceName: PROD:membership-attribute-service
-      DynamoDBTable: arn:aws:dynamodb:*:*:table/MembershipAttributes-PROD
-      DynamoDBTableTestUsers: arn:aws:dynamodb:*:*:table/MembershipAttributes-UAT
+      DynamoDBTable: "Memb-Attributes-Tables-PROD-SupporterAttributesFallback"
+      DynamoDBTableTestUsers: "Memb-Attributes-Tables-UAT-SupporterAttributesFallback"
```
pvighi: these new table names assume that the tables cloudformation stacks were created with the name Memb-Attributes-Tables-[TOUCHPOINT_BACKEND]

cloudformation/membership-attribute-service-tables.yaml
```diff
@@ -0,0 +1,45 @@
...
+      TableName: !Sub SupporterAttributesFallback-${TouchpointBackend}
+      TimeToLiveSpecification:
+        AttributeName: "TTLTimestamp"
+        Enabled: true
+Outputs:
+  SupporterAttributesFallbackTable:
+    Description: supporter attributes fallback table
+    Value: !GetAtt [ SupporterAttributesFallbackTable, "Arn" ]
+    Export:
+      Name: !Sub ${AWS::StackName}-SupporterAttributesFallback
```
pvighi: the output parameter is exported prefixed with the name of the stack which seems to be a common practice and it's helpful for finding where an input parameter to another stack comes from. The only tradeoff that I see is that the name of the stack is manually set when it's created but it is written in the cloudformation templates that include this param so there is some room (but not a lot) for mistakes there.

I could also not use AWS::StackName in the  export name in which case we would have to search for the cfn template that includes the name to identify where an import comes from

cloudformation/membership-attribute-service-tables.yaml
```diff
@@ -0,0 +1,45 @@
...
+  TouchpointBackend:
+    Description: Environment name ( DEV, UAT or PROD)
+    Type: String
+    Default: DEV
+Mappings:
+  StageVariables:
+    DEV:
+          DynamoWriteCapacity: 1
+          DynamoReadCapacity: 5
+    UAT:
```
jacobwinch: Minor point, but the formatting is a bit wonky here

cloudformation/membership-attribute-service-tables.yaml
```diff
@@ -0,0 +1,45 @@
...
+          Fn::FindInMap: [ StageVariables, { Ref: TouchpointBackend }, DynamoReadCapacity ]
+      SSESpecification:
+        SSEEnabled: true
+      TableName: !Sub SupporterAttributesFallback-${TouchpointBackend}
+      TimeToLiveSpecification:
+        AttributeName: "TTLTimestamp"
+        Enabled: true
+Outputs:
+  SupporterAttributesFallbackTable:
+    Description: supporter attributes fallback table
```
jacobwinch: What about: "This Dynamo table is used to provide a failover if we cannot fetch a user's attributes from Zuora."

cloudformation/membership-attribute-service.yaml
```diff
@@ -135,7 +136,8 @@ Resources:
             - dynamodb:BatchGetItem
             - dynamodb:DescribeTable
             Resource:
-              Fn::FindInMap: [ StageVariables, { Ref: Stage }, DynamoDBTableTestUsers ]
+              Fn::ImportValue:
```
jacobwinch: On second thoughts, let's leave the permissions to read/write to the old tables alone. We can remove these when we are happy that we don't need to roll this change back (i.e. when we delete the old, unencrypted tables).

PR #300, opened 2018-02-23
update dynamo table name in readme

pvighi:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Membership Attribute Service
 
-The membership attribute service provides an API for managing and retrieving membership attributes associated with a user. It runs on https://members-data-api.theguardian.com/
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # Membership Attribute Service
 
-The membership attribute service provides an API for managing and retrieving membership attributes associated with a user. It runs on https://members-data-api.theguardian.com/
+The membership attribute service provides an API for managing and retrieving supporter attributes associated with a user. It runs on https://members-data-api.theguardian.com/
```
pvighi: this one was a mistake.. should we keep this linet as it was or should we change it (possibly to something else)

jacobwinch: I think this is OK

PR #301, opened 2018-02-23
Remove CFN references to old Dynamo Tables

jacobwinch:
### Why do we need this?
We started using a new Dynamo table as the fallback in https://github.com/guardian/members-data-api/pull/299, but we left the old policies in place so that we could roll back quickly in the event of a problem.

If we are still happy with this change on Monday morning then we should merge this PR and tidy up the Cloudformation template.

### The changes
* Delete unused mappings
* Delete unused policies

## To do
- [x] Delete MembershipAttributes-DEV table
- [x] Delete MembershipAttributes-UAT table
- [x] Delete MembershipAttributes-PROD table
- [x] Delete encrypted backup of MembershipAttributes table from S3

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -71,8 +71,6 @@ Mappings:
     PROD:
       NotificationAlarmPeriod: 1200
       InstanceName: PROD:membership-attribute-service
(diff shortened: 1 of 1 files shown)


jacobwinch:
Good point @pvighi, I've added it to the PR checklist

PR #302, opened 2018-02-27
On calls to /me, determine if there is an action available for a user with a membership 

lmath:
### Why do we need this? 
We would like to be able to notify a user that there is an action they may need to take to do with their membership (primarily, updating their card details).

Part of https://github.com/guardian/frontend/pull/19176 ensures that calls to /user-attributes/me to update the user's cookies also update an action available cookie. This PR is to add a value `alertAvailableFor="membership"` on calls to /me if a user with a membership needs to update their card details. **Initially, we will just log when we would have returned `alertAvailableFor="membership"` for a user** and then we will later put the calculated value for alertAvailableFor in the response. 

PRs to follow:
- Since this PR just logs the calculated value, one will follow to actually add the value to the response. 
- Potentially we'll need to return the calculated value to test in CODE, so this would follow in another PR
- We still need a change to return the action for the user on calls to /mma-membership, if they have one

### The changes 
- On lookup calls /user-attributes/{me, features, membership}, we now need to make an additional Zuora REST call to get the user's account summary to determine if they have an action available
- Now within the AttributesMaker we calculate if there is an action available for the user based on a list of their zuora accounts and corresponding subscriptions (if any) 

https://github.com/guardian/frontend/pull/19176

cc @paulbrown1982 @jacobwinch @pvighi @johnduffell

### DIFF
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -27,7 +27,7 @@ case class Attributes(
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
   MembershipNumber: Option[String] = None,
   AdFree: Option[Boolean] = None,
-  actionAvailableFor: Option[String] = None) {
+  ActionAvailableFor: Option[String] = None) {
   lazy val isFriendTier = Tier.exists(_.equalsIgnoreCase("friend"))
   lazy val isSupporterTier = Tier.exists(_.equalsIgnoreCase("supporter"))
   lazy val isPartnerTier = Tier.exists(_.equalsIgnoreCase("partner"))
@@ -50,7 +50,7 @@ case class ZuoraAttributes(
   RecurringContributionPaymentPlan: Option[String] = None,
   MembershipJoinDate: Option[LocalDate] = None,
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
-  actionAvailableFor: Option[String] = None)
+  ActionAvailableFor: Option[String] = None)
 
 case class DynamoAttributes(
   UserId: String,
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -27,7 +27,7 @@ case class Attributes(
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
   MembershipNumber: Option[String] = None,
   AdFree: Option[Boolean] = None,
-  ActionAvailableFor: Option[String] = None) {
+  AlertAvailableFor: Option[String] = None) {
   lazy val isFriendTier = Tier.exists(_.equalsIgnoreCase("friend"))
   lazy val isSupporterTier = Tier.exists(_.equalsIgnoreCase("supporter"))
   lazy val isPartnerTier = Tier.exists(_.equalsIgnoreCase("partner"))
@@ -50,7 +50,7 @@ case class ZuoraAttributes(
   RecurringContributionPaymentPlan: Option[String] = None,
   MembershipJoinDate: Option[LocalDate] = None,
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
-  ActionAvailableFor: Option[String] = None)
+  AlertAvailableFor: Option[String] = None)
 
 case class DynamoAttributes(
   UserId: String,
@@ -91,7 +91,7 @@ object Attributes {
       (__ \ "digitalSubscriptionExpiryDate").writeNullable[LocalDate] and
       (__ \ "membershipNumber").writeNullable[String] and
       (__ \ "adFree").writeNullable[Boolean] and
-      (__ \ "actionAvailableFor").writeNullable[String]
+      (__ \ "alertAvailableFor").writeNullable[String]
   )(unlift(Attributes.unapply))
     .addNullableField("digitalSubscriptionExpiryDate", _.latestDigitalSubscriptionExpiryDate)
     .addField("contentAccess", _.contentAccess)
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -26,7 +26,8 @@ case class Attributes(
   MembershipJoinDate: Option[LocalDate] = None,
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
   MembershipNumber: Option[String] = None,
-  AdFree: Option[Boolean] = None) {
+  AdFree: Option[Boolean] = None,
+  actionAvailableFor: Option[String] = None) {
   lazy val isFriendTier = Tier.exists(_.equalsIgnoreCase("friend"))
   lazy val isSupporterTier = Tier.exists(_.equalsIgnoreCase("supporter"))
   lazy val isPartnerTier = Tier.exists(_.equalsIgnoreCase("partner"))
@@ -48,7 +49,8 @@ case class ZuoraAttributes(
   Tier: Option[String] = None,
   RecurringContributionPaymentPlan: Option[String] = None,
   MembershipJoinDate: Option[LocalDate] = None,
-  DigitalSubscriptionExpiryDate: Option[LocalDate] = None)
+  DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
+  actionAvailableFor: Option[String] = None)
 
 case class DynamoAttributes(
   UserId: String,
@@ -88,7 +90,8 @@ object Attributes {
       (__ \ "membershipJoinDate").writeNullable[LocalDate] and
       (__ \ "digitalSubscriptionExpiryDate").writeNullable[LocalDate] and
       (__ \ "membershipNumber").writeNullable[String] and
-      (__ \ "adFree").writeNullable[Boolean]
+      (__ \ "adFree").writeNullable[Boolean] and
+      (__ \ "actionAvailableFor").writeNullable[String]
   )(unlift(Attributes.unapply))
     .addNullableField("digitalSubscriptionExpiryDate", _.latestDigitalSubscriptionExpiryDate)
     .addField("contentAccess", _.contentAccess)
--- a/membership-attribute-service/app/models/CustomerAccount.scala
+++ b/membership-attribute-service/app/models/CustomerAccount.scala
@@ -1,0 +1,7 @@
+package models
+
+import com.gu.memsub.subsv2.Subscription
+import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
+import com.gu.zuora.rest.ZuoraRestService.AccountSummary
+
+case class CustomerAccount(account: AccountSummary, subscription: Option[Subscription[AnyPlan]])
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -5,7 +5,7 @@ import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
 import com.gu.memsub.subsv2.reads.SubPlanReads
 import com.gu.memsub.subsv2.reads.SubPlanReads.anyPlanReads
 import com.gu.scanamo.error.DynamoReadError
-import com.gu.zuora.rest.ZuoraRestService.{AccountSummary, QueryResponse}
+import com.gu.zuora.rest.ZuoraRestService.{AccountSummary, PaymentMethodId, PaymentMethodResponse, QueryResponse}
 import loghandling.LoggingField.LogField
 import loghandling.{LoggingWithLogstashFields, ZuoraRequestCounter}
 import models.{Attributes, DynamoAttributes, ZuoraAttributes}
@@ -26,6 +26,7 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
                     identityIdToAccountIds: String => Future[String \/ QueryResponse],
                     subscriptionsForAccountId: AccountId => SubPlanReads[AnyPlan] => Future[Disjunction[String, List[Subscription[AnyPlan]]]],
                     accountSummaryForAccountId: AccountId => Future[\/[String, AccountSummary]],
+                    paymentMethodForPaymentMethodId: PaymentMethodId => Future[\/[String, PaymentMethodResponse]],
                     dynamoAttributeService: AttributeService,
                     forDate: LocalDate = LocalDate.now()): Future[(String, Option[Attributes])] = {
 
@@ -37,7 +38,7 @@ object AttributesFromZuora extends LoggingWithLogstashFields {
       attributes
     }
 
-    val zuoraAttributesDisjunction: DisjunctionT[Future, String, Option[ZuoraAttributes]] = for {
+    val zuoraAttributesDisjunction: DisjunctionT[Future, String, Future[Option[ZuoraAttributes]]] = for {
       accounts <- EitherT[Future, String, QueryResponse](withTimer(s"ZuoraAccountIdsFromIdentityId", zuoraAccountsQuery(identityId, identityIdToAccountIds), identityId))
       accountIds = queryToAccountIds(accounts)
       accountSummaries <- EitherT[Future, String, List[AccountSummary]](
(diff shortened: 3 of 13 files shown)


membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -21,36 +21,46 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
...
       attributes
     }
 
-    val zuoraAttributesDisjunction: DisjunctionT[Future, String, Option[ZuoraAttributes]] = for {
+    val zuoraAttributesDisjunction: DisjunctionT[Future, String, Future[Option[ZuoraAttributes]]] = for {
       accounts <- EitherT[Future, String, QueryResponse](withTimer(s"ZuoraAccountIdsFromIdentityId", zuoraAccountsQuery(identityId, identityIdToAccountIds), identityId))
       accountIds = queryToAccountIds(accounts)
-      subscriptions <- EitherT[Future, String, List[Subscription[AnyPlan]]](
-        if(accountIds.nonEmpty) withTimer(s"ZuoraGetSubscriptions", getSubscriptions(accountIds, identityId, subscriptionsForAccountId), identityId)
+      accountSummaries <- EitherT[Future, String, List[AccountSummary]](
```
jacobwinch: There's quite a lot going on inside this for comprehension now... it might be cleaner to pull the accountSummaries and/or subscriptions checks into a separate function?

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -21,36 +21,46 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
...
+        if(accountIds.nonEmpty) withTimer(s"ZuoraGetAccountSummaries", getAccountSummaries(accountIds, identityId, accountSummaryForAccountId), identityId)
```
jacobwinch: Is it possible to get the info we need without making an additional REST call? Can we select the fields that we care about as part of the query that we run on line 39 instead?

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -35,18 +44,36 @@ class AttributesMaker extends LazyLogging {
...
-      val membershipJoinDate: Option[LocalDate] = membershipSub.map(_.startDate)
-      val latestDigitalPackExpiryDate: Option[LocalDate] = Some(subsWhichIncludeDigitalPack.map(_.termEndDate)).filter(_.nonEmpty).map(_.max)
-      ZuoraAttributes(
-        UserId = identityId,
-        Tier = tier,
-        RecurringContributionPaymentPlan = recurringContributionPaymentPlan,
-        MembershipJoinDate = membershipJoinDate,
-        DigitalSubscriptionExpiryDate = latestDigitalPackExpiryDate
-      )
+    val maybeMembershipSubAndAccount = membershipSub flatMap { sub: Subscription[AnyPlan] => subsWithAccounts.find(subWithAccount => subWithAccount.subscription == Some(sub)) }
```
jacobwinch: I'd rename this to be 'maybeMembershipAndAccount'

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -35,18 +44,36 @@ class AttributesMaker extends LazyLogging {
...
+
+    val maybeActionAvailable = maybeMembershipSubAndAccount flatMap { subAndAccount: CustomerAccount => subAndAccount.subscription map { sub: Subscription[AnyPlan] =>
+        actionAvailableFor(subAndAccount.account, sub, paymentMethodGetter) map { actionAvailable: Boolean =>
+          if(actionAvailable) Some("membership")
```
jacobwinch: Could we use something more specific like Some("membership_card_update")?

jacobwinch: Ignore this... I just saw that it's called 'ActionAvailableFor' and this makes more sense now.

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -70,12 +98,37 @@ class AttributesMaker extends LazyLogging {
...
+
+    val stillFreshInDays = 27
+    def recentEnough(lastTransationDateTime: DateTime) = lastTransationDateTime.plusDays(stillFreshInDays).isAfterNow
+
+    def actionForSub(accountSummary: AccountSummary, sub: Subscription[AnyPlan]): Future[String \/ Boolean] = {
+      if(accountSummary.balance > 0 && accountSummary.defaultPaymentMethod.isDefined) {
+        val eventualPaymentMethod: Future[\/[String, PaymentMethodResponse]] = paymentMethodGetter(accountSummary.defaultPaymentMethod.get.id)
+        eventualPaymentMethod map { maybePaymentMethod: \/[String, PaymentMethodResponse] =>
+          maybePaymentMethod.map { pm: PaymentMethodResponse =>
+            if(creditCard(pm) && pm.numConsecutiveFailures > 0 && recentEnough(pm.lastTransactionDateTime))
```
jacobwinch: Could extract this into a val called _userInPaymentFailure_?

membership-attribute-service/test/services/AttributesMakerTest.scala
```diff
@@ -1,165 +1,240 @@
...
+    "return none if both Dynamo and Zuora attributes are none" in {
+      AttributesMaker.zuoraAttributesWithAddedDynamoFields(None, None) === None
+    }
 
-      "return none if both Dynamo and Zuora attributes are none" in {
-        AttributesMaker.zuoraAttributesWithAddedDynamoFields(None, None) === None
-      }
+  }
+
+  "actionAvailableFor" should {
```
jacobwinch: I think we're missing a case: "return false for a member who pays via PayPal"?

jacobwinch: Also, for my own understanding more than anything else, does the client decide which subset of paid users (members/subscribers/contributors) to show this message to?

membership-attribute-service/app/services/AttributesFromZuora.scala
jacobwinch: Happy to investigate this as a follow up, rather than slowing down this PR though, as I know we're keen to get this released

membership-attribute-service/test/services/AttributesMakerTest.scala
lmath: I am currently only setting actionAvailableFor for membership. We are already getting current subs and checking if there is a membership sub in 'AttributesMaker' so the filter happens [there](https://github.com/guardian/members-data-api/blob/2107776792c39231cd95fc4eb61842d612cb1eec/membership-attribute-service/app/services/AttributesMaker.scala#L49)

jacobwinch: Thanks for the explanation 👍

membership-attribute-service/app/services/AttributesMaker.scala
lmath: Yep, that's a better idea. Will do.

membership-attribute-service/test/services/AttributesMakerTest.scala
lmath: I will add 'return false for a member who pays via PayPal'. Thanks for spotting that.

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -26,7 +26,8 @@ case class Attributes(
   MembershipJoinDate: Option[LocalDate] = None,
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
   MembershipNumber: Option[String] = None,
-  AdFree: Option[Boolean] = None) {
+  AdFree: Option[Boolean] = None,
+  ActionAvailableFor: Option[String] = None) {
```
johnduffell: is this really Notification rather than action? what does the "for" bit mean?

membership-attribute-service/app/models/CustomerAccount.scala
```diff
@@ -0,0 +1,7 @@
+package models
+
+import com.gu.memsub.subsv2.Subscription
+import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
+import com.gu.zuora.rest.ZuoraRestService.AccountSummary
+
+case class CustomerAccount(account: AccountSummary, subscription: Option[Subscription[AnyPlan]])
```
johnduffell: maybe AccountWithSubscriptions

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -21,36 +21,46 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
   def getAttributes(identityId: String,
                     identityIdToAccountIds: String => Future[String \/ QueryResponse],
                     subscriptionsForAccountId: AccountId => SubPlanReads[AnyPlan] => Future[Disjunction[String, List[Subscription[AnyPlan]]]],
+                    accountSummaryForAccountId: AccountId => Future[\/[String, AccountSummary]],
+                    paymentMethodForPaymentMethodId: PaymentMethodId => Future[\/[String, PaymentMethodResponse]],
                     dynamoAttributeService: AttributeService,
                     forDate: LocalDate = LocalDate.now()): Future[(String, Option[Attributes])] = {
 
+
```
johnduffell: was the extra whitespace intentional?

membership-attribute-service/app/models/Attributes.scala
lmath: https://github.com/guardian/frontend/pull/19176/files
Looks like frontend is actually expecting that /me will have 
`alertAvailableFor=membership` in the response if there needs to be a banner that directs the customer to the membership manage page.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -152,6 +166,23 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
     }
   }
 
+  def getAccountSummaries(accountIds: List[AccountId],
+                       identityId: String,
+                       accountSummaryForAccountId: AccountId => Future[Disjunction[String, AccountSummary]]): Future[Disjunction[String, List[AccountSummary]]] = {
```
johnduffell: better to stick to two space indent, to avoid this problem of misaligned contained inside the brackets 
```scala
def getAccountSummaries(
  accountIds: List[AccountId],
  identityId: String,
  accountSummaryForAccountId: AccountId => Future[Disjunction[String, AccountSummary]]
)
```

membership-attribute-service/app/services/AttributesFromZuora.scala
lmath: It wasn't. I will remove it.

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -4,15 +4,24 @@ import com.github.nscala_time.time.OrderingImplicits._
...
 import scalaz.syntax.std.boolean._
 
-class AttributesMaker extends LazyLogging {
+class AttributesMaker {
 
-  def zuoraAttributes(identityId: String, subs: List[Subscription[AnyPlan]], forDate: LocalDate): Option[ZuoraAttributes] = {
+  def zuoraAttributes(identityId: String,
+                      subsWithAccounts: List[CustomerAccount],
+                      paymentMethodGetter: PaymentMethodId => Future[\/[String, PaymentMethodResponse]],
+                      forDate: LocalDate)(implicit ec: ExecutionContext): Future[Option[ZuoraAttributes]] = {
```
johnduffell: same here
```scala
def zuoraAttributes(
  identityId: String,
  subsWithAccounts: List[CustomerAccount],
  paymentMethodGetter: PaymentMethodId => Future[\/[String, PaymentMethodResponse]],
  forDate: LocalDate
)(implicit ec: ExecutionContext)
```

membership-attribute-service/app/models/Attributes.scala
lmath: Oh, so I'm planning on changing this from ActionAvailableFor -> AlertAvailableFor unless calling it something other than AlertAvailableFor is worth the delay of me also making a change in frontend.

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -70,12 +98,37 @@ class AttributesMaker extends LazyLogging {
...
+        val eventualPaymentMethod: Future[\/[String, PaymentMethodResponse]] = paymentMethodGetter(accountSummary.defaultPaymentMethod.get.id)
+        eventualPaymentMethod map { maybePaymentMethod: \/[String, PaymentMethodResponse] =>
+          maybePaymentMethod.map { pm: PaymentMethodResponse =>
+            if(creditCard(pm) && pm.numConsecutiveFailures > 0 && recentEnough(pm.lastTransactionDateTime))
+              true
+            else false
+          }
+        }
+      }
+      else Future.successful(\/.right(false))
```
johnduffell:
```scala
val eventualPaymentMethod: Future[\/[String, PaymentMethodResponse]] = paymentMethodGetter(accountSummary.defaultPaymentMethod.get.id)
  eventualPaymentMethod map { maybePaymentMethod: \/[String, PaymentMethodResponse] =>
    maybePaymentMethod.map { pm: PaymentMethodResponse =>
      if(creditCard(pm) && pm.numConsecutiveFailures > 0 && recentEnough(pm.lastTransactionDateTime))
        true
      else false
    }
  }
}
```
could be more like
```scala
val eventualPaymentMethod: Future[\/[String, PaymentMethodResponse]] =
  paymentMethodGetter(accountSummary.defaultPaymentMethod.get.id)
  EitherT(eventualPaymentMethod) map { pm: PaymentMethodResponse =>
      creditCard(pm) &&
        pm.numConsecutiveFailures > 0 &&
        recentEnough(pm.lastTransactionDateTime))
    }
  }
}
```
there may be some more possible after that

membership-attribute-service/test/services/AttributesMakerTest.scala
```diff
@@ -1,165 +1,240 @@
...
-      "return up to date Zuora attributes when they match the dynamo attributes" in {
-        val zuoraAttributes = ZuoraAttributes(
-          UserId = testId,
-          Tier = None,
-          RecurringContributionPaymentPlan = Some("Monthly Contribution"),
-          MembershipJoinDate = None,
-          DigitalSubscriptionExpiryDate = None
-        )
+    // We are currently returning actionAvailableFor = None always in AttributesMaker and just logging the value we've calculated
+    //This test should be uncommented once we resume returning the calculated value
```
johnduffell: I think you can use an @Ignore annotation to have the test there (and compiled) but reported as skipped

johnduffell:
sorry most of my comments are about layout and naming! but I think it's a good opprtunity to make the code clearer for the next people.  it's a bit early in the morning for some detailed look at the logic but I'll look again later

membership-attribute-service/test/services/AttributesMakerTest.scala
lmath: You've inspired me to finally look up how to do this in specs2 tests 🙂

membership-attribute-service/app/services/AttributesFromZuora.scala
lmath: Agreed. I've pulled the empty checks out of the for comprehension.

lmath:
Thanks for your comments @jacobwinch @johnduffell! Do you have some time to take another look at my PR?

jacobwinch:
Yes @lmath, I'll have another look now.

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -35,18 +44,36 @@ class AttributesMaker extends LazyLogging {
...
+    val maybeAlertAvailable = maybeMembershipAndAccount flatMap { subAndAccount: AccountWithSubscription => subAndAccount.subscription map { sub: Subscription[AnyPlan] =>
+        alertAvailableFor(subAndAccount.account, sub, paymentMethodGetter) map { alertAvailable: Boolean =>
+          if(alertAvailable) Some("membership")
+          else None
+        }
+      }
+    }
+
+    maybeAlertAvailable.getOrElse(Future.successful(None)) map { maybeAction =>
+      maybeAction map { action => SafeLogger.info(s"User $identityId has an action available: $action") }
```
jacobwinch: should all mentions of 'action' here be 'alert' now?

membership-attribute-service/test/services/AttributesMakerTest.scala
```diff
@@ -1,165 +1,250 @@
...
-          UserId = testId,
-          Tier = None,
-          RecurringContributionPaymentPlan = Some("Monthly Contribution"),
-          MembershipJoinDate = None,
-          DigitalSubscriptionExpiryDate = None
+    // We are currently returning alertAvailableFor = None always in AttributesMaker and just logging the value we've calculated
+    //This test should be un-ignored once we resume returning the calculated value
+
+    "return alertAvailableFor=membership for an active membership in payment failure" in {
+      skipped {
```
jacobwinch: 👍

membership-attribute-service/test/testdata/AccountSummaryTestData.scala
```diff
@@ -0,0 +1,29 @@
+package testdata
+
+import com.gu.memsub.Subscription.AccountId
+import com.gu.zuora.rest.ZuoraRestService.{AccountSummary, BillToContact, DefaultPaymentMethod, PaymentMethodId, SoldToContact}
+
+object AccountSummaryTestData { //todo should this be an object?
```
jacobwinch: Seems reasonable to me - you can probably remove this todo!

membership-attribute-service/app/services/AttributesMaker.scala
lmath: Yes. Thanks for catching that!

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -31,26 +34,30 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
       attributes
     }
 
-    val zuoraAttributesDisjunction: DisjunctionT[Future, String, Option[ZuoraAttributes]] = for {
+    def getResultIf[R](shouldCallFunction: Boolean, whichCall: String, futureResult: Future[Disjunction[String, R]], default: R, identityId: String): Future[Disjunction[String, R]] = {
```
pvighi: do you need identity id and default as params here?
it seems like it always uses the identity id in scope and nil as the default

PR #305, opened 2018-03-01
Add some custom fields to the logging for the payment failure banner

lmath:
https://github.com/guardian/members-data-api/pull/302 adds a log when we would have shown a member a banner asking them to take an action on their account. 

This PR adds some custom fields to the logging so that it's easier to query these logs. 

### Why do we need this? 
We would like to be able to compare the users we would have shown the banner to and see if we consider them to be in payment failure. These custom fields will make that easier. 

### The changes 
- Add some custom fields to the logging of when we would have shown the member a banner 

https://github.com/guardian/members-data-api/pull/302

cc @paulbrown1982 @johnduffell @pvighi

### DIFF
--- a/membership-attribute-service/app/services/AttributesMaker.scala
+++ b/membership-attribute-service/app/services/AttributesMaker.scala
@@ -58,7 +58,7 @@ class AttributesMaker extends LoggingWithLogstashFields{
     maybeAlertAvailable.getOrElse(Future.successful(None)) map { maybeAlert =>
       def customFields(identityId: String, alertAvailableFor: String) = List(LogFieldString("identity_id", identityId), LogFieldString("alert_available_for", alertAvailableFor))
 
-      maybeAlert map { alert => logInfoWithCustomFields(s"User $identityId has an alert available: $alert", customFields(identityId, alert)) }
+      maybeAlert foreach { alert => logInfoWithCustomFields(s"User $identityId has an alert available: $alert", customFields(identityId, alert)) }
 
       hasAttributableProduct.option {
         val tier: Option[String] = membershipSub.flatMap(getCurrentPlans(_).headOption.map(_.charges.benefits.head.id))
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -55,7 +56,9 @@ class AttributesMaker {
     }
 
     maybeAlertAvailable.getOrElse(Future.successful(None)) map { maybeAlert =>
-      maybeAlert map { alert => SafeLogger.info(s"User $identityId has an alert available: $alert") }
+      def customFields(identityId: String, alertAvailableFor: String) = List(LogFieldString("identity_id", identityId), LogFieldString("alert_available_for", alertAvailableFor))
+
+      maybeAlert map { alert => logInfoWithCustomFields(s"User $identityId has an alert available: $alert", customFields(identityId, alert)) }
```
johnduffell: you might get a warning from this - you can use foreach instead of map to avoid that

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -55,7 +56,9 @@ class AttributesMaker {
     }
 
     maybeAlertAvailable.getOrElse(Future.successful(None)) map { maybeAlert =>
-      maybeAlert map { alert => SafeLogger.info(s"User $identityId has an alert available: $alert") }
+      def customFields(identityId: String, alertAvailableFor: String) = List(LogFieldString("identity_id", identityId), LogFieldString("alert_available_for", alertAvailableFor))
```
johnduffell: we already use camelcase elsewhere for identity id:
https://membership-logs.gutools.co.uk/app/kibana#/discover?_g=()&_a=(columns:!(identityId),index:AWAsTQ6MXSopLsUBjTYb,interval:auto,query:(match_all:()),sort:!('@timestamp',desc))

lmath: Ooops, thanks for catching that.

johnduffell:
great, go for it.  Looks like we are very inconsistent of field names, not sure if it's best to standardise on underscore format, so change this one at the same time to match if you think better?
https://github.com/guardian/members-data-api/blob/c800b89372d7b602107faf10953a3acc992cdcfb/membership-attribute-service/app/services/AttributesFromZuora.scala#L195
either way great stuff 👍

PR #306, opened 2018-03-02
Reduce number of calls to Zuora to serve /user-attributes/me

paulbrown1982:
This change saves the /user-attributes/me endpoint from having to make additional calls to get the account summary, and then to inspect the type payment method for each of those accounts. Now we get our list of account Ids to search for subscriptions with, but also in the same record have the account balance, payment method ID for the latter call. This change saves looking up the payment method if the gateway is not one of our Credit Card ones, as only accounts with those gateways can be updated in profile.theguardian.com.

See https://github.com/guardian/membership-common/pull/568

cc @lmath @johnduffell @pvighi @jacobwinch


membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -44,10 +46,8 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
     }
 
     val zuoraAttributesDisjunction: DisjunctionT[Future, String, Future[Option[ZuoraAttributes]]] = for {
-      accounts <- EitherT[Future, String, QueryResponse](withTimer(s"ZuoraAccountIdsFromIdentityId", zuoraAccountsQuery(identityId, identityIdToAccountIds), identityId))
-      accountIds = queryToAccountIds(accounts)
-      accountSummaries <- EitherT[Future, String, List[AccountSummary]](getResultIf(accountIds.nonEmpty, "ZuoraGetAccountSummaries", getAccountSummaries(accountIds, identityId, accountSummaryForAccountId), Nil, identityId))
```
johnduffell: great, that will be a good improvement in calls made

paulbrown1982: Back to what it was before, where we're pretty content that it is a stable baseline.

lmath:
Thanks for doing this. Looks great.  👍

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -44,10 +46,8 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
...
-      subscriptions <- EitherT[Future, String, List[AccountWithSubscription]](getResultIf(accountSummaries.nonEmpty, "ZuoraGetSubscriptions", getSubscriptions(accountSummaries, identityId, subscriptionsForAccountId), Nil, identityId))
+      accounts <- EitherT[Future, String, GetAccountsQueryResponse](withTimer(s"ZuoraAccountIdsFromIdentityId", zuoraAccountsQuery(identityId, identityIdToAccountIds), identityId))
+      subscriptions <- EitherT[Future, String, List[AccountWithSubscription]](getResultIf(accounts.size > 0, "ZuoraGetSubscriptions", getSubscriptions(accounts, identityId, subscriptionsForAccountId), Nil, identityId))
```
pvighi: I'm still trying to figure out if getResultIf works as expected or not. It seems to me that the param FutureResult (in this case getSubscriptions(..) ) should be made lazy to make sure it doesn't execute in case the condition is false.

johnduffell: agree, the future will start straight away, the condition is whether we time and log the result.  Although we don't start the timer until we enter the timing method, by which time it will have already done some of the processing.

I think this was originally added to stop cluttering up the logs as much, the name is kind of true, but I think it's quite misleading (we don't "get" the result if false, but we do calculate it regardless)

paulbrown1982:
This has been tested in CODE with user: 100000250 and works fine.

johnduffell:
👍 given most callers are identity users without zuora accounts, we will make a good saving if we fix the unrelated issue we discovered some time next week

johnduffell:
PS thanks for fixing this so fast!

paulbrown1982:
Tested in PROD and it's working fine: 
picture 379

PR #307, opened 2018-03-02
Be sure that Zuora call futures don't execute early

paulbrown1982:
Ensure the Zuora futures are only actually run inside the timer function and not because it's dereferenced in the for comprehension.

cc @pvighi @johnduffell @davidfurey

### DIFF
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -1,12 +1,10 @@
 package services
-import com.gu.i18n.Currency
 import com.gu.memsub.Subscription.AccountId
 import com.gu.memsub.subsv2.Subscription
 import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
 import com.gu.memsub.subsv2.reads.SubPlanReads
 import com.gu.memsub.subsv2.reads.SubPlanReads.anyPlanReads
 import com.gu.scanamo.error.DynamoReadError
-import com.gu.zuora.rest.ZuoraRestService
 import com.gu.zuora.rest.ZuoraRestService.{AccountObject, AccountSummary, GetAccountsQueryResponse, PaymentMethodId, PaymentMethodResponse}
 import loghandling.LoggingField.LogField
 import loghandling.{LoggingWithLogstashFields, ZuoraRequestCounter}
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -182,10 +182,11 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
     }
   }
 
-  private def withTimer[R](whichCall: String, futureResult: Future[Disjunction[String, R]], identityId: String): Future[Disjunction[String, R]] = {
+  private def withTimer[R](whichCall: String, executeFuture: () => Future[Disjunction[String, R]], identityId: String): Future[Disjunction[String, R]] = {
     import loghandling.StopWatch
     val stopWatch = new StopWatch
 
+    val futureResult = executeFuture()
     futureResult.map { disjunction: Disjunction[String, R] =>
```
davidfurey: Since this is only about the side-effect, would `.onComplete` make more sense?

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -1,23 +1,22 @@
...
 import models.{AccountWithSubscription, Attributes, DynamoAttributes, ZuoraAttributes}
 import org.joda.time.{DateTime, LocalDate}
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
 import scalaz.std.list._
 import scalaz.std.scalaFuture._
 import scalaz.syntax.traverse._
-import scalaz.{-\/, Disjunction, EitherT, \/, \/-, _}
+import scalaz.{-\/, Disjunction, DisjunctionT, EitherT, \/, \/-}
```
johnduffell: I[m' sure Disjunction is an alias of \/ anyway

johnduffell:
👍 thanks for updating this one!

membership-attribute-service/app/services/AttributesFromZuora.scala
johnduffell: https://github.com/scalaz/scalaz/blob/scalaz-seven/core/src/main/scala/scalaz/package.scala#L299

PR #309, opened 2018-03-08
Return text for the mma-membership page, if relevant

lmath:
### Why do we need this? 
We would like to encourage members in payment failure to update their card details by displaying a banner on site. Once they reach the membership manage page, we need to be able to ask them to update their details if relevant. This change returns alertText if we have an alert for that member because they are in payment failure. 

### The changes 
- On calls to /me/mma-membership, if there is an alert available for the member because they are in payment failure, we determine what alertText should be in the response and then log information about it. Once we are confident that members who would see this message should be seeing it, then we will start returning alertText, if relevant, in the response. 

https://github.com/guardian/members-data-api/pull/302

I'm going to leave a few comments where I already know I need some input!

cc @paulbrown1982 @johnduffell @jacobwinch @AWare 

image


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -244,4 +252,10 @@ object OptionEither {
   def liftFutureEither[A](x: Option[A]): OptionT[FutureEither, A] =
     apply(Future.successful(\/.right[String,Option[A]](x)))
 
+  def liftEither[A](x: Future[Option[A]])(implicit ex: ExecutionContext): OptionT[FutureEither, Option[A]] = {
```
lmath: Still not sure of the best approach so that alertText can be None without halting the for comprehension, but this works in CODE. Suggestions very very welcome. Whenever I try to wrangle the types right in the for comprehension I can't get it to compile. 😬

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -0,0 +1,85 @@
...
+        case Some(id) => {
+          val paymentMethod: Future[Disjunction[String, PaymentMethodResponse]] = paymentMethodGetter(id) fallbackTo Future.successful(\/.left("Failed to get payment method"))
+          paymentMethod map { paymentMethodDisjunction: Disjunction[String, PaymentMethodResponse] => paymentMethodDisjunction map { pm => pm.lastTransactionDateTime.toString(formatter)}}
+        }
+        case None => Future.successful(\/.left("No payment method id and so no payment method!"))
+      }
+
+      paymentMethodLatestDateFormatted.map { formattedDateDisjunction: Disjunction[String, String] =>
+        val alertTextWithDate = formattedDateDisjunction map { date: String =>
+          val productName = subscription.plan.name
```
lmath: TODO: find the right place to get productName. Trying this out in CODE gave me:
```
"alertText": "Our attempt to take payment for your Non Founder Supporter - annual membership failed on 8 March 2018. Please check below that your card details are up to date."
```

membership-attribute-service/app/controllers/AccountController.scala
johnduffell: looks fine for this approach, and it should work.  Splitting it out is a good approach, as for comprehensions should just be (in throey) creating a data structure showing how each part of the program depends on the rest!

In terms of polish, we don't care that it's already an option, A is fine, also we are lifting the value inside the future into EitherOption not just Either so the name could be tweaked too:
```scala
def liftEitherOption[A](future: Future[A])(implicit ex: ExecutionContext): OptionT[FutureEither, A] = {
  apply(future map { value: A =>
    \/.right[String, Option[A]](Some(value))
  })
}
```

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -8,13 +8,21 @@ import play.api.mvc.Results.Ok
 
 object AccountDetails {
 
-  implicit class ResultLike(details: (Contact, PaymentDetails, String)) {
+  implicit class ResultLike(details: (Contact, PaymentDetails, String, Option[String])) {
```
johnduffell: can we make this a case class instead of a tuple, then we can give a helpful name (other than _3) to that String and Option[String]?

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -0,0 +1,85 @@
...
+    AccountObject(
+      Id = accountSummary.id,
+      Balance = accountSummary.balance,
+      Currency = accountSummary.currency,
+      DefaultPaymentMethodId = accountSummary.defaultPaymentMethod.map(_.id),
+      PaymentGateway = accountSummary.billToContact.country.map(RegionalStripeGateways.getGatewayForCountry)
+    )
+
+
+  def membershipAlertText(accountSummary: AccountSummary, subscription: Subscription[AnyPlan],
```
johnduffell: I'd do it as one parameter per line, indended two spaces.  Then renaming the method won't mess any alighnment, also adding/removing parameters will be fairly clear.

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -0,0 +1,85 @@
...
+                          paymentMethodGetter: PaymentMethodId => Future[String \/ PaymentMethodResponse])(implicit ec: ExecutionContext) : Future[Option[String]] = {
+
+    def expectedAlertText: Future[Option[String]] = {
+      val formatter = DateTimeFormat.forPattern("d MMMM yyyy").withLocale(Locale.ENGLISH)
+      val paymentMethodLatestDateFormatted: Future[String \/ String] = accountSummary.defaultPaymentMethod.map(_.id) match {
+        case Some(id) => {
+          val paymentMethod: Future[Disjunction[String, PaymentMethodResponse]] = paymentMethodGetter(id) fallbackTo Future.successful(\/.left("Failed to get payment method"))
+          paymentMethod map { paymentMethodDisjunction: Disjunction[String, PaymentMethodResponse] => paymentMethodDisjunction map { pm => pm.lastTransactionDateTime.toString(formatter)}}
```
johnduffell: if you just want to map through both layers, consider wrap in EitherT, then map, then do .run to get it back again.  May be neater?

johnduffell: I'd rather leave the formatting to a string until the last moment, keep it as a proper type. This avoids having a String \/ String type too, which reduces the chance of error.
so just remove the `.toString(formatter)` from here, fix up the types, and do it on line 44/45 instead

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -0,0 +1,85 @@
...
+        }
+        case None => Future.successful(\/.left("No payment method id and so no payment method!"))
+      }
+
+      paymentMethodLatestDateFormatted.map { formattedDateDisjunction: Disjunction[String, String] =>
+        val alertTextWithDate = formattedDateDisjunction map { date: String =>
+          val productName = subscription.plan.name
+          s"Our attempt to take payment for your $productName membership failed on $date. Please check below that your card details are up to date."
+        }
+        alertTextWithDate.toOption
```
johnduffell: did we do anything with the -\/ side? if not should we, or should we just make at an option the whole way through?

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+
+    alertAvailableFor(accountObject(accountSummary), subscription, paymentMethodGetter) flatMap  { shouldShowAlert: Boolean =>
+      expectedAlertText.map { someText => shouldShowAlert.option (someText).flatten }
```
johnduffell: rather than turning the shouldShowAlert into an option that we're going to flatten straight away, consider just an if statement in the flatmap may be clearer:
if (shouldShowAlert) expectedAlertText
else Future.successful(None)

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    val userInPaymentFailure: Future[\/[String, Boolean]] = {
+      if(account.Balance > 0 && account.DefaultPaymentMethodId.isDefined && isActionablePaymentGateway) {
+        val eventualPaymentMethod: Future[\/[String, PaymentMethodResponse]] = paymentMethodGetter(account.DefaultPaymentMethodId.get)
+
+        eventualPaymentMethod map { maybePaymentMethod: \/[String, PaymentMethodResponse] =>
+          maybePaymentMethod.map { pm: PaymentMethodResponse =>
+            creditCard(pm) &&
+              pm.numConsecutiveFailures > 0 &&
+              recentEnough(pm.lastTransactionDateTime)
```
johnduffell: so we don't actually check the dates or status of the subscription, we just care about the account balance and status, which sounds sensible to me.  Matt had a query whether we were pulling in people we shouldn't, but this looks reasonable to me.

johnduffell:
looks good to me, a few comments just to tidy things up, plus the product thing that you pointed out, overall good work!  Just feel free to use more lines to break things down more, and it's also no problem to even split something into a separate object to group related functions together if it helps.

membership-attribute-service/app/services/PaymentFailureAlerter.scala
AWare: Had a look and it's `subscription.plan.productName`

membership-attribute-service/app/services/PaymentFailureAlerter.scala
jacobwinch: Yeah I think me and @lmath discussed this, and decided that if you cancelled in the future but still owed us money for a recent invoice, it was probably still correct to show the banner

membership-attribute-service/app/services/PaymentFailureAlerter.scala
lmath: Thanks @AWare !

membership-attribute-service/app/controllers/AccountController.scala
lmath: Good point! Thanks.

membership-attribute-service/app/models/AccountDetails.scala
lmath: Good shout. I will do this.

membership-attribute-service/app/services/PaymentFailureAlerter.scala
lmath: Yeah, I think it makes more sense to have it be an option earlier. I just want to err on the side of just not coming up with a membershipAlertText if the information from the message isn't available.

paulbrown1982:
👍 LGTM! Nice work!

PR #311, opened 2018-03-15
Fix recognition of Paper+ and Voucher+ subscribers

jacobwinch:
Previously Paper+ and Voucher+ users were:

"digitalPack":false

Now they are:

"digitalPack":true

I will also re-factor the code to make it more obvious when things are going wrong at the membership-common level, but I don't have much time today and I want to get this fix out first.

### The changes 
* Use the [latest membership-common](https://github.com/guardian/membership-common/pull/570), which includes the fix.


johnduffell:
👍

PR #312, opened 2018-03-21
More accurately identify users who would see a payment failure banner

lmath:
### Why do we need this? 
We started logging who we wanted to show a payment failure banner to in https://github.com/guardian/members-data-api/pull/302. ~10% of these members were not actually in payment failure and this appears to primarily a case of cancelled memberships or that it the last payment date was within 27 days but the last invoice date was not. This PR is to address this discrepancy, and then we can see if using the LastInvoiceDate on calls to /me is correct often enough. 

### The changes 

1. Changing how we determine if a payment failure is recent enough to show the user a banner. 

- **On /me** We will use the latest invoice date (This may not be exactly correct in some cases, but saves adding an extra call to zuora). 

- **On /me/mma-membership** We will use the date of the latest unpaid invoice

2. For a user to be shown the banner/the message on the mma manage page, now their membership cannot be cancelled. 

https://github.com/guardian/membership-common/pull/571 - keep invoices and invoice dates 
https://github.com/guardian/members-data-api/pull/302 - identify and log if a member has an action on calls to /me
https://github.com/guardian/members-data-api/pull/309 - identify and log if a member has an alert for the membership manage page

cc @paulbrown1982 @AWare @jacobwinch @johnduffell @pvighi

### DIFF
--- a/membership-attribute-service/app/services/PaymentFailureAlerter.scala
+++ b/membership-attribute-service/app/services/PaymentFailureAlerter.scala
@@ -5,7 +5,7 @@ import java.util.Locale
 import com.gu.memsub.subsv2.Subscription
 import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
 import com.gu.zuora.api.{RegionalStripeGateways, StripeAUMembershipGateway, StripeUKMembershipGateway}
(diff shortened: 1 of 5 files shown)


johnduffell:
great work! 👍 thanks

PR #322, opened 2018-04-20
Use the correct S3 permissions

jacobwinch:
### Why do we need this? 
We don't want to use the AmazonEC2RoleforSSM role any more, as we found it was too permissive. This PR also makes our CFN a bit stricter about which S3 buckets this app can access, which helps us to avoid accidentally granting access to inappropriate buckets.

### The changes
* Delete a bunch of instance types from allowed values (the list seemed excessive)
* Replace AmazonEC2RoleforSSM with custom guardian-ec2-role-for-ssm
* Explicitly deny access to all S3 resources other than those specified in ReadableS3Resources

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -16,23 +16,6 @@ Parameters:
     - m3.medium
     - m3.large
     - m3.xlarge
-    - m3.2xlarge
-    - c3.large
-    - c3.xlarge
-    - c3.2xlarge
-    - c3.4xlarge
-    - c3.8xlarge
-    - g2.2xlarge
-    - r3.large
-    - r3.xlarge
-    - r3.2xlarge
-    - r3.4xlarge
-    - r3.8xlarge
-    - i2.xlarge
-    - i2.2xlarge
-    - i2.4xlarge
-    - i2.8xlarge
-    - hs1.8xlarge
     ConstraintDescription: must be a valid EC2 instance type.
   MaxInstances:
     Description: Maximum number of instances. This should be (at least) double the
@@ -75,6 +58,9 @@ Mappings:
       DynamoDBTableTestUsers: "Memb-Attributes-Tables-UAT-SupporterAttributesFallback"
       DynamoDBFeatureToggleTable: arn:aws:dynamodb:*:*:table/MembershipFeatureToggles-PROD
       DynamoDBFeatureToggleTableTestUsers: arn:aws:dynamodb:*:*:table/MembershipFeatureToggles-UAT
+      ReadableS3Resources:
+        - arn:aws:s3:::gu-membership-attribute-service-dist/*
+        - arn:aws:s3:::gu-reader-revenue-private/membership/members-data-api/PROD/*
 
     CODE:
       NotificationAlarmPeriod: 1200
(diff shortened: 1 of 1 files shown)


cloudformation/membership-attribute-service.yaml
```diff
@@ -104,12 +93,14 @@ Resources:
       - PolicyName: root
         PolicyDocument:
           Statement:
-          - Action: s3:GetObject
-            Resource: arn:aws:s3:::gu-membership-*/*
-            Effect: Allow
-          - Action: s3:GetObject
-            Resource: arn:aws:s3:::gu-reader-revenue-private/${Stack}/members-data-api/${Stage}/*
```
jacobwinch: I think this is actually broken (missing a !Sub), but we never actually noticed because AmazonEC2RoleforSSM provides access to the bucket.

cloudformation/membership-attribute-service.yaml
```diff
@@ -16,23 +16,6 @@ Parameters:
...
-    - c3.xlarge
-    - c3.2xlarge
-    - c3.4xlarge
-    - c3.8xlarge
-    - g2.2xlarge
-    - r3.large
-    - r3.xlarge
-    - r3.2xlarge
-    - r3.4xlarge
-    - r3.8xlarge
```
jacobwinch: I don't think we're ever going to use these instance types

johnduffell: always good to keep your options open

cloudformation/membership-attribute-service.yaml
```diff
@@ -75,6 +58,9 @@ Mappings:
       DynamoDBTableTestUsers: "Memb-Attributes-Tables-UAT-SupporterAttributesFallback"
       DynamoDBFeatureToggleTable: arn:aws:dynamodb:*:*:table/MembershipFeatureToggles-PROD
       DynamoDBFeatureToggleTableTestUsers: arn:aws:dynamodb:*:*:table/MembershipFeatureToggles-UAT
+      ReadableS3Resources:
+        - arn:aws:s3:::gu-membership-attribute-service-dist/*
```
pvighi: why not arn:aws:s3:::gu-membership-attribute-service-dist/membership/PROD ?

jacobwinch: Good point - fixed

jacobwinch:
I'll ship this on Monday

cloudformation/membership-attribute-service.yaml
```diff
@@ -75,6 +58,9 @@ Mappings:
       DynamoDBTableTestUsers: "Memb-Attributes-Tables-UAT-SupporterAttributesFallback"
       DynamoDBFeatureToggleTable: arn:aws:dynamodb:*:*:table/MembershipFeatureToggles-PROD
       DynamoDBFeatureToggleTableTestUsers: arn:aws:dynamodb:*:*:table/MembershipFeatureToggles-UAT
+      ReadableS3Resources:
+        - arn:aws:s3:::gu-membership-attribute-service-dist/PROD/*
```
pvighi: this should probably be arn:aws:s3:::gu-membership-attribute-service-dist/membership/PROD/*

jacobwinch: Thanks - I will re-test this on CODE before I release this

PR #323, opened 2018-04-25
Enable alerts for contributions

pvighi:
We need to add support for contributions in members data api responses related to the payment failure banner in order to alert users about their failed monthly contribution payments.

Before merging this we need to merge the [frontend PR](https://github.com/guardian/frontend/pull/19579) that adds support for contribution messages. 

trello card:
https://trello.com/c/4oLsE7a0/308-roll-out-banner-to-contributions-e-6

### DIFF
--- a/membership-attribute-service/app/services/AttributesMaker.scala
+++ b/membership-attribute-service/app/services/AttributesMaker.scala
@@ -50,7 +50,8 @@ class AttributesMaker extends LoggingWithLogstashFields{
       val firstSubPerProduct = groupedAccountWithNonEmptySubs.collect { case (Some(k), sub :: _) => (k, sub) }
       val alertableSubs = firstSubPerProduct.filterKeys(product => alertableProducts.contains(product))
       val results = alertableSubs.mapValues { case (account, sub) => PaymentFailureAlerter.alertAvailableFor(account, sub, paymentMethodGetter) }
-      val futureList = results.map { case (product, res) => res.map(res1 => (product, res1)) }
+      val sortedResults = results.toList.sortWith { case ((product1, result1), (product2, result2)) => product1.name < product2.name }
+      val futureList = sortedResults.map { case (product, res) => res.map(res1 => (product, res1)) }
       Future.sequence(futureList).map(x => x.collect { case (product, true) => product.name })
     }
 
@@ -62,8 +63,7 @@ class AttributesMaker extends LoggingWithLogstashFields{
 
     val maybeMembershipAndAccount = membershipSub flatMap { sub: Subscription[AnyPlan] => subsWithAccounts.find(_.subscription.contains(sub)) }
 
-    //TODO JUST FOR NOW RETURN IT AS A STRING CONCAT
-    val maybeAlertAvailable = alertsAvailable.map(x=> if (x.isEmpty) None else Some(x.mkString(", ")))
+    val maybeAlertAvailable = alertsAvailable.map(_.headOption)
 
     maybeAlertAvailable map { maybeAlert =>
       def customFields(identityId: String, alertAvailableFor: String) = List(LogFieldString("identity_id", identityId), LogFieldString("alert_available_for", alertAvailableFor))
--- a/membership-attribute-service/app/services/AttributesMaker.scala
+++ b/membership-attribute-service/app/services/AttributesMaker.scala
@@ -53,7 +53,7 @@ class AttributesMaker extends LoggingWithLogstashFields{
 
       val justAlertableProducts = firstSubPerProduct.filterKeys(product => alertableProducts.contains(product))
 
-      justAlertableProducts.map { case (product, (account, subscription)) => ProductData(product, account, subscription) }.toList.sortWith(_.product.name > _.product.name)
+      justAlertableProducts.map { case (product, (account, subscription)) => ProductData(product, account, subscription) }.toList.sortWith(_.product.name < _.product.name)
     }
 
     def findFirstAlert(productData: List[ProductData]): Future[Option[String]] = productData match {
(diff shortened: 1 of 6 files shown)


membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -52,6 +51,14 @@ object PaymentFailureAlerter extends LoggingWithLogstashFields {
         case None => Future.successful(None)
       }
 
+      def getProductDescription(subscription: Subscription[SubscriptionPlan.AnyPlan]) = if (subscription.asMembership.isDefined) {
+        s"${subscription.plan.productName} membership"
+      } else if (subscription.asContribution.isDefined) {
+        "contribution"
+      } else { //TODO SEE WHAT GENERIC TEXT WE SHOULD RETURN HERE
```
pvighi: Do we need to exclude the alertText in the /me/mma- endpoints for the products we do not include in "alertAvailableFor" in /me ?

membership-attribute-service/test/services/PaymentFailureAlerterTest.scala
```diff
@@ -26,19 +26,19 @@ class PaymentFailureAlerterTest(implicit ee: ExecutionEnv)  extends Specificatio
...
       "return none if one of the zuora calls returns a left" in {
-        val result: Future[Option[String]] = PaymentFailureAlerter.membershipAlertText(accountSummaryWithBalance, membership, paymentMethodLeftResponse)
+        val result: Future[Option[String]] = PaymentFailureAlerter.alertText(accountSummaryWithBalance, membership, paymentMethodLeftResponse)
 
         result must be_==(None).await
       }
 
       "return a message for a member who is in payment failure" in {
-        val result: Future[Option[String]] = PaymentFailureAlerter.membershipAlertText(accountSummaryWithBalance, membership, paymentMethodResponseRecentFailure)
+        val result: Future[Option[String]] = PaymentFailureAlerter.alertText(accountSummaryWithBalance, membership, paymentMethodResponseRecentFailure)
```
lmath: Perhaps add a test that it will return a message for a contributor too?

membership-attribute-service/app/services/PaymentFailureAlerter.scala
lmath: I don't think so. We deliberately released the change for the the alertText for /me/mma-membership first since the info is still useful even if they didn't get to the manage page via the banner. Also, we need to make a change for each tab à la https://github.com/guardian/frontend/compare/aw-digipack-warn to see the alertText anyway.

membership-attribute-service/test/services/PaymentFailureAlerterTest.scala
```diff
@@ -77,19 +87,19 @@ class PaymentFailureAlerterTest(implicit ee: ExecutionEnv)  extends Specificatio
...
 
-      "return true for for a member with a failed payment and an invoice in the last 27 days" in {
+      "return true for a member with a failed payment and an invoice in the last 27 days" in {
         val result = PaymentFailureAlerter.alertAvailableFor(accountObjectWithBalance, membership, paymentMethodResponseRecentFailure)
 
         result must be_==(true).await
       }
 
-      "return true for for a non-membership sub with a failed payment in the last 27 days too" in {
-        val result = PaymentFailureAlerter.alertAvailableFor(accountObjectWithBalance, digipack, paymentMethodResponseRecentFailure)
```
pvighi: This test will no longer pass because it now doesn't true for subs that are not contributions or memberships

membership-attribute-service/test/services/PaymentFailureAlerterTest.scala
pvighi: done

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -74,12 +85,14 @@ object PaymentFailureAlerter extends LoggingWithLogstashFields {
       expectedAlertText.map { someText => shouldShowAlert.option (someText).flatten }
     }
   }
+  val alertableProducts = List(Product.Membership, Product.Contribution)
```
lmath: This notion is going to make the Digital Pack rollout so much easier. 😃

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -40,6 +43,27 @@ class AttributesMaker extends LoggingWithLogstashFields{
...
     val groupedSubs: Map[Option[Product], List[Subscription[AnyPlan]]] = subs.groupBy(getTopProduct)
 
+    val sortedProductData: List[ProductData] = {
+      val accountsWithNonEmptySubs = subsWithAccounts.collect { case AccountWithSubscription(account, Some(subscription)) => (account, subscription) }
+
+      val groupedByProduct = accountsWithNonEmptySubs.groupBy { case (account, subscription) => getTopProduct(subscription) }
+
+      val firstSubPerProduct = groupedByProduct.collect { case (Some(k), sub :: _) => (k, sub) }
+
+      firstSubPerProduct.map { case (product, (account, subscription)) => ProductData(product, account, subscription) }.toList.sortWith(_.product.name < _.product.name)
```
johnduffell: may be more readable to put the case statements on each of these collect/group onto a separate line
```
... {
 case ....
}
```

johnduffell:
👍 great!
general comments is there are a few long lines in there making it a little harder to follow, also there are a lot of Future.successful which it would be nice to not have to use Future where the logic didn't need it (for testability it makes it easier).  If we could have structured the code so the future stuff was outside, it might have helped that a bit.  But none of that shoudl stop it shipping!

membership-attribute-service/app/services/AttributesMaker.scala
pvighi: Ok I'll do that, but I think it is especially hard to read here because Github restricts the width of the code to make sure it looks ok in computers from 1994

johnduffell: 🤣

PR #324, opened 2018-05-22
Scrub personally identifiable information before logging to Sentry

rupertbates:
### Why do we need this? 
This PR makes sure that we don't send any personally identifiable information to Sentry while still allowing us to log it elsewhere for diagnostic purposes

[This PR](https://github.com/guardian/support-frontend/pull/494) has more discussion of the approach taken


lmath:
https://github.com/guardian/members-data-api/blob/master/membership-attribute-service/app/loghandling/LoggingWithLogstashFields.scala#L22

This error logger isn't being used, but since it's not being scrubbed might as well remove it to remove the option?

PR #332, opened 2018-06-15
FIX: direct OPTIONS hit (without CORS pre-flight headers) would execute POST behaviour

twrichards:
removed `OPTIONS` routes from `routes` file as were unnecessary and could be called directly which would in-fact perform POST behaviour. 

Also made the CORSFilter build its list of POST paths dynamically from the generated `Routes.scala`

### DIFF
--- a/membership-attribute-service/app/wiring/AppLoader.scala
+++ b/membership-attribute-service/app/wiring/AppLoader.scala
@@ -57,13 +57,7 @@ class MyComponents(context: Context)
     "/user-attributes/me/mma-membership",
     "/user-attributes/me/mma-paper")
 
-  val mmaUpdatePaths = Seq(
-    "/user-attributes/me/membership-update-card",
-    "/user-attributes/me/digitalpack-update-card",
-    "/user-attributes/me/paper-update-card",
-    "/user-attributes/me/contribution-update-card",
-    "/user-attributes/me/cancel-regular-contribution",
-    "/user-attributes/me/contribution-update-amount")
+  val postPaths = router.documentation.filter(triple => triple._1.contains("POST")).map(triple => triple._2)
 
   override lazy val httpFilters: Seq[EssentialFilter] = Seq(
     new CheckCacheHeadersFilter(),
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/wiring/AppLoader.scala
```diff
@@ -57,21 +57,15 @@ class MyComponents(context: Context)
...
     "/user-attributes/me/mma-paper")
 
-  val mmaUpdatePaths = Seq(
-    "/user-attributes/me/membership-update-card",
-    "/user-attributes/me/digitalpack-update-card",
-    "/user-attributes/me/paper-update-card",
-    "/user-attributes/me/contribution-update-card",
-    "/user-attributes/me/cancel-regular-contribution",
-    "/user-attributes/me/contribution-update-amount")
+  val postPaths = router.documentation.filter(triple => triple._1.contains("POST")).map(triple => triple._2)
```
pvighi: we try to avoid using _1 and _2 and use case (some,somethingElse) instead to make the code more readable. In this case you can probably also replace the map and filter with a collect.
maybe you can check if something like this works:
```scala
val postPaths = router.documentation.collect { case ("POST", path, _) => path }
```
in any case I'm not sure if using the generated Routes like this is something we should avoid or not, maybe someone who knows more about play can give their opinion as well

membership-attribute-service/conf/routes
```diff
@@ -9,26 +9,16 @@ GET        /user-attributes/me/mma-membership               controllers.AccountC
 GET        /user-attributes/me/mma-monthlycontribution      controllers.AccountController.monthlyContributionDetails
 GET        /user-attributes/me/mma-paper                    controllers.AccountController.paperDetails
 
+# IMPORTANT: for update endpoints (e.g. POST) to correctly support OPTIONS for CORS they must be added to AppLoader.scala
```
pvighi: It seems like this comment only makes sense if we leave the manually maintained list in AppLoader.scala

twrichards:
tested in `CODE`

PR #336, opened 2018-07-13
Add delete override to dynamo attributes for staff ad-free trial

JustinPinner:
### Why do we need this?
We want to reopen the ad-free trial to staff, but without fully committing everybody by default. Yet. We used to use the membership attributes Dynamo tables to do this, but changes made there for GDPR reasons meant that anyone who had previously signed up for the trial but not signed in for a couple of weeks had their data removed, and unless they had a Zuora account, would be removed every time they sign in. 

So we need some mechanism through which to reinstate the trial, without having a strong dependency on Zuora, at least at this stage.

### The changes
This change adds support for a new field in the Dynamo data: `KeepFreshForStaffAdFree`. This is a **temporary** situation until we're through the internal testing phase and ready to push into automatic adoption for staff accounts, which may or may not be distinctly separate from the roll out to paying subscribers.

To summarise: If a Dynamo record has the `KeepFreshForStaffAdFree` it will have its `TTL` updated on every access to ensure it doesn't fall foul of automatic removal operations. This will work regardless of whether Zuora finds any data for the user or not.

The _only_ way to set the keep-fresh data is by manually adding it to an existing record, or manually creating a new record in Dynamo. This is no better or worse than when we ran the trial last time.

I considered adding a completely separate table to do this, but the overhead in terms of wiring it into AWS via cloudformation, and then adding a third layer of assessment into the already complex code that manages the current Zuora<->Dynamo relationship made that undesirable for a relatively short-lived feature (this is part of an OKR now, so we are approaching a time when it can be properly delivered). Also, if we later revert the changes in this PR, the data in the Dynamo table will gradually be cleared out by the automatic removal process, and we won't have to do anything with it.

In terms of numbers, we're expecting participation in the mid-hundreds range at best. It'd be nice to have more, but until the comms go out, that's unlikely to happen imminently.

### DIFF
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -61,7 +61,7 @@ case class DynamoAttributes(
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
   MembershipNumber: Option[String],
   AdFree: Option[Boolean],
-  KeepFreshForStaffAdFree: Option[Boolean],
+  KeepFreshForStaffAdFree: Option[Boolean] = None,
   TTLTimestamp: Long) {
   lazy val isFriendTier = Tier.exists(_.equalsIgnoreCase("friend"))
   lazy val isSupporterTier = Tier.exists(_.equalsIgnoreCase("supporter"))
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -27,6 +27,7 @@ case class Attributes(
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
   MembershipNumber: Option[String] = None,
   AdFree: Option[Boolean] = None,
+  KeepFresh: Option[Boolean] = None,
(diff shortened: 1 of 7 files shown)


membership-attribute-service/app/models/Attributes.scala
```diff
@@ -60,6 +61,7 @@ case class DynamoAttributes(
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
   MembershipNumber: Option[String],
   AdFree: Option[Boolean],
+  KeepFreshForStaffAdFree: Option[Boolean],
```
jacobwinch: It might be cleaner to default this to None too?

jacobwinch:
Looks sensible to me - we probably need a Trello card to remind us to tidy this up if there isn't one already 👍

JustinPinner:
Thanks @jacobwinch. I've added a comment to that effect to [this Trello card](https://trello.com/c/rQk60PpL/1591-give-ad-free-entitlement-to-active-digital-pack-subscriptions)

PR #341, opened 2018-09-14
Improve CORS setup

paulbrown1982:
### Why do we need this? 

The CORS setup was configured in a confusing way since the Play 2.6 upgrade changed how the configuration was wired in. Even though we had multiple URL paths configured, it was only finding the shortest, and so only evaluating to the standard CORS list, ignoring the mma-specific one.

### The changes 
 This change simplifies and standardises the CORS settings for all endpoints on the API.

cc @jacobwinch

### DIFF
--- a/membership-attribute-service/app/wiring/AppLoader.scala
+++ b/membership-attribute-service/app/wiring/AppLoader.scala
@@ -45,18 +45,6 @@ class MyComponents(context: Context)
     new AccountController(commonActions, controllerComponents)
   )
 
-  val regularCorsPaths = Seq(
-    "/user-attributes/me/membership",
-    "/user-attributes/me/features",
-    "/user-attributes/me"
-  )
-
-  val mmaPaths = Seq(
-    "/user-attributes/me/mma-digitalpack",
-    "/user-attributes/me/mma-monthlycontribution",
-    "/user-attributes/me/mma-membership",
-    "/user-attributes/me/mma-paper")
-
   val postPaths: List[String] = router.documentation.collect { case ("POST", path, _) => path }
 
   override lazy val httpFilters: Seq[EssentialFilter] = Seq(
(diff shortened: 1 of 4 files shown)


membership-attribute-service/conf/DEV.public.conf
```diff
@@ -11,28 +11,18 @@ touchpoint.backend.test=UAT
 
 stage=DEV
 
+play.filters.cors.supportsCredentials = true
```
paulbrown1982: This might be defaulted to true, but to save confusion, I thought it should be hardcoded anyway.

membership-attribute-service/conf/PROD.public.conf
```diff
@@ -11,20 +11,13 @@ logger.application=INFO
 touchpoint.backend.default=PROD
 touchpoint.backend.test=UAT
 
+play.filters.cors.supportsCredentials = true
```
paulbrown1982: This might be defaulted to true, but to save confusion, I thought it should be hardcoded anyway.

membership-attribute-service/app/wiring/AppLoader.scala
```diff
@@ -45,27 +45,14 @@ class MyComponents(context: Context)
...
   val postPaths: List[String] = router.documentation.collect { case ("POST", path, _) => path }
 
   override lazy val httpFilters: Seq[EssentialFilter] = Seq(
     new CheckCacheHeadersFilter(),
     csrfFilter,
     new AddEC2InstanceHeader(wsClient),
     new AddGuIdentityHeaders(),
-    CORSFilter(corsConfig = Config.mmaCorsConfig, pathPrefixes = mmaPaths),
     CORSFilter(corsConfig = Config.mmaUpdateCorsConfig, pathPrefixes = postPaths),
-    CORSFilter(corsConfig = Config.corsConfig, pathPrefixes = regularCorsPaths)
```
paulbrown1982: Previously, this was ONLY matching on "/user-attributes/me" anyway, so all the other paths were redundant.

paulbrown1982:
Tested in CODE and the XHR/CORS request works fine from profile, www, and subscribe subdomains.

paulbrown1982:
Tested in CODE and the XHR/CORS request works fine from profile, m, and subscribe subdomains.

PR #345, opened 2018-11-07
Adding a warning for whomever may decide to increase the desired number of instances

paulbrown1982:
Adding a warning for whomever may decide to increase the desired number of instances:

Members Data API limits the number of concurrent calls it makes via its Zuora Service, so as to protect against breaching [Zuora's Concurrent Request Limits](https://knowledgecenter.zuora.com/BB_Introducing_Z_Business/Policies/Concurrent_Request_Limits)

The value is configured in DynamoDB and is currently set at 15 requests per instance. So 15 x 6 = 90 which is too big anyway! - I'll test setting this to 6 tomorrow.

cc @jacobwinch @johnduffell @lmath

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -23,7 +23,7 @@ Parameters:
     Type: Number
     Default: 12
   MinInstances:
-    Description: Minimum number of instances. NOTE: there is a Zuora concurrent call threshold configured in MembershipFeatureToggles-[STAGE] DynamoDB table which will need adjusting if you up this value.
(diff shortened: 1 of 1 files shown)


cloudformation/membership-attribute-service.yaml
```diff
@@ -178,6 +178,8 @@ Resources:
         Ref: LaunchConfig
       MinSize:
         Ref: MinInstances
+      DesiredCapacity:
```
jacobwinch: I would not explicitly set desired in this template, see the note at: https://riffraff.gutools.co.uk/docs/magenta-lib/types#cloudformation

johnduffell:
👍 thanks!

PR #348, opened 2018-11-30
added new 'update direct debit' endpoints

twrichards:


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -103,6 +100,73 @@ class AccountController(commonActions: CommonActions, override val controllerCom
     }
   }
 
+  private def updateDD[P <: SubscriptionPlan.AnyPlan : SubPlanReads] = BackendFromCookieAction.async { implicit request =>
+    // TODO - refactor to use the Zuora-only based lookup, like in AttributeController.pickAttributes - https://trello.com/c/RlESb8jG
```
jacobwinch: I think it's probably worth taking this approach for the initial implementation rather than adding this as a todo

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -103,6 +100,73 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+    val tp = request.touchpoint
+    val maybeUserId = authenticationService.userId
+    logger.info(s"Attempting to update direct debit for $maybeUserId")
+    (for {
+      user <- EitherT(Future.successful(maybeUserId \/> "no identity cookie for user"))
+      directDebitDetails <- EitherT(Future.successful(updateForm.bindFromRequest().value \/> "no direct debit details submitted with request"))
+      (bankAccountName, bankAccountNumber, bankSortCode) = directDebitDetails
+      sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"no SF user $user")))
+      subscription <- EitherT(tp.subService.current[P](sfUser).map(_.headOption).map (_ \/> s"no current subscriptions for the sfUser $sfUser"))
+      account <- EitherT(annotateFailableFuture(tp.zuoraService.getAccount(subscription.accountId), s"get account with id ${subscription.accountId}"))
```
jacobwinch: I think it might be possible to get all of the data we need here (about the account and billing contact) from a single call: https://www.zuora.com/developer/api-reference/#operation/GET_AccountSummary

jacobwinch: Ah, in fact I think you might already be using it under the hood: https://github.com/guardian/membership-common/blob/401f38f5c5de418429b03e0e94982e17a547db2a/src/main/scala/com/gu/zuora/rest/ZuoraRestService.scala#L309-L311

jacobwinch: So hopefully we don't need the call to `tp.zuoraService.getContact`?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -232,8 +296,12 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   def digitalPackUpdateCard = updateCard[SubscriptionPlan.Digipack]
   def paperUpdateCard = updateCard[SubscriptionPlan.PaperPlan]
   def contributionUpdateCard = updateCard[SubscriptionPlan.Contributor]
+
   def contributionUpdateAmount = updateContributionAmount[SubscriptionPlan.Contributor]
 
+  def contributionUpdateDD = updateDD[SubscriptionPlan.Contributor]
```
jacobwinch: very minor point, but I would avoid using the DD shortening in names

membership-attribute-service/conf/routes
```diff
@@ -13,6 +13,9 @@ POST       /user-attributes/me/digitalpack-update-card      controllers.AccountC
 POST       /user-attributes/me/paper-update-card            controllers.AccountController.paperUpdateCard
 POST       /user-attributes/me/contribution-update-card     controllers.AccountController.contributionUpdateCard
 
+POST       /user-attributes/me/contribution-update-dd       controllers.AccountController.contributionUpdateDD
```
jacobwinch: I know we're following an existing convention here, but is it necessary to have two separate endpoints for this?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -1,40 +1,37 @@
 package controllers
```
jacobwinch: This controller is getting pretty large - we could consider moving the functions related to payment method update  into their own controller?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -103,6 +100,73 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+        paymentMethod = bankTransferPaymentMethod,
+        paymentGateway = account.paymentGateway.get, // this will need to change to use this endpoint for 'payment method' SWITCH
+        billtoContact = billToContact,
+        invoiceTemplateOverride = None
+      )
+      _ <- EitherT(annotateFailableFuture(tp.zuoraService.createPaymentMethod(createPaymentMethod), "create direct debit payment method"))
+      freshDefaultPaymentMethodOption <- EitherT(annotateFailableFuture(tp.paymentService.getPaymentMethod(subscription.accountId), "get fresh default payment method"))
+    } yield freshDefaultPaymentMethodOption match {
+      case Some(dd: GoCardless)
+        if bankAccountName == dd.accountName &&
```
jacobwinch: We could pull this comparison into a def (or a val) to make it easier to follow this code

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -103,6 +100,73 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+          dd.accountNumber.length>3 && bankAccountNumber.endsWith(dd.accountNumber.substring(dd.accountNumber.length - 3)) &&
+          bankSortCode == dd.sortCode =>
+        logger.info(s"Successfully updated direct debit for identity user: $user")
+        Ok(Json.obj(
```
jacobwinch: Do we need to return this? Aren't we just echoing back the data that was posted to us?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -103,6 +100,73 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+      createPaymentMethod = CreatePaymentMethod(
+        accountId = subscription.accountId,
+        paymentMethod = bankTransferPaymentMethod,
+        paymentGateway = account.paymentGateway.get, // this will need to change to use this endpoint for 'payment method' SWITCH
+        billtoContact = billToContact,
+        invoiceTemplateOverride = None
+      )
+      _ <- EitherT(annotateFailableFuture(tp.zuoraService.createPaymentMethod(createPaymentMethod), "create direct debit payment method"))
+      freshDefaultPaymentMethodOption <- EitherT(annotateFailableFuture(tp.paymentService.getPaymentMethod(subscription.accountId), "get fresh default payment method"))
+    } yield freshDefaultPaymentMethodOption match {
```
jacobwinch: We could pull this block into a separate function, which would help with readability and allow us to unit test it.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -103,6 +100,73 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+      )
+      createPaymentMethod = CreatePaymentMethod(
+        accountId = subscription.accountId,
+        paymentMethod = bankTransferPaymentMethod,
+        paymentGateway = account.paymentGateway.get, // this will need to change to use this endpoint for 'payment method' SWITCH
+        billtoContact = billToContact,
+        invoiceTemplateOverride = None
+      )
+      _ <- EitherT(annotateFailableFuture(tp.zuoraService.createPaymentMethod(createPaymentMethod), "create direct debit payment method"))
+      freshDefaultPaymentMethodOption <- EitherT(annotateFailableFuture(tp.paymentService.getPaymentMethod(subscription.accountId), "get fresh default payment method"))
```
jacobwinch: I'm not keen on these really long for comprehensions as they are hard to follow and hard to test (although I appreciate there are already a few like this in the controller).

I think we could benefit from pulling the validation steps and the update steps into two separate functions.

membership-attribute-service/conf/routes
twrichards: how do you propose we ensure we target the right product?

it's worth mentioning; in the coming weeks I will be pairing with @jranks123 to have the endpoints accept a sub id. This work was intended to support concurrent recurring contributions, however we could use this idea (of requiring a sub id) to facilitate more generic endpoints

membership-attribute-service/app/controllers/AccountController.scala
twrichards: they're in fact slightly different given they're masked when they come back - and the equiv. details come back for card update so wanted to mirror for consistency

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -103,6 +100,73 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+          bankSortCode == dd.sortCode =>
+        logger.info(s"Successfully updated direct debit for identity user: $user")
+        Ok(Json.obj(
+          "accountName" -> dd.accountName,
+          "accountNumber" -> dd.accountNumber,
+          "sortCode" -> dd.sortCode
+        ))
+      case Some(_) =>
+        logger.error(s"New payment method for user $maybeUserId, does not match the posted Direct Debit details")
+        InternalServerError(s"Failed to update direct debit for user $maybeUserId")
```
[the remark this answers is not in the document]
twrichards: I've set them to empty for now.

membership-attribute-service/app/controllers/AccountController.scala
twrichards: done

membership-attribute-service/conf/routes
jacobwinch: Yes, I think we would need to have the client pass a sub (or account) id as part of the request. Let's pick this up separately as it requires some further thought (and probably a refactor of the existing endpoints) which shouldn't block this PR 👍

twrichards: as I say I'll be refactoring all the endpoints with @jranks123 next week anyway, so will hopefully be able to clean up these new endpoints, deprecate lots of the other routes and introduce a few new more generic routes

membership-attribute-service/app/controllers/AccountController.scala
twrichards: this TODO is copied from the credit card one - I could try this approach now but based on wider discussion it looks likely we'll need to refactor all methods like this in the near future

membership-attribute-service/app/controllers/AccountController.scala
twrichards: unfortunately the current case class for `BillToContact` that forms part of the `AccountSummary` which one gets back from the Zuora REST service, doesn't include the fields I need (first name, last name etc). It might well be the case that just need to update https://github.com/guardian/membership-common/blob/401f38f5c5de418429b03e0e94982e17a547db2a/src/main/scala/com/gu/zuora/rest/ZuoraRestService.scala#L191-L194 to add those fields as they might be present in the raw response from Zuora - but that would involve a change to https://github.com/guardian/membership-common which I've managed to avoid thus far.

membership-attribute-service/app/controllers/AccountController.scala
twrichards: i'm happy to do this now if you still think it's worth it though?

membership-attribute-service/app/controllers/AccountController.scala
twrichards: since pulling the yield block into its own def (as per one of your other comments) pulling this if statement into its own thing seems like overkill - each clause of the if is on its own line and it being in a separate val/def would be no different

twrichards: happy to change if you still think it would help

membership-attribute-service/app/controllers/AccountController.scala
twrichards: There isn't really any 'validation' to speak of, just building up all the details in order to perform the payment method update, if I were to split it, it would end up just being a function with loads of params, which is arguably what it is already. I did try it but it made it much bigger and less readable given all the things to pass around.

membership-attribute-service/app/controllers/AccountController.scala
jacobwinch: Cool, no need to change this on my account 👍

membership-attribute-service/app/controllers/AccountController.scala
jacobwinch: If that's all we need to do to save the call then I would go for it now. It's only a small membership-common change which shouldn't affect any other apps, so it should be relatively straightforward.

twrichards: On closer inspection, the `CreatePaymentMethod` further down the for comprehension requires a Contact of the soap variety, so I would need to rebuild the 6fields case class in order to reuse the current DD implementation from membership-common which re-raises the question of doing the DD from scratch with rest as new functionality in membership-common.

membership-attribute-service/app/controllers/AccountController.scala
twrichards: The refactor to use sub ID is in the current sprint for me and Jonny so this TODO will be removed before the year is out.

membership-attribute-service/app/controllers/AccountController.scala
pvighi: this might be a refactor for future PRs but I think we could break down this into more specific categories of functionality. 
Especially the big functions like cancelSubscription or updateDirectDebit which contain nested functions inside

twrichards: yeah discussed the same with Jacob and going to do that in a PR immediately following this one, to stop this one from being too large

PR #349, opened 2018-12-07
Upgrade to Scala 2.12

jacobwinch:
### Why do we need this? 

This project depends on a couple of libraries which have now dropped 2.11 support. This PR brings the project up to date, which allows us to pick up: https://github.com/guardian/membership-common/pull/585

This is required so that this app can continue to load the catalog after some planned changes.

### The changes 

* Upgrade to 2.12.6
* Upgrade to latest version of membership-common and identity-play-auth
* Fix tests with ambiguous implicits

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,7 +21,7 @@ def buildInfoSettings = Seq(
 val commonSettings = Seq(
   organization := "com.gu",
   version := appVersion,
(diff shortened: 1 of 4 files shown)


johnduffell:
great! 👍 🚀

Issue #335, opened 2018-07-09
"X-Gu-Membership-Test-User" header always false because using identityID rather than firstName


twrichards:
https://github.com/guardian/members-data-api/blob/a5f602b3046bbc2aee2ac49e029fe4bfe7606156/membership-attribute-service/app/filters/AddGuIdentityHeaders.scala#L28

twrichards:
@rtyley @johnduffell @jacobwinch @pvighi any ideas about what things use this value? NOTE: this is the response headers for members-data-api, the actual [WithBackendFromCookieAction](https://github.com/guardian/members-data-api/blob/a5f602b3046bbc2aee2ac49e029fe4bfe7606156/membership-attribute-service/app/actions/WithBackendFromCookieAction.scala) does it correctly so members-data-api does actually connect to the desired backend - perhaps hence why it hasn't been noticed

johnduffell:
I think that is just for logging, we used to (maybe still do) log it at fastly. But it has to be a header to be loggable there.

twrichards:
in which case its logging incorrect values, i.e. fastly logs won't think there are any test-users coming through. Shall I do a fix?

PR #350, opened 2018-12-12
fixed the 'X-Gu-Membership-Test-User' header as it was always returning false

twrichards:
Previously this was using the _identityID_ to check against test-user lib, when in fact the _first name_ is the thing to use.
Also refactored so the evaluation of test-user status is only defined in one place.

_This needed to be fixed before it can be used in `manage-frontend` to determine a flag to send to the new GoCardless validation endpoint in `payment-api`._ 

---

TESTED IN CODE
image

### DIFF
--- a/membership-attribute-service/app/filters/AddGuIdentityHeaders.scala
+++ b/membership-attribute-service/app/filters/AddGuIdentityHeaders.scala
@@ -20,11 +20,15 @@ class AddGuIdentityHeaders (implicit val mat: Materializer, ex: ExecutionContext
 
 object AddGuIdentityHeaders {
 
+  //Identity checks for test users by first name
+  def isTestUser(displayName: Option[String]) =
+    displayName.flatMap(_.split(' ').headOption).exists(Config.testUsernames.isValid)
+
   def headersFor(request: RequestHeader, result: Result) = (for {
     user <- IdentityAuthService.playAuthService.authenticatedUserFor(request)
   } yield result.withHeaders(
     "X-Gu-Identity-Id" -> user.id,
     "X-Gu-Identity-Credentials-Type" -> user.credentials.getClass.getSimpleName,
-    "X-Gu-Membership-Test-User" -> Config.testUsernames.isValid(user.id).toString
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/filters/AddGuIdentityHeaders.scala
```diff
@@ -20,11 +20,15 @@ class AddGuIdentityHeaders (implicit val mat: Materializer, ex: ExecutionContext
 
 object AddGuIdentityHeaders {
 
+  //Identity checks for test users by first name
+  def isTestUser(displayName: Option[String]) =
+    displayName.flatMap(_.split(' ').headOption).exists(Config.testUsernames.isValid)
```
rupertbates: I know you didn't add this but is there not a firstName field to avoid having to split the displayName? It feels slightly brittle.

twrichards: unfortunately not without doing another call to identity I believe. The `displayName` here is straight out of the cookie. I am double checking before I merge it.

twrichards: yep, the first name is one of the 'Private Fields' in the cookie (unlike `displayName` which is one of the 'Public Fields') 
image

johnduffell: at least it's only for us, won't break any real users, definitely not worth an additional rest call!
first name is correct based on the behaviour of most of the other subs estate, normally email and last name are ignored.

johnduffell:
great, thanks 👍 
fyi I think I checked once and that field was used in fastly for logging, I am not sure I found any code it was used in.  I can't remember why I was looking but I have a feeling it was broken then too, so good work fixing it up.

jacobwinch:
@johnduffell - I guess you were looking because of: https://github.com/guardian/members-data-api/issues/335

twrichards:
> @johnduffell - I guess you were looking because of: #335

I have no recollection of raising that issue - glad I've fixed it lol - I'll close the issue now 😄

PR #353, opened 2018-12-20
remove ad-free flag

lmath:
### Why do we need this? 
Previously, ad-free status was stored in members-data-api.

The ad-free cookie is set here:
https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/user-features.js#L83
(PR to remove stop checking for an ad-free flag in frontend to follow! )

Now, we think about ad-free status being determined by whether someone has a digital pack, so we can remove the code that maintains an ad-free status in members-data-api. It's better to rely on a value that we fetch from Zuora, because then it is up to date, and the members-data-api attributes table has a TTL.

### The changes 
* Stop keeping AdFree in the attributes table
* Remove the code that maintains ad-free

There are no more attributes that have an ad-free status
image

Part 1 of the ad-free code removal:
https://github.com/guardian/members-data-api/pull/351


membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -73,11 +74,12 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
     //return what we know from Dynamo if Zuora returns an error
     val fallbackIfZuoraFails: Future[(String, Option[Attributes])] = attributesFromDynamo map { maybeDynamoAttributes => ("Dynamo", maybeDynamoAttributes map DynamoAttributes.asAttributes)}
 
+    val what: Future[Disjunction[String, Future[Option[ZuoraAttributes]]]] = attributesFromZuora
```
jacobwinch: Is this used?

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -93,8 +103,7 @@ object DynamoAttributes {
     MembershipJoinDate = dynamoAttributes.MembershipJoinDate,
     DigitalSubscriptionExpiryDate = dynamoAttributes.DigitalSubscriptionExpiryDate,
     PaperSubscriptionExpiryDate = dynamoAttributes.PaperSubscriptionExpiryDate,
-    MembershipNumber = dynamoAttributes.MembershipNumber,
-    AdFree = dynamoAttributes.AdFree,
+    MembershipNumber = dynamoAttributes.MembershipNumber
```
jacobwinch: One for a future PR, but I wonder if we can drop this too 🤔

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -73,11 +74,12 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
...
     val attributesFromZuoraUnlessFallback: Future[(String, Option[Attributes])] = attributesFromZuora flatMap {
       case -\/(error) => fallbackIfZuoraFails
       case \/-(maybeZuoraAttributes) => maybeZuoraAttributes flatMap { zuoraAttributes: Option[ZuoraAttributes] =>
         attributesFromDynamo map { dynamoAttributes =>
-          val combinedAttributes = AttributesMaker.zuoraAttributesWithAddedDynamoFields(zuoraAttributes, dynamoAttributes)
+          val combinedAttributes = zuoraAttributes.map(ZuoraAttributes.asAttributes(_))
```
jacobwinch: I don't think combined attributes is an appropriate name for this anymore?

Also, if I've read it correctly I think we're always setting membership number to None here... is that intentional?

membership-attribute-service/app/services/AttributesFromZuora.scala
lmath: oooops

lmath: (nope)

membership-attribute-service/app/services/AttributesFromZuora.scala
lmath: Good points. 

I'm going to put the code back in to maintain membership number for now. It would make sense to deal with membership number in its own PR.

PR #356, opened 2018-12-31
Remove MembershipNumber

lmath:
### Why do we need this? 
To simplify the code behind `/user-attributes/me`. 
1. The callers of `/user-attributes/me` don't use it as a source for membership number 
2. As entries that had membership number dropped out of the cache, we couldn't return membership number anymore when we did previously. `/user-attributes/me` treats Zuora as the source of truth. 

Being able to remove membership number also allows us to remove the notion of values that are only available via the cache (not Zuora) now that the AdFree fields https://github.com/guardian/members-data-api/pull/353 https://github.com/guardian/members-data-api/pull/351 are removed.

### The changes 
- Removes membership number from the Attributes and DynamoAttributes model
- I'm going to remove entries in the cache that have MembershipNumber. 

[trello card](https://trello.com/c/Qripsnpx/2100-christmas-tech-debt-challenge-remove-ad-free-trial-code-from-the-members-data-api)

```
 aws dynamodb scan --table-name SupporterAttributesFallback-PROD --select COUNT --filter-expression "attribute_exists (MembershipNumber)" --profile membership --region eu-west-1
{
    "Count": 70979,
    "ScannedCount": 229024,
    "ConsumedCapacity": null
}
```
There is the issue of there being 71K entries in the cache with a membership number. I previously removed the wallet field https://github.com/guardian/members-data-api/pull/282 so I have a script already to remove all entries in the cache that have a wallet. I am planning to use this script to remove all the entries that have a membership number too, unless anyone thinks there's a better way to go about that?


membership-attribute-service/test/services/AttributesFromZuoraTest.scala
```diff
@@ -157,7 +157,7 @@ class AttributesFromZuoraTest(implicit ee: ExecutionEnv) extends Specification w
         val attributes: Future[(String, Option[Attributes])] = attributesFromZuora.getAttributes(testId, identityIdToAccountIds, subscriptionFromAccountId,  paymentMethodResponseNoFailures, mockDynamoAttributesService, referenceDate)
         attributes must be_==("Zuora", Some(contributorAttributes)).await
 
-        there was one(mockDynamoAttributesService).update(contributorAttributesWithTtl)
+        eventually(there was one(mockDynamoAttributesService).update(contributorAttributesWithTtl))
```
lmath: I'm undecided if using eventually matchers is the best thing to do in these tests. Any other ideas?

lmath:
As you mentioned @jacobwinch , it's not strictly required to remove the entries with MembershipNumber and we could have just let them expire. But since I already have a script from when I removed the wallet, I did indeed remove all the entries with a MembershipNumber so that it's really clear that the cache is not where to find that information. 

```
aws dynamodb scan --table-name SupporterAttributesFallback-PROD --select COUNT --filter-expression "attribute_exists (MembershipNumber)" --profile membership --region eu-west-1
{
    "Count": 0,
    "ScannedCount": 164672,
    "ConsumedCapacity": null
}
```

PR #357, opened 2019-01-03
new endpoint which returns all current subscription details (as JSON array)

twrichards:
… in the long term this will replace the product specific endpoints

The current product specific endpoints (e.g. `/user-attributes/me/mma-monthlycontribution`) have a `headOption` which always limits the result to a single subscription, however since Guest Checkout on `support-frontend` there are legitimate scenarios where there can be multiple recurring contributions - and hence we needed an endpoint to expose this.

Furthermore the plan for Q4 for `manage-frontend` is to consolidate all products into a single view and so this change has been made with that in mind - hence a single endpoint listing all current subs (all products).

_Paired with @johnduffell_ 

The /mma endpoint will now return a json array containing zero or more instances of the subscription response object, before it would have returned exactly one, otherwise an empty Json Object.

| Before | After |
| --- | --- |
| image | image | 
| `/user-attributes/me/mma-monthlycontribution`|`/user-attributes/me/mma` |

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -32,6 +32,7 @@ import scalaz.syntax.monad._
 import scalaz.syntax.std.option._
 import scalaz.syntax.traverse._
 import scalaz.{-\/, EitherT, ListT, OptionT, \/, \/-}
+import utils.{ListEither, OptionEither}
 
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -274,13 +275,18 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   }
 
 
-  def allSubscriptions(contactRepo: SimpleContactRepository, subService: SubscriptionService[Future])(maybeUserId: Option[String]): OptionT[OptionEither.FutureEither, List[Subscription[SubscriptionPlan.AnyPlan]]] = for {
+  def allSubscriptions(
+    contactRepo: SimpleContactRepository,
+    subService: SubscriptionService[Future]
+  )(
+    maybeUserId: Option[String]
+  ): OptionT[OptionEither.FutureEither, List[Subscription[SubscriptionPlan.AnyPlan]]] = for {
     user <- OptionEither.liftFutureEither(maybeUserId)
     contact <- OptionEither(contactRepo.get(user))
     subscriptions <- OptionEither.liftEitherOption(subService.current[SubscriptionPlan.AnyPlan](contact)) // TODO are we happy with an empty list in case of error ?!?!
   } yield subscriptions
 
-  private def allPaymentDetails = BackendFromCookieAction.async { implicit request =>
+  def allPaymentDetails = BackendFromCookieAction.async { implicit request =>
     implicit val tp = request.touchpoint
     def getPaymentMethod(id: PaymentMethodId) = tp.zuoraRestService.getPaymentMethod(id.get)
     val maybeUserId = authenticationService.userId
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -271,6 +273,40 @@ class AccountController(commonActions: CommonActions, override val controllerCom
     }
   }
 
+
+  def allSubscriptions(contactRepo: SimpleContactRepository, subService: SubscriptionService[Future])(maybeUserId: Option[String]): OptionT[OptionEither.FutureEither, List[Subscription[SubscriptionPlan.AnyPlan]]] = for {
```
johnduffell: could be worth splitting it down something along these lines
```suggestion
   def allSubscriptions(
     contactRepo: SimpleContactRepository,
     subService: SubscriptionService[Future]
   )(
     maybeUserId: Option[String]
   ): OptionT[OptionEither.FutureEither, List[Subscription[SubscriptionPlan.AnyPlan]]] =
   for {
```

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -271,6 +273,40 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+    subscriptions <- OptionEither.liftEitherOption(subService.current[SubscriptionPlan.AnyPlan](contact)) // TODO are we happy with an empty list in case of error ?!?!
+  } yield subscriptions
+
+  private def allPaymentDetails = BackendFromCookieAction.async { implicit request =>
+    implicit val tp = request.touchpoint
+    def getPaymentMethod(id: PaymentMethodId) = tp.zuoraRestService.getPaymentMethod(id.get)
+    val maybeUserId = authenticationService.userId
+
+    logger.info(s"Attempting to retrieve payment details for identity user: ${maybeUserId.mkString}")
+    (for {
```
johnduffell: for other reviewers - I think this line to 294 is the only part of `allPaymentDetails` that isn't just copy and paste from the soon to be deprecated version of the function

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -314,6 +350,9 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   def monthlyContributionDetails = paymentDetails[SubscriptionPlan.Contributor, Nothing]
   def digitalPackDetails = paymentDetails[SubscriptionPlan.Digipack, Nothing]
   def paperDetails = paymentDetails[SubscriptionPlan.PaperPlan, Nothing]
+
+  def allDetails = allPaymentDetails
```
johnduffell: probably don't need to have this line any more, as it's not adding anything

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -337,3 +376,27 @@ object OptionEither {
   }
 
 }
+
+object ListEither {
```
johnduffell: not sure if we should remove any of these if they are unused - also perhaps split into a separate file?

johnduffell:
looks good to me! 👍 however since I paired on this ideally someone else can look as well if possible!

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -62,25 +47,30 @@ object AccountDetails {
...
-            "interval" -> paymentDetails.plan.interval.mkString
-          )))
-      )
+        "tier" -> paymentDetails.plan.name,
+        "isPaidTier" -> (paymentDetails.plan.price.amount > 0f)
+      ) ++
+        regNumber.fold(Json.obj())({reg => Json.obj("regNumber" -> reg)}) ++
+        Json.obj(
+          "joinDate" -> paymentDetails.startDate,
+          "optIn" -> !paymentDetails.pendingCancellation,
```
paulbrown1982: Do we need this confusing field if nothing in the manage. site uses it?

twrichards: there appear to be a few _possible_ usages https://github.com/search?q=org%3Aguardian+optIn&type=Code

however, when the deprecated endpoints come to be removed (following everything moving to `manage-frontend`) it will be wise to do a serious spring-clean of this whole serialiser (once theres just one thing relying on it) as there are MANY improvements that could be made (removing things, renaming things, restructuring etc.)

johnduffell: agree, we should spring clean, especially if the endpoint is being moved to a different URL.  I wonder now, based on our chat last time, whether we should totally change to using a Wire model with a suitable hierarchy, rather than having sucha complicated serialiser (that additionaliy loses the account->subscription one to many relationship.)

PR #358, opened 2019-01-03
New 'write' endpoints which target specific subscription

twrichards:
Now that we have a new endpoint for returning multiple subs (**_https://github.com/guardian/members-data-api/pull/357_**) - moving away from the `_.headOption` concept - all the write endpoints need to be able to target specific subs.

This PR refactors all the existing 'write' action functions...

- Update Card
- Update Direct Debit
- Cancel
- Update Contribution Amount

... to take an `Option` of subscription name (aka _A-S number_) which is then passed (**provided as a path parameter**) to a new re-usable function `subscriptionSelector` which if...

- `Some` - filters the list of subs for the authenticated user based on the subscription name
- `None` - uses the `headOption` as before

**NOTE** the old endpoints have been retained but marked as `@Deprecated` and pass a `None` to the newly refactored functions which now require the `Option`

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -42,9 +42,9 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   implicit val executionContext: ExecutionContext= controllerComponents.executionContext
   lazy val authenticationService: AuthenticationService = IdentityAuthService
 
-  def subscriptionSelector[P <: SubscriptionPlan.AnyPlan](subscriptionNameOption: Option[memsub.Subscription.Name])(subscriptions: List[Subscription[P]]): Option[Subscription[P]] = subscriptionNameOption match {
-    case Some(subName) => subscriptions.find(_.name == subName)
-    case None => subscriptions.headOption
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -41,7 +42,12 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   implicit val executionContext: ExecutionContext= controllerComponents.executionContext
   lazy val authenticationService: AuthenticationService = IdentityAuthService
 
-   def cancelSubscription [P <: SubscriptionPlan.AnyPlan : SubPlanReads] = BackendFromCookieAction.async { implicit request =>
+  def subscriptionSelector[P <: SubscriptionPlan.AnyPlan](subscriptionNameOption: Option[memsub.Subscription.Name], messageSuffix: String)(subscriptions: List[Subscription[P]]): String \/ Subscription[P] = subscriptionNameOption match {
+    case Some(subName) => subscriptions.find(_.name == subName) \/> s"$subName was not a subscription for $messageSuffix"
+    case None => subscriptions.headOption \/> s"No current subscriptions for $messageSuffix"
```
paulbrown1982: Bye bye pervasive .headOption 🎉 I'm guessing once all endpoints have been changed in the client, we can remove this one too?

twrichards: correct (we would refactor out the `Option` of the parameter too)

PR #359, opened 2019-01-08
filterable 'all subscriptions' endpoint

twrichards:
Follows on from https://github.com/guardian/members-data-api/pull/357 and https://github.com/guardian/members-data-api/pull/358 and allows filtering on the all subs endpoint...

- Optional `subscriptionName` **path parameter** on the end of `/user-attributes/me/mma/` limits the list to zero or one subs which match the sub name
- Optional `productType` **query parameter** which will filter the list based on the type of `plan.product` (pattern matched against the provided string)

**Importantly** in both cases the filtering is done BEFORE the more expensive calls to Stripe etc

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -281,25 +281,29 @@ class AccountController(commonActions: CommonActions, override val controllerCom
     }
   }
 
+  sealed trait OptionalSubscriptionsFilter
+  case class FilterBySubName(subscriptionName: memsub.Subscription.Name) extends OptionalSubscriptionsFilter
+  case class FilterByProductNameStartsWith(productNameStartsWith: String) extends OptionalSubscriptionsFilter
+  case object NoFilter extends OptionalSubscriptionsFilter
 
   def allSubscriptions(
     contactRepo: SimpleContactRepository,
     subService: SubscriptionService[Future]
   )(
     maybeUserId: Option[String],
-    filter: Option[Either[memsub.Subscription.Name, String]]
+    filter: OptionalSubscriptionsFilter
   ): OptionT[OptionEither.FutureEither, List[Subscription[SubscriptionPlan.AnyPlan]]] = for {
     user <- OptionEither.liftFutureEither(maybeUserId)
     contact <- OptionEither(contactRepo.get(user))
     subscriptions <- OptionEither.liftEitherOption(subService.current[SubscriptionPlan.AnyPlan](contact)) // TODO are we happy with an empty list in case of error ?!?!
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -363,10 +386,11 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   @Deprecated def paperUpdateDirectDebit = updateDirectDebit[SubscriptionPlan.PaperPlan](None)
   def updateDirectDebitOnSpecificSub(subscriptionName: String) = updateDirectDebit[SubscriptionPlan.AnyPlan](Some(memsub.Subscription.Name(subscriptionName)))
 
-
   @Deprecated def membershipDetails = paymentDetails[SubscriptionPlan.PaidMember, SubscriptionPlan.FreeMember]
   @Deprecated def monthlyContributionDetails = paymentDetails[SubscriptionPlan.Contributor, Nothing]
   @Deprecated def digitalPackDetails = paymentDetails[SubscriptionPlan.Digipack, Nothing]
   @Deprecated def paperDetails = paymentDetails[SubscriptionPlan.PaperPlan, Nothing]
+  def allPaymentDetails(productType: Option[String]) = anyPaymentDetails(productType.fold[OptionalSubscriptionsFilter](NoFilter)(FilterByProductType))
```
johnduffell: minor comment:
slightly confusing in this fold on first sight is that NoFilter is a value (an object) but FilterByProductType is actually a function.  There's also sometimes an ambiguity caused where the type can't be inferred by the compiler, although clearly not in this case.  I used to strongly oppose fold on an option in favour of pattern match (for readability reasons), but now I'm just prefer to avoid.
```suggestion
  def allPaymentDetails(productType: Option[String]): OptionalSubscriptionsFilter = anyPaymentDetails(productType.fold[OptionalSubscriptionsFilter](NoFilter)(FilterByProductType.apply))
```

johnduffell:
👍

membership-attribute-service/app/controllers/AccountController.scala
mario-galic: Conceptually, both `NoFilter` and `FilterByProductType` are _constructors_ of `OptionalSubscriptionsFilter` algebraic data type, that is, entities that generate values of `OptionalSubscriptionsFilter`, not the values themselves.  `NoFilter` happens to be a constructor for zero arity, which takes no arguments,  but that is just fine. Scala reuses OO keyword terminology for functional concepts, but the semantics here are that of ADT, so `case object` really means ADT constructor here.

When folding an option over ADT we must provide type annotation because of https://stackoverflow.com/a/29587749/5205022

PR #346, opened 2018-11-19
add 'paperSubscriber' boolean to the 'contentAccess' section of the  /me endpoint

twrichards:
### Why do we need this? 
This facilitates on-boarding print to the payment update flow in manage-frontend - see https://github.com/guardian/manage-frontend/pull/157

### The changes 
add 'paperSubscriber' boolean to the 'contentAccess' section of the simple lightweight' /me endpoint

### trello card/screenshot/json/related PRs etc
image

PR #363, opened 2019-01-22
add 'guardianWeeklySubscriber' boolean to the 'contentAccess' section of the /me endpoint

twrichards:
### Why do we need this? 
Similar to https://github.com/guardian/members-data-api/pull/346 - this adds the last of the main products to the `contentAccess` section so this change is mainly for completeness.

### The changes 
add 'guardianWeeklySubscriber' boolean to the 'contentAccess' section of the /me endpoint

image

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -52,7 +52,7 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
             def customFields(supporterType: String): List[LogField] = List(LogFieldString("lookup-endpoint-description", endpointDescription), LogFieldString("supporter-type", supporterType), LogFieldString("data-source", fromWhere))
 
             attributes match {
-              case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _)) =>
+              case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _, _)) =>
                 logInfoWithCustomFields(s"$identityId is a $tier member - $endpointDescription - $attrs found via $fromWhere", customFields("member"))
                 onSuccessMember(attrs).withHeaders(
                   "X-Gu-Membership-Tier" -> tier,
@@ -64,6 +64,9 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
                 }
                 attrs.PaperSubscriptionExpiryDate.foreach {date =>
                   logInfoWithCustomFields(s"$identityId is a paper subscriber expiring $date", customFields("paper-subscriber"))
+                }
+                attrs.GuardianWeeklySubscriptionExpiryDate.foreach {date =>
+                  logInfoWithCustomFields(s"$identityId is a Guardian Weekly subscriber expiring $date", customFields("guardian-weekly-subscriber"))
                 }
                 attrs.RecurringContributionPaymentPlan.foreach { paymentPlan =>
                   logInfoWithCustomFields(s"$identityId is a regular $paymentPlan contributor", customFields("contributor"))
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -17,7 +17,8 @@ case class ContentAccess(
   paidMember: Boolean,
   recurringContributor: Boolean,
   digitalPack: Boolean,
-  paperSubscriber: Boolean
+  paperSubscriber: Boolean,
+  guardianWeeklySubscriber: Boolean
 )
 
 object ContentAccess {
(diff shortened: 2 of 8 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -65,6 +65,9 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
                 attrs.PaperSubscriptionExpiryDate.foreach {date =>
                   logInfoWithCustomFields(s"$identityId is a paper subscriber expiring $date", customFields("paper-subscriber"))
                 }
+                attrs.PaperSubscriptionExpiryDate.foreach {date =>
```
jacobwinch: Should this be attrs.GuardianWeeklySubscriptionExpiryDate?

membership-attribute-service/app/services/AttributesMaker.scala
```diff
@@ -79,6 +80,14 @@ class AttributesMaker extends LoggingWithLogstashFields{
       }
     }
 
+    def paidSubs = subs.filter{
```
jacobwinch: I think this can all be removed?

membership-attribute-service/app/controllers/AttributeController.scala
twrichards: good spot - fixed - thanks

membership-attribute-service/app/services/AttributesMaker.scala
twrichards: ahh i thought i'd unticked this from the commit - removed now - thanks

twrichards:
@paulbrown1982 
>  👍 I assume that the end date is only filled if the sub is cancelled or auto-renew = false?

No thats not the case, its just the greatest date if there are multiple Guardian Weeklys - however this is the same behaviour for the other top level date fields...
image
... I'm not sure they're serving any purpose, so we could remove them?!

paulbrown1982:
Auto-renewing subs only have an end date if they are cancelled, so if that helps you decide if the date is needed - you might need to talk to the Apps team about why the use the date in the iOS Daily Edition app.

twrichards:
@paulbrown1982 
> Auto-renewing subs only have an end date if they are cancelled, so if that helps you decide if the date is needed - you might need to talk to the Apps team about why the use the date in the iOS Daily Edition app.

Investigated this with @jorgeazevedo, at least on of these expiry date fields are currently being used in the Daily Edition (at least) so can't change that behaviour right now and best to keep this new one consistent with the existing.

Created https://github.com/guardian/members-data-api/issues/364 to track this and will do a separate PR to change this behaviour in bulk.

Merging this now, as is.

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -83,6 +89,7 @@ case class DynamoAttributes(
   MembershipJoinDate: Option[LocalDate] = None,
   DigitalSubscriptionExpiryDate: Option[LocalDate] = None,
   PaperSubscriptionExpiryDate: Option[LocalDate] = None,
+  GuardianWeeklySubscriptionExpiryDate: Option[LocalDate] = None,
```
johnduffell, line 92: we never actually write this field into dynamo! so cached responses to guardian weekly subscribers cause support asks to appear
https://github.com/guardian/members-data-api/blob/9970f2e01cd60d21abb0b77c7207b70785882746/membership-attribute-service/app/services/ScanamoAttributeService.scala#L58-L62

twrichards: wow good find. is someone going to raise a PR to address this?

paulbrown1982: There was actually a determination from the business a year-or-so ago that Guardian Weekly subscription would not count towards a support ask. It's worth reconfirming that before we wire up to the boolean which takes down asks. However, `guardianWeeklySubscriber` Content Access should still be stored in the Dynamo cache.

johnduffell: yes we are going to fix it.. but it shows an issue with the structure of the code that there's no hint that the write and read paths are in completely different places!  And no tests that picked that up.
I thought I would document it here in case someone is referring to this in future!  But I will at least add a comment in the code if nothing stronger.

johnduffell: @paulbrown1982 It's already wired up, it's been like that the whole time and it would work if the attributes don't come from the cache.
https://github.com/guardian/members-data-api/blob/26e95187feeb2431c705f75825d822453d2f4811/membership-attribute-service/app/models/Attributes.scala#L72
if it was decided not to include it at some point more recently, no one actually implemented it by the looks.
If someone confirms that then it's very easy to change the above line.  But we also need to let userhelp know.

paulbrown1982: OK if it's already in there it must be OK by now! Let's just get Dynamo working properly.

johnduffell: https://github.com/guardian/members-data-api/pull/482

PR #365, opened 2019-01-25
new 'showSupportMessaging' boolean on the /me response - facilitates centrally controlling support messaging on app & dotcom

twrichards:
### Why do we need this? 
Currently app and dotcom rely on the booleans in the `contentAccess` section, but this means any time we want to change when _support messaging_* should be shown, apps and dotcom have to change their logic (particularly long release/uptake cycle for the app) - this PR makes it centrally controlled.

_* = header, footer, epic, banner, thrasher etc_

### The changes 
Initially this just introduces the **`showSupportMessaging`** boolean which is `false` if the signed in user has any active products and `true` if they don't. However once the app & dotcom are updated to use it, we can make the calculation of the boolean more sophisticated, for example re-showing support messaging as users approach renewal, including one-off contributions (as we are planning to do in `members-data-api` anyway).

| No Products | Some Products |
| --- | --- |
| image | image |

### DIFF
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -57,6 +57,10 @@ case class Attributes(
     guardianWeeklySubscriber = isGuardianWeeklySubscriber
   )
 
+  // show support messaging (in app & on dotcom) if they do NOT have any active products
+  // TODO in future this could become more sophisticated (e.g. two weeks before their products expire)
+  lazy val showSupportMessaging = !(isPaidTier || isContributor || digitalSubscriberHasActivePlan || isPaperSubscriber || isGuardianWeeklySubscriber)
+
 }
 
 case class ZuoraAttributes(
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/models/Attributes.scala
```diff
@@ -57,6 +57,10 @@ case class Attributes(
     guardianWeeklySubscriber = isGuardianWeeklySubscriber
   )
 
+  // show support messaging (in app & on dotcom) if they do NOT have any active products
+  // TODO in future this could become more sophisticated (e.g. two weeks before their products expire)
+  lazy val showSupportMessaging = !(isPaidTier || isContributor || digitalSubscriberHasActivePlan || isPaperSubscriber || isGuardianWeeklySubscriber)
```
paulbrown1982: Maybe also add: isStaffTier

twrichards: `isPaidTier` already includes `isStaffTier `

twrichards: https://github.com/guardian/members-data-api/blob/8bf4a976781e64be59d98f97f2ed48ce3c9610dd/membership-attribute-service/app/models/Attributes.scala#L43

PR #366, opened 2019-01-28
introduced 'ContentSubscription' to facilitate combined Subscriptions tab in manage-frontend

twrichards:
### Why do we need this? 
In order to consolidate Paper, Weekly & Digipack (but not Membership or Contributions) into a **single Subscriptions tab** (see https://github.com/guardian/manage-frontend/pull/185) the recent `?productType=` functionality (introduced in https://github.com/guardian/members-data-api/pull/359) needed to support an additional value which would return this meaningful grouping of products, but since technically everything is a subscription, this sub-grouping was entitled **'ContentSubscription`**

### The changes 
A simple update to the pattern match, to return instances of `Paper`, `Weekly` and `ZDigipack` if 
`/user-attributes/me/mma?productType=ContentSubscription`

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -286,14 +286,17 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   case class FilterByProductType(productType: String) extends OptionalSubscriptionsFilter
   case object NoFilter extends OptionalSubscriptionsFilter
 
-  private def productIsInstanceOfProductType(product: Product, productType: String) = product match {
-    // this ordering prevents Weekly subs from coming back when Paper is requested (which is different from the type hierarchy where Weekly extends Paper)
-    case _: Product.Weekly => productType == "Weekly"
-    case _: Product.Paper => productType == "Paper"
-    case _: Product.Contribution => productType == "Contribution"
-    case _: Product.Membership => productType == "Membership"
-    case _: Product.ZDigipack => productType == "Digipack"
-    case _ => productType == product.name // fallback
+  private def productIsInstanceOfProductType(product: Product, requestedProductType: String) = {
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -286,14 +286,17 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
-  private def productIsInstanceOfProductType(product: Product, productType: String) = product match {
-    // this ordering prevents Weekly subs from coming back when Paper is requested (which is different from the type hierarchy where Weekly extends Paper)
-    case _: Product.Weekly => productType == "Weekly"
-    case _: Product.Paper => productType == "Paper"
-    case _: Product.Contribution => productType == "Contribution"
-    case _: Product.Membership => productType == "Membership"
-    case _: Product.ZDigipack => productType == "Digipack"
-    case _ => productType == product.name // fallback
+  private def productIsInstanceOfProductType(product: Product, productType: String) = {
+    val requestedProductTypeIsSoCalledSubscription: Boolean = productType == "SoCalledSubscription"
```
jacobwinch: I find this name a bit weird, but I understand the problem you're trying to avoid! Would guardianSubscription be clearer?

johnduffell: agree, SoCalled is not specific enough.
it's sometimes important to distinguish between zuora subscriptions (i.e. anything billed recurringly) and guardian subscriptions (anything you get regular content) and it can get confusing where they don't really map well.
From the outside of an API we provide, a subscription should really be a guardian one, however we probably already leak quite a bit of zuora terminology through the layers.  It doesn't hurt to be explicit, but without being confusing.

johnduffell:
see comment but apart from that sounds sensible 👍

membership-attribute-service/app/controllers/AccountController.scala
twrichards: renamed to `ContentSubscription`

PR #368, opened 2019-02-01
changed to use StageMappings for the MinInstances and MaxInstances

twrichards:
using parameters provided no distinction for CODE where we need far fewer instances, RiffRaff was reverting CODE to the `MinInstances` param default of `6` on every CODE deploy 😢 

**So now the defaults are in the `StageMappings`.**

In an emergency scale up scenario, changes can be made in the AWS console to scale up immediately, then a follow-up PR to change the values in `StageMappings` if necessary.

**This should save some wasted resources in CODE.**

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -17,15 +17,6 @@ Parameters:
     - m3.large
     - m3.xlarge
     ConstraintDescription: must be a valid EC2 instance type.
-  MaxInstances:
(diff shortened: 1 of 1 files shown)


twrichards:
CODE deploy looks to be successful and instance count is within the expected bounds - https://riffraff.gutools.co.uk/deployment/view/2b6379ed-f967-415f-8619-f596dc181562?verbose=1

johnduffell:
thanks for fixing this up tom! 👍 🏅

PR #370, opened 2019-02-18
Provide more information about bad payment states to clients

jacobwinch:
### Why do we need this?

_Note: This will not pass CI until https://github.com/guardian/membership-common/pull/589 is merged._

When a payment method is not correctly linked to a Zuora account _and_ the prerequisites for auto-cancellation are not met, it's possible for users to build up 'debts' whilst continuing to receive their subscription benefits. Currently many users in this bad state are unable to add a new payment method to fix the situation (which means that we cannot recover them easily). For other users, it _is_ possible to add a new payment method, but if they choose to do so this will inadvertently cause the payment for several invoices to be collected at once.

This PR adds a boolean to the (new) MMA responses, so that clients know if its unsafe to process a payment method update.

If a user doesn't have a correctly linked payment method (and it's safe to collect one), the response has also been updated to advise clients that a new payment method should be collected (and an appropriate Stripe key is included to allow them to do so).

### The changes
* Inform clients whether or not it is 'safe' for a payment method to be updated
* Inform clients that the user's payment method should be 'reset' if it is necessary and safe to do so
* Add unit tests
* Update existing code due to a change in the model (see related membership-common PR)

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -270,7 +270,16 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       stripeService = accountSummary.billToContact.country.map(RegionalStripeGateways.getGatewayForCountry).flatMap(tp.stripeServicesByPaymentGateway.get).getOrElse(tp.ukStripeService)
       alertText <- OptionEither.liftEitherOption(alertText(accountSummary, sub, getPaymentMethod))
       isAutoRenew = sub.autoRenew
-    } yield AccountDetails(contact.regNumber, accountSummary.billToContact.email, upToDatePaymentDetails, stripeService.publicKey, isAutoRenew, alertText).toJson).run.run.map {
+    } yield AccountDetails(
+      contact.regNumber,
+      accountSummary.billToContact.email,
+      upToDatePaymentDetails,
+      stripeService.publicKey,
+      accountHasMissedRecentPayments = false,
+      safeToUpdatePaymentMethod = true,
+      isAutoRenew = isAutoRenew,
+      alertText
+    ).toJson).run.run.map {
       case \/-(Some(result)) =>
         logger.info(s"Successfully retrieved payment details result for identity user: ${maybeUserId.mkString}")
         Ok(result)
@@ -339,7 +348,16 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       stripeService = accountSummary.billToContact.country.map(RegionalStripeGateways.getGatewayForCountry).flatMap(tp.stripeServicesByPaymentGateway.get).getOrElse(tp.ukStripeService)
       alertText <- ListEither.liftEitherList(alertText(accountSummary, subscription, getPaymentMethod))
       isAutoRenew = subscription.autoRenew
-    } yield AccountDetails(None, accountSummary.billToContact.email, upToDatePaymentDetails, stripeService.publicKey, isAutoRenew, alertText).toJson).run.run.map {
+    } yield AccountDetails(
+      regNumber = None,
+      email = accountSummary.billToContact.email,
+      paymentDetails = upToDatePaymentDetails,
+      stripePublicKey = stripeService.publicKey,
+      accountHasMissedRecentPayments = accountHasMissedRecentPayments(subscription.accountId, freeOrPaidSub.isRight, accountSummary.invoices, accountSummary.payments),
+      safeToUpdatePaymentMethod = safeToAllowPaymentUpdate(subscription.accountId, accountSummary.invoices),
+      isAutoRenew = isAutoRenew,
+      membershipAlertText = alertText
+    ).toJson).run.run.map {
       case \/-(subscriptionJSONs) =>
         logger.info(s"Successfully retrieved payment details result for identity user: ${maybeUserId.mkString}")
         Ok(Json.toJson(subscriptionJSONs))
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -353,7 +353,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       email = accountSummary.billToContact.email,
       paymentDetails = upToDatePaymentDetails,
       stripePublicKey = stripeService.publicKey,
-      accountHasMissedRecentPayments = accountHasMissedPayments(subscription.accountId, freeOrPaidSub.isRight, accountSummary.invoices, accountSummary.payments),
+      accountHasMissedRecentPayments = if (freeOrPaidSub.isRight) accountHasMissedPayments(subscription.accountId, accountSummary.invoices, accountSummary.payments) else false,
       safeToUpdatePaymentMethod = safeToAllowPaymentUpdate(subscription.accountId, accountSummary.invoices),
       isAutoRenew = isAutoRenew,
       membershipAlertText = alertText
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -353,7 +353,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       email = accountSummary.billToContact.email,
       paymentDetails = upToDatePaymentDetails,
       stripePublicKey = stripeService.publicKey,
-      accountHasMissedRecentPayments = accountHasMissedRecentPayments(subscription.accountId, freeOrPaidSub.isRight, accountSummary.invoices, accountSummary.payments),
+      accountHasMissedRecentPayments = accountHasMissedPayments(subscription.accountId, freeOrPaidSub.isRight, accountSummary.invoices, accountSummary.payments),
       safeToUpdatePaymentMethod = safeToAllowPaymentUpdate(subscription.accountId, accountSummary.invoices),
       isAutoRenew = isAutoRenew,
       membershipAlertText = alertText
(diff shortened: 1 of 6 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -270,7 +270,16 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       stripeService = accountSummary.billToContact.country.map(RegionalStripeGateways.getGatewayForCountry).flatMap(tp.stripeServicesByPaymentGateway.get).getOrElse(tp.ukStripeService)
       alertText <- OptionEither.liftEitherOption(alertText(accountSummary, sub, getPaymentMethod))
       isAutoRenew = sub.autoRenew
-    } yield AccountDetails(contact.regNumber, accountSummary.billToContact.email, upToDatePaymentDetails, stripeService.publicKey, isAutoRenew, alertText).toJson).run.run.map {
+    } yield AccountDetails(
+      contact.regNumber,
+      accountSummary.billToContact.email,
+      upToDatePaymentDetails,
+      stripeService.publicKey,
+      accountHasMissedRecentPayments = false,
```
jacobwinch: This is the old (deprecated) endpoint. I was planning on checking if anything is still calling this and ripping it out if not...

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -125,4 +127,28 @@ object PaymentFailureAlerter extends LoggingWithLogstashFields {
...
     alertAvailable.getOrElse(Future.successful(false))
   }
+
+  def mostRecentPayableInvoicesOlderThanOneMonth(recentInvoices: List[Invoice]): List[Invoice] = {
+    implicit def localDateOrdering: Ordering[DateTime] = Ordering.fromLessThan(_ isAfter _)
+    val nonZeroInvoicesOlderThanOneMonth = recentInvoices.filter(invoice => invoice.invoiceDate.isBefore(DateTime.now().minusMonths(1)) && invoice.amount > 0)
+    nonZeroInvoicesOlderThanOneMonth.sortBy(_.invoiceDate).take(2)
+  }
+
+  def accountHasMissedRecentPayments(accountId: AccountId, isPaidSub: Boolean, recentInvoices: List[Invoice], recentPayments: List[Payment]): Boolean = {
```
jacobwinch: This logic is all a bit fiddly (see unit tests). The logic would be simpler if we knew whether the account had autopay enabled or not, but I can't get that info without making an extra API call and this endpoint is already making too many requests (IMO).

membership-attribute-service/app/controllers/AccountController.scala
pvighi: minor: I would rather pass all the params by name instead of some by name and some by position.

membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -125,4 +127,28 @@ object PaymentFailureAlerter extends LoggingWithLogstashFields {
 
     alertAvailable.getOrElse(Future.successful(false))
   }
+
+  def mostRecentPayableInvoicesOlderThanOneMonth(recentInvoices: List[Invoice]): List[Invoice] = {
```
pvighi: why is "payable" part of this name ? is it because they are non zero invoices ?

membership-attribute-service/app/services/PaymentFailureAlerter.scala
pvighi: are the recentInvoices all from the same subscription? or does isPaidSub mean that there is at least one paid sub in the account ?

membership-attribute-service/test/services/PaymentFailureAlerterTest.scala
```diff
@@ -164,62 +169,172 @@ class PaymentFailureAlerterTest(implicit ee: ExecutionEnv)  extends Specificatio
...
+        PaymentFailureAlerter.accountHasMissedRecentPayments(accountId, isPaidSub = true, List(), List()) === false
+      }
+
+      "return false if the user has a single paid invoice" in {
+        PaymentFailureAlerter.accountHasMissedRecentPayments(accountId, isPaidSub = true, List(paidInvoice), List(paymentForPaidInvoice)) === false
+      }
+
+      "return false if the user has a recent unpaid invoice" in {
+        PaymentFailureAlerter.accountHasMissedRecentPayments(accountId, isPaidSub = true, List(recentUnpaidInvoice), List()) === false
+        PaymentFailureAlerter.accountHasMissedRecentPayments(accountId, isPaidSub = true, List(recentUnpaidInvoice), List(failedPaymentForRecentUnpaidInvoice)) === false
```
pvighi: do you test 2 cases here on purpose ?

membership-attribute-service/app/services/PaymentFailureAlerter.scala
pvighi: I find the name of this function slightly confusing because it never really checks that the payments or invoices are recent, that seems to be just a prerequisite to the input params. But it does check the opposite, that the invoices should be older than one month.

membership-attribute-service/app/services/PaymentFailureAlerter.scala
jacobwinch: Yes, unfortunately Zuora will still raise an invoice even though the amount is 0 (e.g. if there is 100% discount), so I wanted to make it explicit that such invoices will be ignored

membership-attribute-service/app/services/PaymentFailureAlerter.scala
jacobwinch: > are the recentInvoices all from the same subscription

This should be passed the invoices and payments sections of an account summary. I didn't want to pass a big object around unnecessarily, but I can amend this if you think it'd be clearer?

membership-attribute-service/test/services/PaymentFailureAlerterTest.scala
jacobwinch: Yes, I wanted to capture the case where we tried to take a payment and it failed vs the case where we didn't try to take a payment at all. Perhaps I should call it out as two separate scenarios?

membership-attribute-service/app/services/PaymentFailureAlerter.scala
jacobwinch: > I find the name of this function slightly confusing because it never really checks that the payments or invoices are recent, that seems to be just a prerequisite to the input params

Good point, I will drop 'recent' from the name 👍

membership-attribute-service/test/services/PaymentFailureAlerterTest.scala
pvighi: I think it would be slightly better to have one assertion per test case, but really I was just asking in case it was a copy and paste error or something. I'm fine with leaving it as it is

jacobwinch: > it would be slightly better to have one assertion per test case

Agreed, have fixed this.

membership-attribute-service/app/services/PaymentFailureAlerter.scala
pvighi: > > are the recentInvoices all from the same subscription
> 
> This should be passed the invoices and payments sections of an account summary. I didn't want to pass a big object around unnecessarily, but I can amend this if you think it'd be clearer?
 I was asking that because isPaidSub refers to a subscription but the invoices could potentially belong to multiple subs in the same account right ? So when isPaidSub = false does it mean all the subs in this account are not paid subs ?

jacobwinch: > I was asking that because isPaidSub refers to a subscription but the invoices could potentially belong to multiple subs in the same account right?

Correct.

> So when isPaidSub = false does it mean all the subs in this account are not paid subs ?

Not necessarily. Now that we support multiple subs in the same billing account I guess we should be open to the possibility that some will be free but others will be paid. The current implementation should omit any payment method details if clients ask about a free sub (based on https://github.com/guardian/membership-common/blob/aa309b43c45a3d6436eba1eb6c782f14512b04f0/src/main/scala/com/gu/memsub/services/PaymentService.scala#L27-L28). Similarly, using this boolean here allows us to omit details about a payment method reset in cases where clients are specifically asking about a free sub.

jacobwinch: Improved this slightly with https://github.com/guardian/members-data-api/pull/370/commits/c9add6659d515936823a357377ca8cd1a67cc8db.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -353,7 +353,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       email = accountSummary.billToContact.email,
       paymentDetails = upToDatePaymentDetails,
       stripePublicKey = stripeService.publicKey,
-      accountHasMissedRecentPayments = accountHasMissedPayments(subscription.accountId, freeOrPaidSub.isRight, accountSummary.invoices, accountSummary.payments),
+      accountHasMissedRecentPayments = if (freeOrPaidSub.isRight) accountHasMissedPayments(subscription.accountId, accountSummary.invoices, accountSummary.payments) else false,
```
pvighi: maybe you could write this as :
```scala
      accountHasMissedRecentPayments = freeOrPaidSub.isRight && accountHasMissedPayments(subscription.accountId, accountSummary.invoices, accountSummary.payments) 

```

jacobwinch: Yes that's much nicer, thanks for the suggestion

PR #371, opened 2019-02-19
expose new `currentPlans` and `futurePlans` arrays in the response of the detailed endpoint(s)

twrichards:
### Why do we need this? 
Currently `members-data-api` doesn't expose enough information about plans to correctly show price rise information nor which type of physical subscription people have - so this needs to be exposed for the likes of `manage-frontend` to display things correctly (see https://github.com/guardian/manage-frontend/pull/196).

### The changes 
Expose new `currentPlans` and `futurePlans` arrays in the response of the detailed endpoint(s)...

- `currentPlans` is all the plans which `start` today or in the past and end after today (where applicable)
- `futurePlans` is all the plans which `start` after today

... each is an array of objects representing the plans, including notable fields like...

- `name` which is only present if provides extra information, for example the delivery day(s), if a +Digipack or if '6 for 6' Guardian Weekly
 _NOTE this is different from the `name` field on the old `plan` entry which should've been labelled `productName` (but is redundant anyway)_
- `shouldBeVisible`- which is `true` if the acceptanceDate is in the future (e.g. like with 6for6 Guardian Weekly), else it is `true` if the rate plan starts in the next 30 days (just like https://github.com/guardian/subscriptions-frontend/pull/1246) - this field is then used to determine whether to display the price rise in `manage-frontend` (see https://github.com/guardian/manage-frontend/pull/196)

https://trello.com/c/nBLEUyKI
image
image

####  Examples of where `name` is not null...
image
image
image

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -69,8 +69,6 @@ object AccountDetails {
       })
 
       def jsonifyPlan(plan: SubscriptionPlan.AnyPlan) = Json.obj(
-        "productRatePlanId" -> plan.productRatePlanId.get, // consider exposing hash of this to avoid exposing internal IDs
-        "productName" -> plan.productName,
         "name" -> plan.name,
         "start" -> plan.start,
         // if the customer acceptance date is future dated (e.g. 6for6) then always display, otherwise only show if starting less than 30 days from today
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -1,6 +1,6 @@
 package models
 import com.gu.memsub.subsv2.{Subscription, SubscriptionPlan}
-import com.gu.memsub.{GoCardless, PayPalMethod, PaymentCard}
+import com.gu.memsub.{GoCardless, PayPalMethod, PaymentCard, Product}
 import com.gu.services.model.PaymentDetails
 import json.localDateWrites
 import play.api.libs.json.{Json, _}
@@ -68,8 +68,14 @@ object AccountDetails {
         case _ => true
       })
 
+      def externalisePlanName(plan: SubscriptionPlan.AnyPlan): Option[String] = plan.product match {
+        case _: Product.Weekly => if(plan.name.contains("Six for Six")) Some("currently on '6 for 6'") else None
+        case _: Product.Paper => Some(plan.name.replace("+",  " plus Digital Pack"))
+        case _ => None
+      }
+
       def jsonifyPlan(plan: SubscriptionPlan.AnyPlan) = Json.obj(
-        "name" -> plan.name,
+        "name" -> externalisePlanName(plan),
         "start" -> plan.start,
         // if the customer acceptance date is future dated (e.g. 6for6) then always display, otherwise only show if starting less than 30 days from today
         "shouldBeVisible" -> (subscription.acceptanceDate.isAfter(now) || plan.start.isBefore(now.plusDays(30)))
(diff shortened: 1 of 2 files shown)


pvighi:
just out of curiosity:
why does the api returns startsBeforeXDaysFromToday and a list of current and future plans instead of just returning a list of plans with the start date and let the clients decide how to classify them ?

If the calculation for startsBeforeXDaysFromToday is too complicated and we want to keep it in the server side is there a more descriptive name for the field ?

twrichards:
> just out of curiosity:
> why does the api returns startsBeforeXDaysFromToday and a list of current and future plans instead of just returning a list of plans with the start date and let the clients decide how to classify them ?
> 
> If the calculation for startsBeforeXDaysFromToday is too complicated and we want to keep it in the server side is there a more descriptive name for the field ?

@pvighi its intentionally server-side nut not so much for complexity but rather to use consistent date logic/library (rather than re-parsing the dates client-side and potentially also suffering from local clock hacks) - I have toyed with the name but settled for this, it's not perfect... if you have any suggestions?

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -1,18 +1,21 @@
...
   regNumber: Option[String],
   email: Option[String],
+  subscription : Subscription[SubscriptionPlan.AnyPlan],
   paymentDetails: PaymentDetails,
   stripePublicKey: String,
   accountHasMissedRecentPayments: Boolean,
   safeToUpdatePaymentMethod: Boolean,
   isAutoRenew: Boolean,
-  membershipAlertText: Option[String]
+  alertText: Option[String]
```
jacobwinch: 💯

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -60,16 +63,35 @@ object AccountDetails {
...
-        case Some(text) => Json.obj("alertText" -> text)
-        case None => Json.obj()
-      }
+      def planIsCurrent(plan: SubscriptionPlan.AnyPlan) = (plan.start == now || plan.start.isBefore(now)) && (plan match {
+        case paidPlan: SubscriptionPlan.Paid => paidPlan.end.isAfter(now)
+        case _ => true
+      })
+
+      def jsonifyPlan(plan: SubscriptionPlan.AnyPlan) = Json.obj(
+        "productRatePlanId" -> plan.productRatePlanId.get, // consider exposing hash of this to avoid exposing internal IDs
```
jacobwinch: Generally I am not keen on the idea of exposing productRatePlanIds. This is Zuora jargon and it would be nice if we could abstract it away at this level so clients didn't need to care about such things!

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -60,16 +63,35 @@ object AccountDetails {
...
+        "productName" -> plan.productName,
+        "name" -> plan.name,
```
jacobwinch: Same point as above really - I would prefer it if the rate plan names from the Zuora catalog didn't get passed all the way through to a client (appreciate that this is already being leaked in the current responses though)

jacobwinch: For example, we probably want clients to treat all the different versions of Guardian Weekly that we have in the product catalog the same, so it would be nicer if we defined a clean type here and returned that instead of the exact string from Zuora.

membership-attribute-service/app/models/AccountDetails.scala
twrichards: Removed in https://github.com/guardian/members-data-api/pull/371/commits/5b03a52c8c8b69960b7ea6928b19f9d909632f31 (and similarly in https://github.com/guardian/manage-frontend/commit/3dd81ad407ac94a95d6158ed5e428bb39e2365fc)

membership-attribute-service/app/models/AccountDetails.scala
twrichards: @jacobwinch I'll have a go at doing a cleaner function here

twrichards: done in https://github.com/guardian/members-data-api/pull/371/commits/89f6ea4f6f0f6f30eab0866210ff1739683f72b3 (with accompanying changes in https://github.com/guardian/manage-frontend/pull/196/commits/a0bb42cf24f8074eb3852161f9b094dce6210606)

PR #373, opened 2019-02-28
add logging for when there is more than one 'current plan'

twrichards:
### Why do we need this? 
Subscriptions shouldn't ever have multiple current plans... however this has been observed client side (due to https://github.com/guardian/manage-frontend/blob/a0bb42cf24f8074eb3852161f9b094dce6210606/app/shared/productResponse.ts#L122-L126) but want to log it here so we can investigate the specific subscriptions.


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -91,6 +92,12 @@ object AccountDetails {
         case _ => Json.obj()
       })
 
+      val sortedPlans = subscription.plans.list.sortBy(_.start.toDate)
```
paulbrown1982: You're not using 'sortedPlans' anywhere

PR #374, opened 2019-02-28
FIX: always use 'end' in the 'currentPlans' filter now it's available for all plans (see https://github.co…

twrichards:
Now that 'end' date is available for all plans (see https://github.com/guardian/membership-common/pull/590) use it in the 'currentPlans' filter (so that free plans like Friend which are ended, don't trump genuine current plans in the display)

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -1,5 +1,5 @@
 package models
-import com.gu.memsub.subsv2.{Subscription, SubscriptionPlan}
+import com.gu.memsub.subsv2.{PaidSubscriptionPlan, Subscription, SubscriptionPlan}
 import com.gu.memsub.{GoCardless, PayPalMethod, PaymentCard, Product}
 import com.gu.services.model.PaymentDetails
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -78,11 +74,11 @@ object AccountDetails {
       def jsonifyPlan(plan: SubscriptionPlan.AnyPlan) = Json.obj(
         "name" -> externalisePlanName(plan),
         "start" -> plan.start,
+        "end" -> plan.end,
         // if the customer acceptance date is future dated (e.g. 6for6) then always display, otherwise only show if starting less than 30 days from today
         "shouldBeVisible" -> (subscription.acceptanceDate.isAfter(now) || plan.start.isBefore(now.plusDays(30)))
       ) ++ (plan match {
-        case paidPlan: SubscriptionPlan.Paid => Json.obj(
-          "end" -> paidPlan.end,
+        case paidPlan: PaidSubscriptionPlan[_, _] => Json.obj(
```
twrichards: changing this line was to avoid a misleading Type Erasure warning (which came from using the type alias `SubscriptionPlan.Paid` which references type params which do get erased but aren't needed for the pattern match

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -93,8 +89,8 @@ object AccountDetails {
       })
 
       val sortedPlans = subscription.plans.list.sortBy(_.start.toDate)
-      val currentPlans = sortedPlans.filter(planIsCurrent)
-      val futurePlans = sortedPlans.filter(_.start.isAfter(now))
+      val currentPlans = sortedPlans.filter(plan => plan.start.isBefore(now.plusDays(1)) && plan.end.isAfter(now))
```
pvighi: maybe it might be more clear to say 
```scala
      val currentPlans = sortedPlans.filter(plan => !plan.start.isAfter(now) && plan.end.isAfter(now))

```
assuming that is what we are trying to do

pvighi: too late

twrichards: great suggestion, unfortunately just merged the PR. i'll do it in a follow up

twrichards: https://github.com/guardian/members-data-api/pull/375

PR #376, opened 2019-03-13
new endpoint for 'existing payment options'

twrichards:
### Why do we need this? 
In order to facilitate displaying **existing payment options** in the checkout flows (initially contributions https://github.com/guardian/support-frontend/pull/1673), (which in turn facilitate re-subscribes) we needed a new endpoint which can expose `billingAccountId` for each eligible account for the customer and enough payment detail to display.

### The changes

- New endpoint `/user-attributes/me/existing-payment-options` (with new accompanying controller `ExistingPaymentOptionsController`), which takes one mandatory query parameter `currencyFilter` which selects only billing accounts with the currency matching the provided currency ISO (e.g. `GBP`)
- enabled CORs access for `support-frontend` in all environments (as this will be the main consumer of this new endpoint)
- integrated IDAPI redirect service in order to show different level of detail for the existing payment options (full detail if signed in within the last hour, or simply the `paymentType` of each suitable account if signed in for longer than that)
- billing accounts are ONLY options in the response IF...
  - the term end date of a subscription under a billing account is greater than 3 months ago
  - the account has a default payment method
  - _the currency query parameter matches_
  - the payment method is still valid (for card that its not expired or about to expire, for direct debit that the `mandateId` is not null which occurs in mandate failure)
  - the account is not in payment failure (consecutive failed payments > 0)
  - the payment method must be in 'Active' status (not 'Closed')
- FINALLY, all the billing accounts are consolidated by the card/account number to avoid confusion for the user (which although the grouping is in fact just on the last 4 digit mask, the assumption that it is very unlikely for single person to have two cards/accounts ending in the same last four digits - 10<sup>8</sup> in fact, which is  1 in 100,000,000) - as it stands the first billingAccountId will be used for cloning the payment method but in future once we write 'cloned from' and 'cloned to' custom fields onto the payment methods we can perhaps make a nice choice)

image

#### When signed-in but not in the last hour...
image
... notice the lack of any sensitive information, just confirmation that the `?currencyFilter=` has been effective, and shows how many available options there are.
#### Full details (signed-in in the last hour)
image
... plus an example of consolidation (notice the one entry, but with the differing `billingAccountId`s in the subscriptions list, the first of which is at the top level of the available payment option)...
image


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -215,7 +215,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
     }
   }
 
-  def allSubscriptions(
+  def allCurrentSubscriptions(
```
twrichards: made the name more representative

membership-attribute-service/app/models/ExistingPaymentOption.scala
```diff
@@ -0,0 +1,56 @@
...
+      )
+      case _ => Json.obj()
+    }
+
+    private val sensitiveDetailIfApplicable = if (freshlySignedIn) {
+      Json.obj(
+        "billingAccountId" -> objectAccount.id.get,
+        "subscriptions" -> subscriptions.map(subscription => Json.obj(
+          "isCancelled" -> subscription.isCancelled,
+          "isActive" -> (!subscription.isCancelled && subscription.termEndDate.isAfter(now)),
```
jacobwinch: Bit of an edge case, but I think this check might accidentally exclude users with an auto-renewing sub who are due to renew later that day...

membership-attribute-service/conf/routes
```diff
@@ -2,6 +2,8 @@ GET         /healthcheck
 
 GET         /user-attributes/me                                                 controllers.AttributeController.attributes
 
+GET         /user-attributes/me/existing-payment-options                        controllers.ExistingPaymentOptionsController.existingPaymentOptions(cardEnabled: Boolean, directDebitEnabled: Boolean, currencyFilter: String)
```
jacobwinch: I'm not quite sure I understand the benefit of having a client pass through card/direct debit enabled - would you mind explaining the rationale for this?

membership-attribute-service/app/models/ExistingPaymentOption.scala
twrichards: Note that this doesn't exclude anything, it just sets some booleans in the response which is then used in the render in `support-frontend` to prefix the subscription name with 'current' or 'recent' depending on if `isActive` is true or false, respectively.
https://github.com/guardian/support-frontend/blob/bb12211b026c45d38ce0481ffaea809e4f013045/support-frontend/assets/helpers/existingPaymentMethods/existingPaymentMethods.js#L86

membership-attribute-service/conf/routes
twrichards: The main initial client `support-workers` will be sending these flags (see https://github.com/guardian/support-admin-console/pull/14). In the case where the login is valid but older than an hour, currently the response from this new endpoint will be really lightweight essentially just revealing how many (not any details about the type of payment method) meaning it can't be filtered client-side. 
If this is a problem, I could introduce an additional field (present regardless of staleness of the login) which describes the payment type, so it could filter client-side  - however IMO this is just the same net complexity. This approach also has the minor benefit of smaller response payload in the event that one of the switches is turned OFF.

membership-attribute-service/app/models/ExistingPaymentOption.scala
jacobwinch: My mistake - would probably just lead to a very short-lived copy bug then!

membership-attribute-service/conf/routes
jacobwinch: I'm not too keen on members-data-api having to care about switch state which belongs to `support-frontend` - IMO the responsibility for filtering is there. Not a huge problem though so if it's a lot of work feel free to disregard this comment

twrichards: OK will move it to the client

twrichards: done in https://github.com/guardian/members-data-api/pull/376/commits/0d360cb9b9e671f3e2b501b29e57d1368072779a (plus accompanying UI change https://github.com/guardian/support-frontend/pull/1673/commits/e9fc837b18e70edb819811e9f098a7e29f4c3aad)

membership-attribute-service/app/models/ExistingPaymentOption.scala
twrichards: yeah which i don't think is a problem, unless you're sugessting `isAfter(now.minusDays(1))` ?

twrichards: in fact DONE in https://github.com/guardian/members-data-api/pull/376/commits/520f4469089204cef9b4cb74fbc18773d4dd4b94

https://github.com/guardian/members-data-api/blob/520f4469089204cef9b4cb74fbc18773d4dd4b94/membership-attribute-service/app/models/ExistingPaymentOption.scala#L45

membership-attribute-service/app/models/ExistingPaymentOption.scala
```diff
@@ -0,0 +1,62 @@
...
+        ))) ++ sensitivePaymentPart
+    } else {
+      Json.obj()
+    }
+
+    def toJson = Json.obj(
+      "paymentType" ->  (paymentMethodOption match {
+        case Some(_: PaymentCard) => "Card"
+        case Some(_: GoCardless) => "DirectDebit"
+        case Some(_: PayPalMethod) => "PayPal"
```
jacobwinch: Do you think we will offer the ability to reuse PayPal in the future?

twrichards: Absolutely, it's _supported_ in CSR acquisition currently.

PR #379, opened 2019-03-22
NO LONGER disable auto-pay for self-service cancellation (so long as there are no raised invoices yet to be paid)

twrichards:
### Why do we need this? 
In order to facilitate offering resubscribe (aka existing-payment-options) to users with recently cancelled subscriptions, we should NO LONGER disable AutoPay on self-service cancellation because we don't want a new subscription to be added to a billing account with AutoPay OFF otherwise they will never be billed **(so long as there are no raised invoices yet to be paid)**

### The changes 
Move the `disableAutoPay` line within `executeCancellation` out to another function which wraps it in a check for invoice balances > 0 (i.e. only disable auto pay if that condition is true)

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -93,10 +93,15 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       }
     }
 
+    def disableAutoPayIfApplicable(zuoraSubscription: Subscription[P]) = for {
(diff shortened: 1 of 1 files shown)


twrichards:
CLOSING since we're going another way with the `support-workers` change that inspired this PR (we're planning to clone payment methods and billing accounts rather than creating multiple subs under the existing ones) - this change of direction is mainly down to how deeply the 'auto-pay' assumptions are baked in across many codebases

PR #380, opened 2019-03-26
Transition Members Data API to t3 instances

paulbrown1982:
### Why do we need this? 
Transition Members Data API to use t3 EC2 instance types, in preparation for 2019-20 reservations.

### The changes 
Cloudformation config.


paulbrown1982:
Tested in CODE.

PR #381, opened 2019-03-26
Revert "Transition Members Data API to t3 instances"

paulbrown1982:
Reverts guardian/members-data-api#380 if necessary.


jacobwinch:
Moving so many services to t3 is causing us problems so I'm going to back this change out whilst making the request mentioned below:

`Launching a new EC2 instance. Status Reason: You have requested more instances (21) than your current instance limit of 20 allows for the specified instance type. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit. Launching EC2 instance failed.`

PR #384, opened 2019-04-03
Integrate IDAPI as action wrapper

twrichards:
### Why do we need this? 
In order to facilitate 'auto sign-in tokens' (aka magic links) for the payment update flow (following CC Expiry, Direct Debit Mandate Failure and Payment Failure emails) `members-data-api` needs to be able to handle the new `SC_GU_RS` cookie (see https://github.com/guardian/identity/pull/1527) which is a 'scoped' sign-in cookie. With the accompanying `X_GU_ID_FORWARDED_SCOPE` header, the IDAPI redirect service is capable of interpreting this new cookie, the service also checks the recency of sign-in and checks for invalidated sessions (something that the existing cookie check mechanism in `members-data-api` cannot do) - so there are additional but very important side benefits to this work, beyond just facilitating  'auto sign-in tokens' (aka magic links).

### The changes 
Introduced `AuthAndBackendViaIdapiAction` which calls the IDAPI redirect service before processing the request and makes available some of the useful information that comes back. This will eventually replace the existing 'BackendFromCookieAction' (which just checks the cookie locally - using an identity library).
This new `AuthAndBackendViaIdapiAction` takes an ADT parameter (which informs it how to handle the redirect advice back from IDAPI), either...

- **`Return401IfNotSignedInRecently`** which if the `signInStatus` back from IDAPI is not `signedInRecently` then will return a 401 response (with the redirect URL in a new header `X-GU-IDAPI-Redirect`)
- **`ContinueRegardlessOfSignInRecency`** as in the case of `ExistingPaymentOptionsController.scala` where the recency of the sign-in is just used to limit the level of detail in the response.

### DIFF
--- a/membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
+++ b/membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
@@ -30,7 +30,9 @@ class AuthAndBackendViaIdapiAction(
 
       howToHandleRecencyOfSignedIn match {
         case Return401IfNotSignedInRecently if redirectAdvice.signInStatus != SignedInRecently =>
-          Left(Results.Unauthorized) //TODO add Location header containing returnUrl
+          Left(Results.Unauthorized.withHeaders(
+            ("X-GU-IDAPI-Redirect", redirectAdvice.redirect.map(_.url).getOrElse(""))
+          ))
         case _ =>
           Right(new AuthAndBackendRequest[A](redirectAdvice, backendConf, request))
       }
--- a/membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
+++ b/membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
@@ -1,6 +1,6 @@
 package actions
 
-import com.gu.identity.SignedInRecently
+import com.gu.identity.{IdapiService, SignedInRecently}
 import components.TouchpointBackends
 import filters.AddGuIdentityHeaders
 import play.api.mvc.{ActionRefiner, Request, Result, Results}
@@ -9,8 +9,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class AuthAndBackendViaIdapiAction(
   touchpointBackends: TouchpointBackends,
-  howToHandleRecencyOfSignedIn: HowToHandleRecencyOfSignedIn,
-  scope: Option[String]
+  howToHandleRecencyOfSignedIn: HowToHandleRecencyOfSignedIn
 )(
   implicit ex: ExecutionContext
 ) extends ActionRefiner[Request, AuthAndBackendRequest] {
@@ -18,8 +17,8 @@ class AuthAndBackendViaIdapiAction(
   override val executionContext = ex
   override protected def refine[A](request: Request[A]): Future[Either[Result, AuthAndBackendRequest[A]]] =
     touchpointBackends.normal.idapiService.RedirectAdvice.getRedirectAdvice(
-      request.headers.get("Cookie").getOrElse(""),
-      scope
+      request.headers.get(IdapiService.HeaderNameCookie).getOrElse(""),
+      request.headers.get(IdapiService.HeaderNameIdapiForwardedScope)
     ).map(redirectAdvice => {
 
       val backendConf = if (AddGuIdentityHeaders.isTestUser(redirectAdvice.userId)) {
--- a/membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
+++ b/membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
@@ -1,0 +1,39 @@
+package actions
+
+import com.gu.identity.SignedInRecently
+import components.TouchpointBackends
+import filters.AddGuIdentityHeaders
+import play.api.mvc.{ActionRefiner, Request, Result, Results}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class AuthAndBackendViaIdapiAction(
+  touchpointBackends: TouchpointBackends,
+  howToHandleRecencyOfSignedIn: HowToHandleRecencyOfSignedIn,
+  scope: Option[String]
+)(
+  implicit ex: ExecutionContext
+) extends ActionRefiner[Request, AuthAndBackendRequest] {
+
+  override val executionContext = ex
+  override protected def refine[A](request: Request[A]): Future[Either[Result, AuthAndBackendRequest[A]]] =
+    touchpointBackends.normal.idapiService.RedirectAdvice.getRedirectAdvice(
+      request.headers.get("Cookie").getOrElse(""),
+      scope
+    ).map(redirectAdvice => {
+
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
```diff
@@ -0,0 +1,40 @@
...
+)(
+  implicit ex: ExecutionContext
+) extends ActionRefiner[Request, AuthAndBackendRequest] {
+
+  override val executionContext = ex
+  override protected def refine[A](request: Request[A]): Future[Either[Result, AuthAndBackendRequest[A]]] =
+    touchpointBackends.normal.idapiService.RedirectAdvice.getRedirectAdvice(
+      request.headers.get(IdapiService.HeaderNameCookie).getOrElse(""),
+      request.headers.get(IdapiService.HeaderNameIdapiForwardedScope)
+    ).map(redirectAdvice => {
```
Mullefa: Considering the pain of changing membership common, I'm not advocating for an immediate change, but I found the class name `RedirectAdviceResponse` slightly confusing. Isn't `HowToHandleRecencyOfSignedIn` using it to offer redirect advice,  rather than `RedirectAdviceResponse` offering the advice itself?

membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
```diff
@@ -0,0 +1,40 @@
...
+      } else {
+        touchpointBackends.normal
+      }
+
+      howToHandleRecencyOfSignedIn match {
+        case Return401IfNotSignedInRecently if redirectAdvice.signInStatus != SignedInRecently =>
+          Left(Results.Unauthorized.withHeaders(
+            ("X-GU-IDAPI-Redirect", redirectAdvice.redirect.map(_.url).getOrElse(""))
+          ))
+        case _ =>
```
Mullefa: My preference would be to match on each case explicitly. It's clearer why we're taking the given action for each code path and it forces us to think about how we want handle any new ways of "how to handle recency of signed in" that are introduced in the future (especially if we add the compiler option to [fail on warnings](https://stackoverflow.com/questions/28327383/make-compile-fail-on-non-exhaustive-match-in-sbt)).

membership-attribute-service/app/actions/CommonActions.scala
```diff
@@ -1,15 +1,25 @@
...
+case object ContinueRegardlessOfSignInRecency extends HowToHandleRecencyOfSignedIn
+
 class CommonActions(touchpointBackends: TouchpointBackends, bodyParser: BodyParser[AnyContent])(implicit ex: ExecutionContext, mat:Materializer) {
   def noCache(result: Result): Result = NoCache(result)
 
   val NoCacheAction = resultModifier(noCache)
   val BackendFromCookieAction = NoCacheAction andThen new WithBackendFromCookieAction(touchpointBackends)
+  def AuthAndBackendViaIdapiAction(howToHandleRecencyOfSignedIn: HowToHandleRecencyOfSignedIn) =
+    NoCacheAction andThen new AuthAndBackendViaIdapiAction(touchpointBackends, howToHandleRecencyOfSignedIn)
+  def AuthAndBackendViaIdapiAction: ActionBuilder[AuthAndBackendRequest, AnyContent] =
```
Mullefa: My personal preference would be to omit this utility `def` and instead use `AuthAndBackendViaIdapiAction(Return401IfNotSignedInRecently)`. 

Pros: clearer what's going on in-line; cons: more verbose. If you did want a utility def, I'd advocate for referencing how we are handling a user not being signed in recently in the action builder name.

membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
```diff
@@ -72,9 +70,10 @@ class ExistingPaymentOptionsController(commonActions: CommonActions, override va
   def cardThatWontBeExpiredOnFirstTransaction(cardDetails: PaymentCardDetails) =
     new LocalDate(cardDetails.expiryYear, cardDetails.expiryMonth, 1).isAfter(now.plusMonths(1))
 
-  def existingPaymentOptions(currencyFilter: Option[String]) = BackendFromCookieAction.async { implicit request =>
+  def existingPaymentOptions(currencyFilter: Option[String]) = AuthAndBackendViaIdapiAction(ContinueRegardlessOfSignInRecency).async { implicit request =>
     implicit val tp = request.touchpoint
-    val maybeUserId = authenticationService.userId
+    val maybeUserId = request.redirectAdvice.userId
+    val isFreshlySignedIn = request.redirectAdvice.signInStatus == SignedInRecently
```
Mullefa: Variable name `isFreshlySignedIn` could be changed to `isSignedInRecently` to make it consistent with the check we're doing to derive the variable?

membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
twrichards: `RedirectAdviceResponse` contains the actual _advice_ (the reason I'm using the word _advice_ is because the IDAPI endpoint isn't perfectly named, since it doesn't actually perform a redirect, just returns some JSON advising what redirect should take place)... hence the response is the advice and `HowToHandleRecencyOfSignedIn` is the instruction to `AuthAndBackendViaIdapiAction` in how to action that advice.

membership-attribute-service/app/actions/CommonActions.scala
twrichards: good point - done in https://github.com/guardian/members-data-api/pull/384/commits/1ff58275001f05802ed8dfe034a0e36e40cbbafe

membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
twrichards: good point - done in https://github.com/guardian/members-data-api/pull/384/commits/1ff58275001f05802ed8dfe034a0e36e40cbbafe

membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
Mullefa: Ok - just got a little confused as I felt that "how to handle recency of signed in" constituted part of that redirect advice.

Mullefa: But see what you mean 👌

membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala
twrichards: sorry but i don't agree on this point - handling each case explicitly will just lead to repeated code and this is IMO clearer as it is, where its clear there is only one case (comprised of two conditions that must be met) where the response will be a 401 - otherwise the useful details (e.g. identity ID) is passed onto the main handler of the route

Mullefa: ok

PR #385, opened 2019-04-08
filter out non CreditCardReferenceTransaction for existing payment options

twrichards:
### Why do we need this? 
It turns out that card details entered directly into Zuora (`CreditCard` rather than `CreditCardReferenceTransaction`) are NOT clone-able and so break `support-workers`, hence these rare but valid payment method should not be available to select as an 'existing payment method'

### The changes 

- Following https://github.com/guardian/membership-common/pull/597 make use of `isReferenceTransaction` to make sure for card payments we only allow CreditCardReferenceTransaction through.
- Also given that `membership-common` had moved on in https://github.com/guardian/membership-common/pull/596 but the accompanying https://github.com/guardian/members-data-api/pull/384 is not yet complete a minor tweak needed to be made to the use of `IdapiService` in order for things to compile.

### DIFF
--- a/membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
+++ b/membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
@@ -81,7 +81,7 @@ class ExistingPaymentOptionsController(commonActions: CommonActions, override va
     val defaultMandateIdIfApplicable = "CLEARED"
 
     def paymentMethodStillValid(paymentMethodOption: Option[PaymentMethod]) = paymentMethodOption match {
-      case Some(card: PaymentCard) => card.paymentCardDetails.exists(cardThatWontBeExpiredOnFirstTransaction)
+      case Some(card: PaymentCard) => card.isReferenceTransaction && card.paymentCardDetails.exists(cardThatWontBeExpiredOnFirstTransaction)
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
```diff
@@ -81,7 +81,7 @@ class ExistingPaymentOptionsController(commonActions: CommonActions, override va
     val defaultMandateIdIfApplicable = "CLEARED"
 
     def paymentMethodStillValid(paymentMethodOption: Option[PaymentMethod]) = paymentMethodOption match {
-      case Some(card: PaymentCard) => card.paymentCardDetails.exists(cardThatWontBeExpiredOnFirstTransaction)
+      case Some(card: PaymentCard) => card.isReferenceTransaction && card.paymentCardDetails.exists(cardThatWontBeExpiredOnFirstTransaction)
```
johnduffell: looks good re the changes.
also I know it's old code but in the original code I'd probably pass in the PaymentMethod itself rather than an Option[PaymentMethod] to the functions here, as the functions wouldn't do anything in the case of a None so it's an extra input that's not needed.  This would also make unit testing easier (if members-data-api had them)

johnduffell:
I can't see the version bump in the build.sbt? normally it would be 0.1-SNAPSHOT when membership-common hasn't been shipped yet.  Or did you exclude that from what you pushed?
Anyway, assuming that's the case once that once the build passes it looks good to me! 👍

PR #66, opened 2015-12-04
Membership common 114

tomverran:

PR #387, opened 2019-04-15
remove 'pendingAmendment' from calculation of "cancelledAt" field in the output 

twrichards:
no historical PR description to explain why (https://github.com/guardian/members-data-api/pull/66) - but is clearly wrong and was causing misinformation (identified by a userhelp case)

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -110,7 +110,7 @@ object AccountDetails {
             "nextPaymentDate" -> paymentDetails.nextPaymentDate,
             "lastPaymentDate" -> paymentDetails.lastPaymentDate,
             "renewalDate" -> paymentDetails.termEndDate,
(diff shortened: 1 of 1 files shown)


johnduffell:
sounds good, the membership downgrade to friend is effectively a cancellation, so it's worth bearing that in mind, but since membership is just ticking over it's probabyl not going to cause a big issue.
👍 thanks for investigating!

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -110,7 +110,7 @@ object AccountDetails {
             "nextPaymentDate" -> paymentDetails.nextPaymentDate,
             "lastPaymentDate" -> paymentDetails.lastPaymentDate,
             "renewalDate" -> paymentDetails.termEndDate,
-            "cancelledAt" -> (paymentDetails.pendingAmendment || paymentDetails.pendingCancellation),
```
paulbrown1982: 😱

PR #388, opened 2019-04-30
Fetch one-off contributions from postgres

tomrf1:
~~Merging into @jranks123 's original branch.~~ (switched to master to avoid confusion)

We want to expose one-off contributions data (which is in a postgres table, not zuora) in 2 ways:
1. A full history of one-off contributions, for use on the manage site
2. Take into account the last one-off contribution when setting the `showSupportMessaging` field in the attributes response. This means storing the date of the last one-off in the dynamodb cache

TODO:
- [x] Only use/expose one-off data for email-verified accounts
- [x] Decide how recent the last one-off needs to be to hide epic/banner
- [ ] Add tests
- [ ] Consider how read capacity of postgres will scale


membership-attribute-service/app/services/ContributionsStoreDatabaseService.scala
```diff
@@ -1,60 +1,63 @@
...
-import cats.syntax.applicativeError._
-import cats.instances.future._
 import com.typesafe.scalalogging.StrictLogging
 import models.ContributionData
 import play.api.db.Database
+import services.DatabaseService.DatabaseGetResult
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.control.NonFatal
+import scalaz.{EitherT, \/}
```
tomrf1: reluctantly switching to scalaz here because the calling code uses it heavily, and it would get really messy otherwise.
It would be nice to rip out scalaz completely, there's no reason to use `\/` these days

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -41,7 +42,8 @@ case class Attributes(
   lazy val isPatronTier = Tier.exists(_.equalsIgnoreCase("patron"))
   lazy val isStaffTier = Tier.exists(_.equalsIgnoreCase("staff"))
   lazy val isPaidTier = isSupporterTier || isPartnerTier || isPatronTier || isStaffTier
-  lazy val isContributor = RecurringContributionPaymentPlan.isDefined
+  //TODO - only use OneOffContributionDate for verified users
+  lazy val isContributor = RecurringContributionPaymentPlan.isDefined || OneOffContributionDate.isDefined
```
jranks123: I think we should separate out one-offs and recurring here

tomrf1: good point, `isContributor` is used for more than just setting `showSupportMessaging`

jranks123: in fact, further down, this variable passed as to `ContentAccess` as `recurringContributor`

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -53,15 +54,16 @@ case class Attributes(
...
     digitalPack = digitalSubscriberHasActivePlan,
     paperSubscriber = isPaperSubscriber,
     guardianWeeklySubscriber = isGuardianWeeklySubscriber
   )
 
   // show support messaging (in app & on dotcom) if they do NOT have any active products
   // TODO in future this could become more sophisticated (e.g. two weeks before their products expire)
-  lazy val showSupportMessaging = !(isPaidTier || isContributor || digitalSubscriberHasActivePlan || isPaperSubscriber || isGuardianWeeklySubscriber)
+  lazy val showSupportMessaging =
+    !(isPaidTier || isRecurringContributor || isOneOffContributor || digitalSubscriberHasActivePlan || isPaperSubscriber || isGuardianWeeklySubscriber)
```
jranks123: The current rule is to not show asks to people who have made a contribution *in the past 6 months*, so maybe `isRecentOneOffContributor` would be more appropriate.

tomrf1:
Tested in CODE with verified + unverified users

membership-attribute-service/conf/application.conf
```diff
@@ -24,4 +24,22 @@ identity.prerelease-users = []
...
+    executor = "thread-pool-executor"
+    throughput = 1
+    thread-pool-executor {
+      fixed-pool-size = 30
+    }
+  }
+}
+
+db.oneOffStore {
+  url=null
```
joelochlann: do these come from S3? did we create a read-only user for the members-data-api?

tomrf1: yes they'll be in the s3 config file. I'll make sure there's a new read-only user (have done so in CODE)
I've set them to null here for the sake of documentation, but open to other ideas.
The url will be that of the read replica

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -121,6 +126,7 @@ object Attributes {
     (__ \ "userId").write[String] and
       (__ \ "tier").writeNullable[String] and
       (__ \ "recurringContributionPaymentPlan").writeNullable[String] and
+      (__ \ "oneOffContributionDate").writeNullable[LocalDate].contramap[Option[LocalDate]](_ => None) and  //omit
```
paulbrown1982: Why do you need the contramap here?

membership-attribute-service/app/services/ContributionsStoreDatabaseService.scala
```diff
@@ -0,0 +1,66 @@
...
+    Future(database.withConnection { implicit conn =>
+      statement.as(parser)
+    })
+      .map(\/.right)
+      .recover { case NonFatal(err) =>
+        \/.left(s"Error querying contributions store. Error: $err")
+      }
+
+  override def getAllContributions(identityId: String): DatabaseGetResult[List[ContributionData]] = {
+    val statement = SQL"""
```
paulbrown1982: Should Identity ID be forced into the format "integer string" to protect against SQL injection? Or written another way: Do we need a type: NumericString, or maybe just use Integer, if that's what's the identity_id type is in the PostGres table?

membership-attribute-service/app/models/Attributes.scala
tomrf1: to exclude this field from the api response by setting it to `None`

membership-attribute-service/app/services/ContributionsStoreDatabaseService.scala
tomrf1: `identity_id` has type `text` - https://github.com/guardian/contributions-platform/blob/master/contributions-store/sql/create-table.sql#L17
Though it looks like our identity ids are always numeric.
I can add some (scala) code to verify that the string contains a numeric value - is that what you mean?

tomrf1: Ok, at this point in the code it must be an authenticated user id so I think it's pretty safe as it is

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -99,4 +120,20 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
...
   def features = lookup("features", onSuccessMember = Features.fromAttributes, onSuccessSupporter = _ => Features.unauthenticated, onNotFound = Features.unauthenticated)
 
+
+  def oneOffContributions = {
+    AuthAndBackendViaIdapiAction(Return401IfNotSignedInRecently).async { implicit request =>
+      if (request.redirectAdvice.emailValidated.contains(true)) {
+        authenticationService.userId(request) match {
+          case Some(identityId) =>
+            oneOffContributionDatabaseService.getAllContributions(identityId).map {
+              case -\/(err) => Ok(err)
```
johnduffell, line 130: was there a reason to return errors as a 200 OK?  Probably shouldn't expose the error messsage in the rresponse though.

PR #392, opened 2019-07-29
Add a logline on each call to a deprecated endpoint.

lmath:
### Why do we need this? 
We have a number of deprecated endpoints in members-data-api. 

I'd like to make it so that we can search `DeprecatedEndpointCalled` in cloudwatch and see which endpoints are still in use.

### The changes 
* Adds a logline that logs `DeprecatedEndpointCalled`, the request method and path, the request headers and the request body for each deprecated endpoint

There is going to be some work to follow that changes how we know cookies and tokens are valid in identity (see: https://trello.com/c/DjF6XvBE/1539-members-data-api-uses-identity-auth). We will need to eventually decide if 
1) deprecated endpoints be updated with this new definition of validity
2) we should remove the deprecated endpoints

So, we might as well start logging now.

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -277,7 +277,10 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   }
 
   private def updateContributionAmount(subscriptionNameOption: Option[memsub.Subscription.Name]) = BackendFromCookieAction.async { implicit request =>
-    DeprecatedRequestLogger.logDeprecatedRequest(request)
+    if(subscriptionNameOption.isEmpty){
+      DeprecatedRequestLogger.logDeprecatedRequest(request)
+    }
+
     val updateForm = Form { single("newPaymentAmount" -> bigDecimal(5, 2)) }
     val tp = request.touchpoint
     val maybeUserId = authenticationService.userId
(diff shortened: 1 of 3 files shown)


lmath:
Q: If there is no traffic for an endpoint for two weeks, is that enough evidence it's worth removing the endpoint?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -274,6 +277,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   }
 
   private def updateContributionAmount(subscriptionNameOption: Option[memsub.Subscription.Name]) = BackendFromCookieAction.async { implicit request =>
```
twrichards: this function is not deprecated - just the endpoint of the same name is, in favour of `/updateAmountForSpecificContribution` see https://github.com/guardian/members-data-api/blob/09ab6bafaa0b77eb8106b78568ee7911347f961a/membership-attribute-service/conf/routes#L12 and https://github.com/guardian/members-data-api/blob/09ab6bafaa0b77eb8106b78568ee7911347f961a/membership-attribute-service/app/controllers/AccountController.scala#L305

twrichards: think you might want to do an if statement a bit like the `lookup` one you've done

lmath: Ooof, good point. See: https://github.com/guardian/members-data-api/pull/392/commits/b1bfde67af5a835f0645283bebe8b2baaa7582cd

johnduffell:
nice work! 👍 header should let us know if it's app or web which would be useful.  Due to the existing log retention period there shouldn't be any issue with data retention.
I think a 2 week period would be ample to monitor, as most things would happen fairly regularly, at least once a week if not once a day.

PR #393, opened 2019-08-12
Authenticate users using identity-auth-play library

lmath:
### Why do we need this? 

Our definition of how to authenticate users is changing. 

As of
https://github.com/guardian/identity/pull/1571

Determining if a user is authenticated now requires a call back to IDAPI (so that if we've invalidated their session, we can say any cookies/tokens generated before the time the user was invalidated cannot be used to authenticate a user). 

So, we are phasing out the identity-cookie library and the identity-play-auth library which check the expiry and the signature of a cookie/token but do not check if we've invalidated the user. 

### The changes 

For `members-data-api`, this means
* Requests that authenticate a user should make use of either `AuthAndBackendViaIdapiAction` or `AuthAndBackendViaAuthLibAction`
* `AuthAndBackendViaAuthLibAction` replaces `BackendFromCookieAction`, so each request using `AuthAndBackendViaAuthLibAction` will make a request to IDAPI to authenticate the user, then and include `Some(User)` (if they've authenticated, or None if they have not) along with the version of backend services to be used with the request (depending on if it's a test user or not)

https://trello.com/c/DjF6XvBE/1539-members-data-api-uses-identity-auth


membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
```diff
@@ -0,0 +1,30 @@
...
+import play.api.mvc.{ActionRefiner, Request, Result, Results}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class AuthAndBackendViaAuthLibAction(touchpointBackends: TouchpointBackends)(implicit ex: ExecutionContext) extends ActionRefiner[Request, AuthenticatedUserAndBackendRequest] {
+  override val executionContext = ex
+
+  override protected def refine[A](request: Request[A]): Future[Either[Result, AuthenticatedUserAndBackendRequest[A]]] = {
+    // On each request via this action, we make a call to IDAPI and see if we can authenticate the user.
+    // The test config and the normal config are the same for IDAPI.
```
johnduffell: that is a little bit worrying, if they really are the same, they shouldn't really be in the touchpoint backend. and if they may be different, we should do the test user check before contacting identity if possible

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -60,17 +60,16 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
   }
 
   private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessSupporter: Attributes => Result, onNotFound: Result, sendAttributesIfNotFound: Boolean = false) = {
-    AuthAndBackendViaIdapiAction(ContinueRegardlessOfSignInRecency).async { implicit request: AuthAndBackendRequest[AnyContent] =>
-      if(endpointDescription == "membership" || endpointDescription == "features") {
-        DeprecatedRequestLogger.logDeprecatedRequest(request)
-      }
+    AuthAndBackendViaAuthLibAction.async { implicit request =>
+
+      val userHasValidatedEmail = request.user.flatMap(_.statusFields.userEmailValidated).getOrElse(false)
```
johnduffell: out of interest, nowadays is this field only true if the email is valid and we are sure the password is owned by the email owner.  or is it still the case that it's possible to validate the email address without knowing the password?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -60,17 +60,16 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
   }
 
   private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessSupporter: Attributes => Result, onNotFound: Result, sendAttributesIfNotFound: Boolean = false) = {
-    AuthAndBackendViaIdapiAction(ContinueRegardlessOfSignInRecency).async { implicit request: AuthAndBackendRequest[AnyContent] =>
-      if(endpointDescription == "membership" || endpointDescription == "features") {
```
johnduffell: has the need for this logging gone away now?

membership-attribute-service/app/monitoring/ErrorHandling.scala
```diff
@@ -35,8 +37,8 @@ class ErrorHandler(
   }
 
   override protected def onProdServerError(request: RequestHeader, ex: UsefulException): Future[Result] = {
-       SafeLogger.error(scrub"Error handling request request: $request", ex)
-       Future { AddGuIdentityHeaders.headersFor(request, internalError) }
+    SafeLogger.error(scrub"Error handling request request: $request", ex)
+    AddGuIdentityHeaders.headersFor(request, internalError, identityAuthService)
```
johnduffell: thanks for fixing the indentation while you were in

membership-attribute-service/app/services/IdentityAuthService.scala
```diff
@@ -1,24 +1,38 @@
...
+  val identityPlayAuthService = IdentityPlayAuthService.unsafeInit(idApiUrl, apiConfig.token, Some("membership"))
+
+  def user(implicit requestHeader: RequestHeader): Future[Option[User]] = {
+    getUser(requestHeader)
+      .map(user => Option(user))
+      .handleError { err =>
+        if(err.isInstanceOf[UserCredentialsMissingError])
+          SafeLogger.info(s"unable to authorize user - $err")
+        else
+          SafeLogger.error(scrub"unable to authorize user - $err")
```
johnduffell: is "err" an exception? very minor, but if so I think you can pass it to the second parameter of `SafeLogger.error` in order to get the stack trace in the logs

johnduffell:
looks good to me, I have mentioned a few comments but nothing to stop it going out so but might be useful - thanks for doing this though!
Also, as you know, members data api has a fair amount of traffic so be aware that idenetity will get more traffic.  If you are concerned about the level of traffic for any reason, it would be possible to put in a switch so that if identity is struggling it can be turned off more quickly than a redeploy.
👍

membership-attribute-service/app/controllers/AttributeController.scala
lmath: Oh good spot; this was a merge accident! I will put it back

membership-attribute-service/app/filters/AddGuIdentityHeaders.scala
```diff
@@ -11,24 +12,29 @@ import scala.concurrent.{ExecutionContext, Future}
...
 
-  def headersFor(request: RequestHeader, result: Result) = (for {
-    user <- IdentityAuthService.playAuthService.authenticatedUserFor(request)
-  } yield result.withHeaders(
-    "X-Gu-Identity-Id" -> user.id,
-    "X-Gu-Identity-Credentials-Type" -> user.credentials.getClass.getSimpleName,
-    "X-Gu-Membership-Test-User" -> isTestUser(user.displayName).toString
-    )).getOrElse(result)
+  def headersFor(request: RequestHeader, result: Result, identityAuthService: IdentityAuthService)(implicit ec: ExecutionContext): Future[Result] = {
+    for {
```
tomrf1: a tiny bit neater?
```
identityAuthService.user(request).map {
  case Some(user) => ...
  case None => ...
}
```

membership-attribute-service/app/controllers/AttributeController.scala
lmath: I think it's still possible to validate an email via SSO if the user has no password, but as of https://github.com/guardian/identity/pull/1545 it shouldn't be possible to validate via SSO when we don't know the password is owned by the email owner

membership-attribute-service/app/services/IdentityAuthService.scala
lmath: Good point. Will do.

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: didn't realise the SSO route existed. the route I was thinking of was validation via email link - certainly in the past you could click a verify email link which would show the log in screen, but would already have marked the email address as valid even if you didn't actually log in.

membership-attribute-service/app/filters/AddGuIdentityHeaders.scala
lmath: oh yes I will do that

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: I have checked and that route still exists - I have validated 101682274 just by clicking a link in incognito mode and it's showing me the "enter your email address" screen

lmath: my team'll discuss this: https://trello.com/c/yjif1Gc0/1584-stop-making-it-possible-to-mark-an-account

thanks for walking me through it

PR #394, opened 2019-08-13
Revert "Authenticate users using identity-auth-play library"

lmath:
Reverts guardian/members-data-api#393

There were more errors in the logs than expected so reverting to investigate


lmath:
@johnduffell 👍 'd in person

PR #395, opened 2019-08-14
Revert "Revert "Authenticate users using identity-auth-play library""

lmath:
It was necessary to revert https://github.com/guardian/members-data-api/pull/393 because of a requests that were submitting crypto access tokens to members-data-api weren't authenticating correctly.

We are changing how we do authentication across the apps, and determining if a cookie or token is valid now requires calling back to IDAPI, so that we can check if we invalidated the user. More details on:
https://github.com/guardian/members-data-api/pull/393 

This PR uses the version of `identity-auth-play` with a fix, and also updates log levels when a user fails to authenticate.

### The fix
This PR uses version 3.184 of `identity-auth-play` which contains the fix.

identity-auth-play will take a request from a play app, and interact with IDAPI to return a `User` if the user can be authenticated. When we were attempting to authenticate a user with via a crypto access token, we were sending header `X-GU-Crypto-Access-Token`: \<token value\> but IDAPI was expecting `X-GU-Crypto-Access-Token`:Bearer \<token value\> and thus failing to recognise that a crypto access token had been included in the request. 

https://github.com/guardian/identity/pull/1590/files introduced the fix, available via version 3.184 used in this PR. 

Here's a diagram of how authenticating crypto access tokens works in the happy case
image

### Testing in CODE 
I obtained a crypto access token in CODE with targetClientId = membership (how to do that here: https://github.com/guardian/identity-platform/blob/master/docs/03-access-tokens/01-access-tokens-overview.md) 

I made this Happy Path request:
```
curl -X GET \
  https://members-data-api.code.dev-theguardian.com/user-attributes/me \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-us' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Host: members-data-api.code.dev-theguardian.com' \
  -H 'Postman-Token: 1b5946c9-932a-41b0-b088-7cd7992551b6,89081766-3cbb-4be1-933d-29ef22bc419f' \
  -H 'User-Agent: Guardian/17422 CFNetwork/978.0.7 Darwin/18.5.0' \
  -H 'accept-encoding: gzip, deflate' \
  -H 'cache-control: no-cache' \
  -H 'gu-identitytoken: <that token I obtained>'

```
on 3 different builds on CODE 

| description of build  |  expected behaviour | expected behaviour verified |
|---|---|---|
| 10316 - the build with identity-play-auth 2.5, the library we are phasing out that does not call back to IDAPI  | Happy path request returns 200 and the user's data. Sending a nonsense token or no token returns 401. | ✅ |
| 10332 - the build with the changes on this PR, using identity-auth-play 3.184, the latest version of the library that authenticates via calling back to IDAPI  |   Happy path request returns 200 and the user's data. Sending a nonsense token or no token returns 401. | ✅ |
| 10324 - the build that used identity-auth-play 3.184-M7 and did not send the header IDAPI expected  | Happy path request returns 401. Sending a nonsense token or no token also returns 401.  | ✅ |

and also double checked

| description of build  |  expected behaviour | expected behaviour verified |
|---|---|---|
| 10332 - the build with the changes on this PR, using identity-auth-play 3.184, the latest version of the library that authenticates via calling back to IDAPI   | that a happy path request with cookies still returns user data, and no cookies/nonsense values for cookies still return 401. | ✅ |

I'm hoping that I can get a 👍 from both an identity dev and a RR dev on this one


membership-attribute-service/app/services/IdentityAuthService.scala
```diff
@@ -1,24 +1,38 @@
...
+  val idApiUrl = Uri.unsafeFromString(apiConfig.url)
+
+  val identityPlayAuthService = IdentityPlayAuthService.unsafeInit(idApiUrl, apiConfig.token, Some("membership"))
+
+  def user(implicit requestHeader: RequestHeader): Future[Option[User]] = {
+    getUser(requestHeader)
+      .map(user => Option(user))
+      .handleError { err =>
+        if(err.isInstanceOf[UserCredentialsMissingError])
+          SafeLogger.error(scrub"unable to authorize user - $err", err)
```
lmath: `identity-auth-play` will try to authenticate the user and this needs either a `SC_GU_U` cookie or a `GU-IdentityToken` header, and will throw a `UserCredentialsMissingError` the request doesn't have appropriate credentials to even try and authenticate with IDAPI. 

It seems like UserCredentialsMissingError should be an error log (something is submitting bad requests to members-data-api) 

IDAPI returning unauthorised can actually just be business as usual, because the cookie/token is expired or otherwise invalid, so I'm suggesting putting that as a warn.

lmath: I got these backwards in https://github.com/guardian/members-data-api/pull/393, so the log levels in this handleError block and the version of identity-auth-play are the only changes in this PR vs https://github.com/guardian/members-data-api/pull/393.

johnduffell: might be worth making it clearer in the warning message that it's an expected case, just in case someone sees it in future

membership-attribute-service/app/services/IdentityAuthService.scala
```diff
@@ -1,24 +1,38 @@
...
+        else
+          SafeLogger.warn(s"unable to authorize user - $err")
```
johnduffell:
```suggestion
          SafeLogger.warn(s"valid request but expired token so user must log in again - $err")
```

membership-attribute-service/app/services/IdentityAuthService.scala
johnduffell:
```suggestion
          SafeLogger.error(scrub"invalid request as no token provided - unable to authorize user - $err", err)
```

lmath:
@tomrf1 oops yes, I do mean 3.184 and I've updated the description.
@johnduffell I've applied your logline suggestions. Good idea.

PR #396, opened 2019-08-16
Return last single contribution date

tomrf1:
Makes the /me endpoint return a `oneOffContributionDate` field, which is the date of the last single contribution. Note - this is for email-validated users only.
This is useful for showing different messaging on the website after users have passed their 6 month ask-pause.

Tested in CODE:
Screenshot 2019-08-16 at 11 43 46


jranks123:
@tomrf1 is it too late to change this to "lastContributionDate"?

PR #397, opened 2019-08-21
Downgrade logging to warn when there is no token/cookie provided.

lmath:
### Why do we need this? 
https://github.com/guardian/members-data-api/pull/395 moved members-data-api to use the new identity auth library

I mistakenly thought that requests to members-data-api that did not include any way to even attempt to authenticate a user would be relatively rare (since the only endpoint that doesn't need an authenticated user is the healthcheck). 

I was wrong. There are so many requests that that error logging ate up a bunch of the sentry quota. I'd like to downgrade this log line to a warn, but add some info about the request that has come in to help investigate what is happening. 

### The changes 
* Downgrades the case where the the request doesn't have a means to authenticate a user to a warn
* Adds request logging for debugging, so that I can look into it further.

### DIFF
--- a/membership-attribute-service/app/services/IdentityAuthService.scala
+++ b/membership-attribute-service/app/services/IdentityAuthService.scala
@@ -22,7 +22,8 @@ class IdentityAuthService(apiConfig: IdapiConfig)(implicit ec: ExecutionContext)
       .map(user => Option(user))
       .handleError { err =>
         if(err.isInstanceOf[UserCredentialsMissingError])
-          SafeLogger.error(scrub"invalid request as no token or cookie provided - unable to authorize user - $err", err)
+          SafeLogger.warn(s"invalid request as no token or cookie provided - unable to authorize user - $err." +
+            s"We received headers ${requestHeader.headers}", err)
         else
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/services/IdentityAuthService.scala
```diff
@@ -22,7 +22,8 @@ class IdentityAuthService(apiConfig: IdapiConfig)(implicit ec: ExecutionContext)
       .map(user => Option(user))
       .handleError { err =>
         if(err.isInstanceOf[UserCredentialsMissingError])
-          SafeLogger.error(scrub"invalid request as no token or cookie provided - unable to authorize user - $err", err)
+          SafeLogger.warn(s"invalid request as no token or cookie provided - unable to authorize user - $err." +
+            s"Request attempted: ${requestHeader.method} ${requestHeader.path} with headers ${requestHeader.headers}", err)
```
johnduffell: have you tested locally to see how this is formatted? I only ask because it might be worth considering a mkString('\n') or some other tweaking to make it more readable

lmath: Good idea. Adding in some newlines the logging would now look like: 
image

PR #398, opened 2019-08-22
Tidy up logging when incoming requests cause a UserCredentialsMissingError

lmath:
As of https://github.com/guardian/members-data-api/pull/395 members-data-api authenticates users using the new library identity-auth-play. As part of those changes, we logged when a request came in that didn't contain either a SC_GU_U cookie or a crypto access token. To get any meaningful response from members-data-api, we need an authenticated user so it seemed like those errors should have been relatively sparse.

They weren't.

It's because frontend doesn't bother calling members-data-api if the user is clearly not logged in, but the js code is unable to check if SC_GU_U is present. The best check that's available is to check for a GU_U cookie. So, this is not an error but actually expected behaviour. I've made this info logging, and removed the logging that logged the request headers. 

### Why do we need this? 
- The logging should be updated to make it clearer that this is expected behaviour.

### The changes 
- When we attempt to authenticate a user but this creates a UserCredentialsMissingError, this is now an info log because it's actually expected behaviour. 
- We are no longer logging request headers in that case because that was some temporary logging adding for debugging.

https://trello.com/c/DjF6XvBE/1539-members-data-api-uses-identity-auth

https://github.com/guardian/members-data-api/pull/395
https://github.com/guardian/members-data-api/pull/397

### DIFF
--- a/membership-attribute-service/app/services/IdentityAuthService.scala
+++ b/membership-attribute-service/app/services/IdentityAuthService.scala
@@ -22,8 +22,9 @@ class IdentityAuthService(apiConfig: IdapiConfig)(implicit ec: ExecutionContext)
       .map(user => Option(user))
       .handleError { err =>
         if(err.isInstanceOf[UserCredentialsMissingError])
-          SafeLogger.warn(s"invalid request as no token or cookie provided - unable to authorize user. \n" +
-            s"Request attempted: ${requestHeader.method} ${requestHeader.path} with headers:\n ${requestHeader.headers.toMap.mkString("\n")}", err)
+          //IdentityPlayAuthService throws an error if there is no SC_GU_U cookie or crypto auth token
+          //frontend decides to make a request based on the existence of a GU_U cookie, so this case is expected.
+          SafeLogger.info(s"unable to authorize user - no token or cookie provided")
         else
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/services/IdentityAuthService.scala
```diff
@@ -22,8 +22,9 @@ class IdentityAuthService(apiConfig: IdapiConfig)(implicit ec: ExecutionContext)
       .map(user => Option(user))
       .handleError { err =>
         if(err.isInstanceOf[UserCredentialsMissingError])
-          SafeLogger.warn(s"invalid request as no token or cookie provided - unable to authorize user. \n" +
-            s"Request attempted: ${requestHeader.method} ${requestHeader.path} with headers:\n ${requestHeader.headers.toMap.mkString("\n")}", err)
```
lmath: This was introduced in https://github.com/guardian/members-data-api/pull/397 as a temporary measure for debugging.

lmath: Logging the request headers in this case, I mean

johnduffell:
nice description, I like the dramatic way you gave "they weren't" its own paragraph 😆

PR #405, opened 2019-09-17
Add some extra logging to troubleshoot caching Zuora data

kelvin-chappell:
### Why do we need this? 
There are some cases where an anomaly may be causing the Zuora lookup to use the wrong Identity ID to find a subscription.  These extra logging lines should prove that this is what's happening.

### The changes 
Some extra cloudwatch logging lines.


arausuy:
Lets start here and see what comes out from these changes

PR #407, opened 2019-09-26
Support mobile subscriptions

alexduf:
### Why do we need this? 
This will add a new field allowing the members data api to display if a user had a mobile subscription.
It will have a non trivial impact on the load (and therefore cost) of the mobile API, but I believe this is justified.

For instance, this would allow a user to see their mobile subscription in their dotcom profile, or we could have a cross platform ad-free experience.

It will be used by the mobile app as a way to provide cross device service, so long as the user is logged in.

### The changes 
 - Slightly refactor the AttributesController for readability and performance. Futures need to be triggered before a for comprehension to be executed in parallel. This might reduce latency.
 - Add a new service that queries the mobile purchases service

### TODO

- [x] Add configuration for CODE and PROD with relevant API Keys
- [x] Test on CODE

### DIFF
--- a/membership-attribute-service/app/configuration/Config.scala
+++ b/membership-attribute-service/app/configuration/Config.scala
@@ -53,7 +53,6 @@ object Config {
   }
 
   object Mobile {
-    val subscriptionURL = config.getConfig("mobile.subscription.base")
     val subscriptionApiKey = config.getConfig("mobile.subscription.apiKey")
   }
 
--- a/membership-attribute-service/app/configuration/Config.scala
+++ b/membership-attribute-service/app/configuration/Config.scala
@@ -52,4 +52,9 @@ object Config {
     val enabled = Try{config.getBoolean("logstash.enabled")}.toOption.contains(true)
   }
 
+  object Mobile {
+    val subscriptionURL = config.getConfig("mobile.subscription.base")
+    val subscriptionApiKey = config.getConfig("mobile.subscription.apiKey")
+  }
+
 }
--- a/membership-attribute-service/app/configuration/Config.scala
+++ b/membership-attribute-service/app/configuration/Config.scala
@@ -53,7 +53,7 @@ object Config {
   }
 
   object Mobile {
-    val subscriptionApiKey = config.getConfig("mobile.subscription.apiKey")
+    val subscriptionApiKey: String = config.getString("mobile.subscription.apiKey")
   }
 
 }
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -86,7 +86,7 @@ class AttributeController(
     val mobileExpiryDate = mobileSubscriptionStatus.map(_.endDate.toLocalDate)
     zuoraAttributes.map(_.copy(
       OneOffContributionDate = latestOneOffDate,
-      MobileSubscriptionExpiryDate = mobileExpiryDate
+      LiveAppSubscriptionExpiryDate = mobileExpiryDate
     ))
   }
 
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -18,9 +18,16 @@ import org.joda.time.LocalDate
 import scalaz.{-\/, \/-}
 
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
+import scala.util.control.NonFatal
 
-class AttributeController(attributesFromZuora: AttributesFromZuora, commonActions: CommonActions, override val controllerComponents: ControllerComponents, oneOffContributionDatabaseService: OneOffContributionDatabaseService) extends BaseController with LoggingWithLogstashFields {
+class AttributeController(
+  attributesFromZuora: AttributesFromZuora,
+  commonActions: CommonActions,
+  override val controllerComponents: ControllerComponents,
+  oneOffContributionDatabaseService: OneOffContributionDatabaseService,
+  mobileSubscriptionService: MobileSubscriptionService
+) extends BaseController with LoggingWithLogstashFields {
   import attributesFromZuora._
   import commonActions._
   implicit val executionContext: ExecutionContext = controllerComponents.executionContext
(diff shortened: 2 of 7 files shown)


membership-attribute-service/app/models/Attributes.scala
```diff
@@ -131,6 +132,7 @@ object Attributes {
       (__ \ "digitalSubscriptionExpiryDate").writeNullable[LocalDate] and
       (__ \ "paperSubscriptionExpiryDate").writeNullable[LocalDate] and
       (__ \ "guardianWeeklyExpiryDate").writeNullable[LocalDate] and
+      (__ \ "mobileSubscriptionExpiryDate").writeNullable[LocalDate] and
```
paulbrown1982: should this be liveNewsAppSubscriptionExpiryDate - I'm thinking what would we add if we wanted to do the same for Daily Edition in-app purchases?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -61,22 +70,46 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
...
   private def lookup(endpointDescription: String, onSuccessMember: Attributes => Result, onSuccessSupporter: Attributes => Result, onNotFound: Result, sendAttributesIfNotFound: Boolean = false) = {
     AuthAndBackendViaAuthLibAction.async { implicit request =>
 
       if(endpointDescription == "membership" || endpointDescription == "features") {
         DeprecatedRequestLogger.logDeprecatedRequest(request)
       }
 
-      val userHasValidatedEmail = request.user.flatMap(_.statusFields.userEmailValidated).getOrElse(false)
-
       request.user.map(_.id) match {
```
paulbrown1982: Could you have this return a tuple of (user => (user.id, user.statusFields.userEmailValidated.getOrElse(false))) then you don't need to pass the full user into getLatestOneOffContributionDate, nor map over the user option twice?

alexduf: I see the issue, I pushed something slightly different to your suggestion but I believe it should solve the underlying point you were making.

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -61,54 +69,78 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
...
+        log.warn("Exception while fetching mobile subscription, assuming none", error)
+        Success(None)
+      case Success(-\/(error)) =>
+        log.warn(s"Unable to fetch mobile subscription, assuming none: $error")
+        Success(None)
+      case Success(\/-(status)) => Success(status)
+    }
+  }
+
+  def enrichZuoraAttributes(zuoraAttributes: Option[Attributes], latestOneOffDate: Option[LocalDate], mobileSubscriptionStatus: Option[MobileSubscriptionStatus]): Option[Attributes] = {
```
johnduffell: enrich doesn't really mean anything specific, could we call it withNonZuoraAttributes or something similar

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -61,54 +69,78 @@ class AttributeController(attributesFromZuora: AttributesFromZuora, commonAction
     else Future.successful(None)
   }
 
+  def getLatestMobileSubscription(identityId: String)(implicit executionContext: ExecutionContext): Future[Option[MobileSubscriptionStatus]] = {
+    mobileSubscriptionService.getSubscriptionStatusForUser(identityId).transform {
+      case Failure(NonFatal(error)) =>
+        log.warn("Exception while fetching mobile subscription, assuming none", error)
+        Success(None)
```
johnduffell: is it a problem is this stops working? if so how would we know? is it worth pushing a metric and alarming on it?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -18,9 +18,16 @@ import org.joda.time.LocalDate
...
+import scala.util.{Failure, Success, Try}
+import scala.util.control.NonFatal
+
+class AttributeController(
+  attributesFromZuora: AttributesFromZuora,
+  commonActions: CommonActions,
+  override val controllerComponents: ControllerComponents,
+  oneOffContributionDatabaseService: OneOffContributionDatabaseService,
+  mobileSubscriptionService: MobileSubscriptionService
+) extends BaseController with LoggingWithLogstashFields {
```
johnduffell: nice work laying that out neatly 🥇

membership-attribute-service/app/services/MobileSubscriptionService.scala
```diff
@@ -0,0 +1,42 @@
...
+class MobileSubscriptionServiceImpl(wsClient: WSClient)(implicit ec: ExecutionContext) extends MobileSubscriptionService {
+
+  override def getSubscriptionStatusForUser(identityId: String): Future[String \/ Option[MobileSubscriptionStatus]] = {
+    val response = wsClient.url(s"${Config.Mobile.subscriptionURL}/user/subscriptions/$identityId")
+      .withHttpHeaders("Authorization" -> s"Bearer ${Config.Mobile.subscriptionApiKey}")
+      .get()
+
+    response.map { resp =>
+      if (resp.status != 200) {
+        \/.left(s"Unable to fetch the mobile subscription status for $identityId")
```
johnduffell: would be useful to log the body so we know what was return by mobile api

johnduffell: some http libraries actually keep the inputstream lying around if you don't read it (okhttp) so it might be worth checking that

membership-attribute-service/app/configuration/Config.scala
```diff
@@ -52,4 +52,9 @@ object Config {
     val enabled = Try{config.getBoolean("logstash.enabled")}.toOption.contains(true)
   }
 
+  object Mobile {
+    val subscriptionURL = config.getConfig("mobile.subscription.base")
```
johnduffell: I mentioned already on chat, but any piece of config is a risk to prod as it is only checked at run time.  If the config is not a secret, then there's no reason not to put it in the source code as a pattern match or if statement.  This gets full static checking and can be reviewed and rolled back together with the changes.
The disadvantage is you have to know all your environments at compile time, but a PR to add an environment wouldn't be arduous for us.

alexduf: So just to confirm, you're okay with me breaking the existing pattern and have a if "PROD" in here?

membership-attribute-service/app/controllers/AttributeController.scala
alexduf: >is it a problem is this stops working
Yes
>if so how would we know?
We wouldn't

Should I remove the handling of the exception and let it bubble up?

membership-attribute-service/app/controllers/AttributeController.scala
alexduf: "Enrich" doesn't mean enough if you stop reading at the method name, but if you read the method + the parameters then I'm of the opinion it carries the intended meaning.

How strongly do you feel about this one?

membership-attribute-service/app/configuration/Config.scala
johnduffell: tl-dr: yes I'm ok with that. I think compile time safety > following a pattern

Personally I wouldn't even put it on that line, but at the relevant part of the code, but I appreciate a big refactor isn't what you want at this point.
For example the stripe base url is a default parameter https://github.com/guardian/support-frontend/blob/c47c47b2726a12f586b062738c97a405b0377b97/support-workers/src/main/scala/com/gu/stripe/StripeService.scala#L15

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: bubbling would cause a 5xx error and get no useful data at all. So we would know about it, but it would be more severe in terms of breaking swimlaning. I like that you've made an error barrier, I would just like an alert as well.

I'm not surprised if the existing code isn't too smart on this, and I'm happy to have alerting in a subsequent PR (it's not blocking this one)

My suggestion would be to push a metric to cloudwatch and add an alarm on that.
Here is a ready made implementation example for infrequent metrics: https://github.com/guardian/support-frontend/blob/master/support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala

membership-attribute-service/app/controllers/AttributeController.scala
johnduffell: if you've thought about it, it's enough for me!  😆 So no need to change it.

alexduf:
Ready to be merged on Monday ✅

twrichards:
Amazing stuff @alexduf - I've been looking forward to this for a while. I'd love to build it out some more in a few sprints time so it can show as subscription like any other in MMA (Manage My Account) area.

Is there a reason this hasn't yet been merged/deployed?

alexduf:
> Is there a reason this hasn't yet been merged/deployed?

Yes we're currently waiting for more data to be sent from the client side so we don't pay for the aws bill for nothing. Right now the main table is empty so it will always be useless.

PR #409, opened 2019-10-04
updating nginx scripts to use dev-nginx scripts

frj:
This PR contains changes to the scripts in /nginx which configure your local nginx instance to support running the app locally.

The scripts no longer worked as they referenced certificates in intentity-platform that no longer exist, as that project has been refactored to use the dev-nginx tool.

These changes refactor the scripts to create their own certificates and configure nginx using the dev-nginx tool.


frj:
@johnduffell good point probably not, I think the dev-nginx scripts put the config files in a different place which might cause conflict if the old version is also getting loaded.

I will add some cleanup for the old config.

PR #411, opened 2019-10-15
add more granular 'Voucher' and 'HomeDelivery' to the options for 'productType' query param

twrichards:
This `productType` query param was originally introduced to make the tab based groupings of products in `manage-frontend` easier. For some reason I thought it wise at the time to combine 'voucher' and 'home delivery' as a single product 'paper'... however NOW to support _Voucher holiday stops_ independently of _home delivery holiday stops_ I am now splitting 'paper' in `manage-frontend` (see https://github.com/guardian/manage-frontend/pull/302) and this change is therefore necessary.

**_It still supports `Paper` as a value and behaves the same - this is an additive change._**


twrichards:
Tested in CODE ✅ (no regression)

PR #412, opened 2019-10-21
Add a metric filter for a full http client queue

arausuy:
I'm adding an alarm on a metric filter, this provides us with a more refined metric and alarm. Currently we have a generic alarm on 400s - this frequently fires as a result of downstream bottlenecks.

### Why do we need this? 
It would be preferable to have this type of alarm as the current 400s alarm was put in tactically and the regular (and valid) 400s trigger the alarm causing initially a mild panic and now don't register with people.

### The changes

New metric filter on the log string `Max wait queue limit of 256 reached, not scheduling.` and an alarm with a low threshold as the condition is very different to having the app respond with 400s.
The threshold may need tweaking as time goes on.

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -283,6 +283,16 @@ Resources:
       - MetricValue: 1
         MetricNamespace: "members-data-api"
         MetricName: !Sub "${Stage} - Default Payment Method set to nothing"
+  MembersDataHttpQueuesFullMetricFilter:
+    Type: AWS::Logs::MetricFilter
+    DependsOn: MembersDataApiLogGroup
+    Properties:
+      FilterPattern: "\"Max wait queue limit of 256 reached, not scheduling.\""
+      LogGroupName: !Sub members-data-api-${Stage}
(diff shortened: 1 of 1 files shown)


cloudformation/membership-attribute-service.yaml
```diff
@@ -283,6 +283,16 @@ Resources:
       - MetricValue: 1
         MetricNamespace: "members-data-api"
         MetricName: !Sub "${Stage} - Default Payment Method set to nothing"
+  MembersDataHttpQueuesFullMetricFilter:
+    Type: AWS::Logs::MetricFilter
+    DependsOn: MembersDataApiLogGroup
+    Properties:
+      FilterPattern: "\"Max wait queue limit of 256 reached, not scheduling.\""
```
kelvin-chappell: Is this condition going to be a bit brittle, or is it the only feasible thing to measure?

arausuy: Currently it is the only message visible to us and only happens under extreme load. We have more work for further investigation - It is just more specific than monitoring a number of 4xx which happens on every busy news day

PR #414, opened 2019-11-28
Update Readme based on experience of getting set up

kelvin-chappell:
This just brings together the steps I had to take to get the service running on my machine, and it incorporates some links that were in other repos.


arausuy:
This is very handy - good stuff!

PR #415, opened 2019-12-03
Expose delivery address in subscription response

kelvin-chappell:
### Why do we need this? 
This change adds a `deliveryAddress` object beneath the `subscription` object in the response body of a `user-attributes/me/mma` request, so that a subscriber's delivery address can be populated in MMA.

### The changes 
The `AccountController.allCurrentSubscriptions` method now returns a `List[ContactAndSubscription]` instead of a `List[Subscription]` which allows us to populate a new `deliveryAddress` object in the output json using the mailing address fields from the contact.

TODO:
I have not considered the case where the buyer and recipient have different addresses yet, ie in the case of a gift.

* https://trello.com/c/GhHtFhg4
* https://trello.com/c/MZJqIC1H

The json will look something like this:
image

Tested in Code.

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -183,6 +183,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       alertText <- OptionEither.liftEitherOption(alertText(accountSummary, sub, getPaymentMethod))
       isAutoRenew = sub.autoRenew
     } yield AccountDetails(
+      contactId = contact.salesforceContactId,
       regNumber = contact.regNumber,
       email = accountSummary.billToContact.email,
       deliveryAddress = None,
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -222,21 +224,34 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
-      case NoFilter =>
-        subscriptions
-    }
-  } yield filteredIfApplicable
+      maybeUserId: Option[String],
+      filter: OptionalSubscriptionsFilter
+  ): OptionT[OptionEither.FutureEither, List[ContactAndSubscription]] = for {
+      user <- OptionEither.liftFutureEither(maybeUserId)
+      contact <- OptionEither(contactRepo.get(user))
+      subscriptions <-
```
frj: Could be contactAndSubscription for consistency

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -103,6 +104,7 @@ object AccountDetails {
           "joinDate" -> paymentDetails.startDate,
           "optIn" -> !paymentDetails.pendingCancellation,
           "subscription" -> (paymentMethod ++ Json.obj(
+            "deliveryAddress" -> accountDetails.deliveryAddress,
```
twrichards: please can we expose the Salesforce contact ID aswell as `contactID`? @rBangay can elaborate...

PR #416, opened 2019-12-06
Add endpoint for updating delivery addresses

kelvin-chappell:
### Why do we need this? 
This change allows us to save updated delivery addresses from MMA to Salesforce.

### The changes 
I've added a new route taking a contact ID in the URL and the updated delivery address as the json body of the request.  This is sent on to Salesforce as a patch request on the Contact record.

* https://trello.com/c/GhHtFhg4
* https://trello.com/c/kW8RPfW9

Depends on release 0.554 of membership-common, incorporating [this change](https://github.com/guardian/membership-common/pull/605).

### DIFF
--- a/membership-attribute-service/app/controllers/ContactController.scala
+++ b/membership-attribute-service/app/controllers/ContactController.scala
@@ -1,0 +1,56 @@
+package controllers
+
+import actions.CommonActions
+import com.gu.salesforce.SFContactId
+import models.DeliveryAddress
+import play.api.mvc.{Action, AnyContent, BaseController, ControllerComponents}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.control.Exception
+
+class ContactController(
+    commonActions: CommonActions,
+    override val controllerComponents: ControllerComponents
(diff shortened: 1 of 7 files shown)


membership-attribute-service/app/controllers/ContactController.scala
```diff
@@ -0,0 +1,56 @@
...
+            val mergedAddressLines = DeliveryAddress.mergeAddressLines(address)
+            Map() ++
+              contactField("MailingStreet", mergedAddressLines) ++
+              contactField("MailingCity", address.town) ++
+              contactField("MailingState", address.region) ++
+              contactField("MailingPostalCode", address.postcode) ++
+              contactField("MailingCountry", address.country)
+          }
+
+          val update = contactRepo.salesforce.Contact.update(SFContactId(contactId), contactFields)
```
frj: Is there any validation that the contact id matches the validated user, could this be a vulnerability?

kelvin-chappell: It's a good point.  I'll have a look at what we get from the user.

PR #424, opened 2020-02-20
add new '4XX Ratio' alarm

twrichards:
We had an [incident recently](https://docs.google.com/document/d/1JqPOrC0YGNJre6q6hx8gPzMd8R0qzWGqhoMjjybZj0U/edit). For most of the incident we were serving **majority** `4XX` responses.

This PR introduces a new alarm which uses **_'Metric Math'_** to to alert when the **ratio of `4XX` responses to `Total Requests` exceeds 20%** (which is indicative of a major problem - probably communicating with identity). 

The below screenshot shows this new calculated metric around the time of the incident, annotated with the the proposed alarm...
Screenshot 2020-02-20 at 17 45 32

There will be subsequent PRs for the other alarms improvements, just doing this one in isolation to keep the PR manageable. One of the subsequent PRs will be to adopt the naming convention of most of our other alarms, see https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -370,6 +370,47 @@ Resources:
           Period: 60
           Statistic: Sum
           Threshold: 1
+
+  4XXRatioAlarm:
+    Type: 'AWS::CloudWatch::Alarm'
+    Condition: CreateProdMonitoring
+    Properties:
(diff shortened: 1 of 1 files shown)


tomrf1:
👍 great work

PR #425, opened 2020-02-21
change 5XX Alarm to ELB (rather than backend) and codify NoHealthyInstancesAlarm

twrichards:
We had an [incident recently](https://docs.google.com/document/d/1JqPOrC0YGNJre6q6hx8gPzMd8R0qzWGqhoMjjybZj0U/edit).

First off neither the 'High 5XX' alarm or the 'No Healthy Instances' alarm were in the CloudFormation (manually created in AWS Console by the looks of it). So this PR adds those, with the same `Threshold`, `Period`, `EvaluationPeriods`, `ComparisonOperator` and `Statistic` as the existing.

**Crucially** though, the 5XX alarm now uses  the **SUM** of `HTTPCode_ELB_5XX` and `HTTPCode_Backend_5XX` because during the incident the backend was unresponsive so the ELB was serving `504 Gateway Timeout` to consumers/clients, but no alarm fired because it was based on `HTTPCode_Backend_5XX` count and the backend was unresponsive (OutOfMemory). See...
image

There will be subsequent PRs for the other alarms improvements, just doing this one in isolation to keep the PR manageable. One of the subsequent PRs will be to adopt the naming convention of most of our other alarms, see https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -411,6 +411,68 @@ Resources:
             Unit: Count
           ReturnData: false
 
+  NoHealthyInstancesAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdMonitoring
+    Properties:
+      AlarmActions:
+        - !Sub 'arn:aws:sns:${AWS::Region}:${AWS::AccountId}:fulfilment-dev'
+      AlarmName: !Sub No healthy instances for members-data-api in ${Stage}
+      MetricName: HealthyHostCount
+      Namespace: AWS/ELB
+      Dimensions:
+          - Name: LoadBalancerName
+            Value: !Ref LoadBalancer
+      ComparisonOperator: LessThanOrEqualToThreshold
+      Threshold: 0.5
+      Period: 60
+      EvaluationPeriods: 10
+      Statistic: Average
+    DependsOn:
+      - LoadBalancer
+
+  High5XXRateAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdMonitoring
+    Properties:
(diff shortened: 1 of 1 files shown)


mario-galic:
What is the difference between `HTTPCode_Target_5XX_Count` and `HTTPCode_Backend_5XX`? It seems they are the same metric but for different load balancer versions (Application vs Classic).

twrichards:
> What is the difference between `HTTPCode_Target_5XX_Count` and `HTTPCode_Backend_5XX`? It seems they are the same metric but for different load balancer versions (Application vs Classic).
> 
> **HTTPCode_Target_5XX_Count**
> 
> > The number of HTTP response codes generated by the targets. This does not include any response codes generated by the load balancer.
> 
> **HTTPCode_Backend_5XX**
> 
> > The number of HTTP response codes generated by registered instances. This count does not include any response codes generated by the load balancer.

@mario-galic great catch, I had intended to change it to `HTTPCode_ELB_5XX` but bad copy/pasta 🙈. Updated now.

mario-galic:
AFAIK, `HTTPCode_ELB_5XX` means 

- ELB cannot forward request to instance, or
- instance returned un-parsable nonsense back to ELB

however it does **NOT** mean instance responded with 5xx. @adamnfish @sihil @jacobwinch can you confirm this? In other words, instance can return 5xx but ELB metric will not count it. Thus alarming on `HTTPCode_ELB_5XX` is insufficient, and we should alarm on the **SUM** of `HTTPCode_Backend_5XX + HTTPCode_ELB_5XX` as Tom has done in this PR?

PR #431, opened 2020-03-19
enable digipack access for paper customers (voucher & HD)

twrichards:
...because Coronavirus

image


twrichards:
TESTED in `CODE` ✅ 
image
The above user just has a voucher sub 🙂

PR #433, opened 2020-03-30
un-invoiced subs should be cancelled effective today

twrichards:
_Utilising the new functionality in https://github.com/guardian/membership-common/pull/607, we can now choose to cancel a sub 'effective today'._

Un-invoiced subs would error when cancelling, because the default was to cancel effective `EndOfLastInvoicePeriod`... makes sense.

To **facilitate cancelling digipacks during trial period** (i.e. un-invoiced) we now cancel 'effective today' - which works nicely 🙂

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -8,7 +8,7 @@ import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.memsub.subsv2.services.SubscriptionService
-import com.gu.memsub.subsv2.{PaidChargeList, Subscription, SubscriptionPlan}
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -98,7 +98,11 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       val cancellationSteps = for {
         _ <- EitherT(tp.zuoraRestService.disableAutoPay(zuoraSubscription.accountId)).leftMap(message => s"Error while trying to disable AutoPay: $message")
         _ <- EitherT(tp.zuoraRestService.updateCancellationReason(zuoraSubscription.name, reason)).leftMap(message => s"Error while updating cancellation reason: $message")
-        cancelResult <- EitherT(tp.zuoraRestService.cancelSubscription(zuoraSubscription.name)).leftMap(message => s"Error while cancelling subscription: $message")
+        subHasBeenInvoiced = zuoraSubscription.plans.list.exists{
```
twrichards: I welcome any suggestions on how I could invert this, such that the variable could be `subHasNotBeenInvoiced` so there is no negation on line 105

PR #439, opened 2020-06-04
Remove dead dependency postgres

mario-galic:


twrichards:
@tomrf1 so how does it connect to the contributions-store if not via this dependency?

tomrf1:
I expect it's something Play provides

PR #441, opened 2020-06-05
Measure cache hit versus zuora direct hit

mario-galic:
Cache here is `SupporterAttributesFallback` dynamodb table.


mario-galic:
> Perhaps in the Zuora service, so it also captures the multitude of calls made on the more detailed mma endpoint
Yes I will do it in separate PR. There seems to be pre-existing code I could use for that in https://github.com/guardian/membership-common/blob/0a52c2ce11a7f49491799a0930e1fb7ad57ed9dc/src/main/scala/com/gu/okhttp/RequestRunners.scala#L48

mario-galic:
[Measure outbound requests to Zuora](https://trello.com/c/LuEH1lJT/1384-measure-outbound-requests-to-zuora)

PR #443, opened 2020-06-09
Measure total requests made by zuora rest client

mario-galic:
Related PR: https://github.com/guardian/membership-common/pull/614


mario-galic:
Zuora is being hit at least approx 900 per second via REST client

image

PR #448, opened 2020-06-11
Inline logic to fetch attributes from zuora or cache

mario-galic:
No changes in behaviour in this PR. Just trying to make the logic easer to follow by having a single inlined for comprehension.

It seems the overall Zuora concurrency limit handling logic is as follows

1. Count Zuora concurrent requests 
1. Get the concurrency limit set in `AttributesFromZuoraLookup` dynamodb table
1. If the count is greater than limit, then hit cache
1. If the count is less than limit and Zuora is healthy, then hit Zuora
1. If the count is less than limit and Zuora is unhealthy, then hit cache

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -40,8 +40,8 @@ class AttributeController(
    * 1. Count Zuora concurrent requests
    * 1. Get the concurrency limit set in `AttributesFromZuoraLookup` dynamodb table
    * 1. If the count is greater than limit, then hit cache
-   * 1. If the count is less than limit, then hit Zuora if Zuora is healthy
-   * 1. If the count is less than limit, then hit cache if Zuora is unhealthy
+   * 1. If the count is less than limit and Zuora is healthy, then hit Zuora
+   * 1. If the count is less than limit and Zuora is unhealthy, then hit cache
    */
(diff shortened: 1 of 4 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -33,14 +33,24 @@ class AttributeController(
...
 
-  def pickAttributes(identityId: String) (implicit request: AuthenticatedUserAndBackendRequest[AnyContent]): Future[(String, Option[Attributes])] = {
+  /**
+   * Zuora enforces 40 concurrent requests limit, without providing caching mechanisms.
+   * So the following custom workaround logic is attempted:
+   *
+   * 1. Count Zuora concurrent requests
+   * 1. Get the concurrency limit set in `AttributesFromZuoraLookup` dynamodb table
+   * 1. If the count is greater than limit, then hit cache
+   * 1. If the count is less than limit, then hit Zuora if Zuora is healthy
```
kelvin-chappell, line 43:
```suggestion
   * 1. If the count is less than limit and Zuora is healthy, then hit Zuora
```

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -33,14 +33,24 @@ class AttributeController(
...
+   * 1. If the count is less than limit, then hit cache if Zuora is unhealthy
```
kelvin-chappell, line 44:
```suggestion
   * 1. If the count is less than limit and Zuora is unhealthy, then hit cache
```

PR #450, opened 2020-06-11
Fall back to cache on Zuora networking outage

mario-galic:
## Background

Failed future is not the same as future of failure. Consider

```
Future { new Exception("boom") } andThen { case Success(value) => println("success") } 
Future { None                  } andThen { case Success(value) => println("success") }
Future { Left("boom")          } andThen { case Success(value) => println("success") }
```

None of these is a failed future, that is, `success` is printed in all cases. Only **throw expression** fails the future

```
Future(throw new Exception("boom"))
```

Networking issue usually manifests itself as a failed future because something throws, say, `SocketTimeout`, however, server responding with 500 does not necessarily mean failed future because server did indeed respond correctly at networking level.

## Root cause

The fact that Play had to handle uncaught exception means that fallback mechanisms on cache never executed in case of networking failure.

```
ERROR: Error handling request request: GET /user-attributes/me
members-data-api-PROD 2020-06-11/i-00b59b71c324e9a17/membership-attribute-service.log play.api.UnexpectedException: Unexpected exception[SocketTimeoutException: timeout]
	at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:340)
	at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:263)
	at play.core.server.AkkaHttpServer$$anonfun$1.applyOrElse(AkkaHttpServer.scala:443)
members-data-api-PROD 2020-06-11/i-00b59b71c324e9a17/membership-attribute-service.log Caused by: java.net.SocketTimeoutException: timeout
	at okio.Okio$3.newTimeoutException(Okio.java:210)
	at okio.AsyncTimeout.exit(AsyncTimeout.java:288)
	at okio.AsyncTimeout$2.read(AsyncTimeout.java:242)
```

It is difficult to see what is actually going on due to all the indirection, but there is a 

- custom Scala async wrapper over Java okhttp client https://github.com/guardian/membership-common/blob/401f38f5c5de418429b03e0e94982e17a547db2a/src/main/scala/com/gu/lib/okhttpscala/package.scala
- and then another layer of parsing https://github.com/guardian/membership-common/blob/master/src/main/scala/com/gu/zuora/rest/SimpleClient.scala
- And then bunch of layers on top of that, all spiced up with Scalaz

That `SimpleClient` layer seems to move just parsing errors to left, but does not move networking errors to the left.

## Solution

So the fix here is

```
.recoverWith { case e => fallbackToCache(e.toString) }`
```

which actually handles networking failure, whilst

```
.leftMap(fallbackToCache)
```

handles `Future` with success value but which happens to be a `Left`.

### DIFF
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -50,6 +50,11 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
           case Failure(e) => SafeLogger.error(scrub"Failed to retrieve attributes from cache for $identityId because $e")
         }
 
+    lazy val fallbackToCache = (zuoraError: String) => {
+      SafeLogger.error(scrub"Failed to retrieve attributes from Zuora for $identityId because $zuoraError so falling back on cache.")
+      attributesFromDynamo.map(maybeDynamoAttributes => ("Dynamo", maybeDynamoAttributes.map(DynamoAttributes.asAttributes(_))))
+    }
+
     (for { // try zuora first and fallback to cache
       accounts         <- EitherT(identityIdToAccounts(identityId))
       subscriptions    <- EitherT(if (accounts.size > 0) getSubscriptions(accounts, subscriptionsForAccountId) else Future.successful(\/.right[String, List[AccountWithSubscriptions]](Nil)))
@@ -58,10 +63,11 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
       val maybeAttributes = maybeZAttributes.map(ZuoraAttributes.asAttributes(_))
       attributesFromDynamo.foreach(maybeUpdateCache(maybeZAttributes, _, maybeAttributes))
       Future.successful("Zuora", maybeAttributes)
-    }).leftMap { zuoraError => // fallback to Dynamo if Zuora fails
-      SafeLogger.error(scrub"Failed to retrieve attributes from Zuora for $identityId because $zuoraError so falling back on cache.")
-      attributesFromDynamo.map(maybeDynamoAttributes => ("Dynamo", maybeDynamoAttributes.map(DynamoAttributes.asAttributes(_))))
-    }.merge.flatten // both branches return same type so can be merged
+    }) // fallback to cache if Zuora networking error or un-parsable response
+      .leftMap(fallbackToCache) // handles un-parsable response etc.
+      .merge // both branches return the same type so we can merge
+      .flatten
+      .recoverWith { case e => fallbackToCache(e.toString) } // handles timeouts etc.
   }
 
   def dynamoUpdateRequired(dynamoAttributes: Option[DynamoAttributes], zuoraAttributes: Option[ZuoraAttributes], identityId: String, twoWeekExpiry: => DateTime): Boolean = {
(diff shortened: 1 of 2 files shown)


mario-galic:
[Make sure cache fallback mechanism works](https://trello.com/c/FyKmT081/1402-make-sure-cache-fallback-mechanism-works)

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -58,10 +63,11 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
...
       Future.successful("Zuora", maybeAttributes)
-    }).leftMap { zuoraError => // fallback to Dynamo if Zuora fails
-      SafeLogger.error(scrub"Failed to retrieve attributes from Zuora for $identityId because $zuoraError so falling back on cache.")
-      attributesFromDynamo.map(maybeDynamoAttributes => ("Dynamo", maybeDynamoAttributes.map(DynamoAttributes.asAttributes(_))))
-    }.merge.flatten // both branches return same type so can be merged
+    }) // fallback to cache if Zuora networking error or un-parsable response
+      .leftMap(fallbackToCache) // handles un-parsable response etc.
+      .merge // both branches return the same type so we can merge
+      .flatten
+      .recoverWith { case e => fallbackToCache(e.toString) } // handles timeouts etc.
```
arausuy, line 70: Worth putting a test to ensure we don't have this mistake again here?

mario-galic: I'll give it a go. It might be fiddly.

mario-galic: I've added the test. Also manually traced with debugger to make sure fallback executes. I can see now that without this fix exception was indeed thrown which would then be caught by Play's top level error handler which converts it to 500, however fallback would not be executed.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -58,10 +63,11 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext) exten
       val maybeAttributes = maybeZAttributes.map(ZuoraAttributes.asAttributes(_))
       attributesFromDynamo.foreach(maybeUpdateCache(maybeZAttributes, _, maybeAttributes))
       Future.successful("Zuora", maybeAttributes)
-    }).leftMap { zuoraError => // fallback to Dynamo if Zuora fails
-      SafeLogger.error(scrub"Failed to retrieve attributes from Zuora for $identityId because $zuoraError so falling back on cache.")
-      attributesFromDynamo.map(maybeDynamoAttributes => ("Dynamo", maybeDynamoAttributes.map(DynamoAttributes.asAttributes(_))))
-    }.merge.flatten // both branches return same type so can be merged
+    }) // fallback to cache if Zuora networking error or un-parsable response
+      .leftMap(fallbackToCache) // handles un-parsable response etc.
+      .merge // both branches return the same type so we can merge
```
johnduffell, line 68: nice comment 😂

johnduffell: PS I realise it was already there

mario-galic: My thinking was `merge` might be unfamiliar so try to be explicit about it.

mario-galic:
Here is cache successfully reacting to an actual PROD networking problem event of Zuora `SocketTimeoutException` kind

```
 ERROR: Failed to retrieve attributes from Zuora for ***** because java.net.SocketTimeoutException: timeout so falling back on cache
```

image

PR #451, opened 2020-06-12
Measure failure to serve entitlements from /user-attributes/me

mario-galic:
Currently we serve 5xx approx 250 times per day

image

but we do not really know how many of those translate to user not receiving their paid-for entitlements so this PR adds `attributes-failed-to-serve-entitlements` metric.

Previously there was no explicit handing of 5xx on `/user-attributes/me`, see https://github.com/guardian/members-data-api/pull/450

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -120,7 +120,7 @@ class AttributeController(
           val futureOneOffContribution = getLatestOneOffContributionDate(user.id, user)
           val futureMobileSubscriptionStatus = getLatestMobileSubscription(user.id)
 
-          for {
+          ((for {
             //Fetch one-off data independently of zuora data so that we can handle users with no zuora record
             (fromWhere: String, zuoraAttributes: Option[Attributes]) <- futureAttributes
             latestOneOffDate: Option[LocalDate] <- futureOneOffContribution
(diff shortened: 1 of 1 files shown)


mario-galic:
[Make sure cache fallback mechanism works](https://trello.com/c/FyKmT081/1402-make-sure-cache-fallback-mechanism-works)

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -154,11 +154,19 @@ class AttributeController(
...
-                enrichZuoraAttributes(Attributes(user.id), latestOneOffDate, latestMobileSubscription)
+                enrichZuoraAttributes(Attributes(user.id), latestOneOffDate, latestMobileSubscription) // FIXME: This seem like ti should be returning Result not Attributes
               case _ =>
                 onNotFound
             }
-          }
+          }): Future[Result]) /* FIXME: This type annotation should not be necessary. It probably means there is bug above */
+            .recover { case e =>
+              // This branch indicates a serious error to be investigated ASAP, because it likely means we could not
+              // serve from either Zuora or DynamoDB cache. Likely multi-system outage in progress or logic error.
```
arausuy, line 164: I think there is a case where you can be a new acquisition so you won't be in the cache yet and then Zuora doesn't respond in time so you don't get anything served

mario-galic: I think currently that will respond with 404, because if zuora times out but dynamo does not, then future will still be something like `Future.success(None)`.

mario-galic: I think this scenario could be addressed in many cases with a retry of Zuora call. I could put the retry logic in the Zuora http client in membership-common. What do you think?

mario-galic: @arausuy https://github.com/guardian/members-data-api/pull/453

PR #452, opened 2020-06-12
How do we handle Zuora 40 concurrent requests limit?

mario-galic:


README.md
```diff
@@ -3,6 +3,52 @@
...
+There are few ways we try to manage the load 
+1. dotcom cookies that expire after 24 hours
+1. mem-data-api DynamoDB table with particular TTL
+1. mechanism to control number of concurrent Zuora requests each mem-data-api instance can make (currently there are 6)
+
+With current load management we hit Zuora around 1000 per minute.
+
+### Dotcom cookies caching
+
+TODO: Tom to explain. Maybe list the names of cookies with example of their content
```
twrichards, line 23: will do

README.md
```diff
@@ -3,6 +3,52 @@
...
+  hit cache
+```
+
+Effect of different values for `ConcurrentZuoraCallThreshold`
+- 1 results in about 50/50 split between Zuora and cache
+- 2 results in about 80/20
+- 7 results in likely limit hits because (6 instances) x (7 concurrent reqests) = 42
+
+**WARNING: Remember to reduce `ConcurrentZuoraCallThreshold` if instances need to scale, say in expectation of 
+drastic increase of load due to breaking news.**
```
twrichards, line 50: in that we have AWS Integration already (via SDK, for the likes of Metrics etc.) could we perhaps **lookup the number of instances currently active** and divide by `ConcurrentZuoraCallThreshold` (having re-purposed that variable to be the _total_ we're happy for `members-data-api` to use, e.g. 30, which means under normal circumstances means 5 per instance, but if we scaled up then it might mean say 3 per instance)

README.md
```diff
@@ -3,6 +3,52 @@
 The members' data API is a Play app that manages and retrieves supporter attributes associated with a user.  
 It runs on https://members-data-api.theguardian.com/.
 
+
+## How do we handle Zuora 40 concurrent requests limit?
```
johnduffell, line 7: actually on 16th april the API concurrency limits on your tenant have been doubled.  So we actually have 80 now!

README.md
```diff
@@ -3,6 +3,52 @@
...
+
+TODO: Tom to explain. Maybe list the names of cookies with example of their content
+
+### DynamoDB table
+
+1. Count Zuora concurrent requests (per instance)
+1. Get the concurrency limit set in `AttributesFromZuoraLookup` dynamodb table
+1. If the count is greater than limit, then hit cache
+1. If the count is less than limit and Zuora is healthy, then hit Zuora
+1. If the count is less than limit and Zuora is unhealthy, then hit cache
```
johnduffell, line 31:
```suggestion
1. If Zuora is unhealthy, then hit cache
```

README.md
twrichards: they were indeed doubled, but doubled from 20 back to 40
somewhere in the last few years they had been reduced to 20 without us knowing

README.md
twrichards: @johnduffell your suggestion is less explicit

README.md
johnduffell: 🤦 so they took away with one hand and gave back with the other...

README.md
johnduffell: there's no difference, although I realise the code doesn't work the way I wrote.  for clarity I'd probably write it with indentation a bit like the pseudocode below

README.md
twrichards: done in https://github.com/guardian/members-data-api/commit/de517f2d81ffd582f598b35e78ce7f7dd835b4b6

PR #456, opened 2020-06-16
Automatically calculate zuora concurrency threshold by instance count

mario-galic:
https://trello.com/c/S3wmxF0D/1408-automatically-adjust-zuora-concurrency-limit-when-scaling-server-instances

Address https://github.com/guardian/members-data-api/pull/452#discussion_r439481083


README.md
```diff
@@ -51,16 +52,16 @@ Various things from the `/user-attributes/me` response are stored in cookies, to
...
 else
   hit cache
 ```
 
 Effect of different values for `ConcurrentZuoraCallThreshold`
-- 1 results in about 50/50 split between Zuora and cache
-- 2 results in about 80/20
-- 7 results in likely limit hits because (6 instances) x (7 concurrent reqests) = 42
+- 6 total across 6 instances results in about 50/50 split between Zuora and cache
+- 12 total across 6 instances results in about 80/20
```
twrichards, line 63:
```suggestion
- 6 total across 6 instances (i.e. limit of one per instance) results in about 50/50 split between Zuora and cache
- 12 total across 6 instances (i.e. limit of two per instance) results in about 80/20
```

membership-attribute-service/app/prodtest/InstanceCountOnSchedule.scala
```diff
@@ -0,0 +1,44 @@
...
+      .build()
+
+  private def getCurrentNumberOfInstances(stage: String): Int =
+    autoScalingClient
+      .describeAutoScalingGroups()
+      .getAutoScalingGroups
+      .asScala
+      .toList
+      .filter(_.getAutoScalingGroupName.startsWith(s"Memb-Attributes-$stage"))
+      .map(_.getInstances.asScala.size)
```
twrichards, line 29: does the `getInstances` need to be filtered on those which are _active/healthy_ (to prevent it spiking during a release - since the number of instances will double for a while)

twrichards: if that wouldn't help - perhaps instead we should be asking the load balancer how many instances it is sending traffic to?

mario-galic:
This seems to have worked

image

It is set now to 12 concurrent Zuora requests across all instances which gives it approx 80/20 Zuora/Cache split

image

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -69,6 +67,25 @@ class AttributeController(
...
   }
 
+  private def calculateZuoraConcurrencyLimitPerInstance(implicit request: AuthenticatedUserAndBackendRequest[AnyContent]): Int = {
+    val featureToggleData = request.touchpoint.featureToggleData.getZuoraLookupFeatureDataTask.get()
+    val totalConcurrentCallThreshold = featureToggleData.ConcurrentZuoraCallThreshold
+    val awsInstanceCount = request.touchpoint.instanceCountOnSchedule.getInstanceCountTask.get()
+
+    val cappedTotalConcurrentCallThreshold =
+      if (totalConcurrentCallThreshold > 40) {
+        log.error("Total Zuora concurrent requests limit set too high. Capping it to 40...")
```
johnduffell, line 77: good idea!

johnduffell:
nice work, great idea, much simpler to use now 👍

PR #464, opened 2020-06-23
Publish high frequency zuora count metrics on schedule

mario-galic:
Designed for high-frequency metrics, for example, 1000 per minute is about $400 per month. Uses `ConcurrentHashMap[String, AtomicInteger]` to count concurrent requests in thread-safe manner and periodically publishes the sum.

### DIFF
--- a/membership-attribute-service/app/monitoring/Metrics.scala
+++ b/membership-attribute-service/app/monitoring/Metrics.scala
@@ -1,9 +1,74 @@
 package monitoring
 
+import java.util.concurrent.ConcurrentHashMap
+
+import akka.actor.ActorSystem
+import com.amazonaws.services.cloudwatch.model.{MetricDatum, PutMetricDataRequest, StatisticSet}
 import com.gu.monitoring.CloudWatch
+import com.gu.monitoring.CloudWatch.cloudwatch
+import com.typesafe.scalalogging.LazyLogging
 import configuration.Config
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
 
 case class Metrics(service: String) extends CloudWatch {
   val stage = Config.stage
   val application = Config.applicationName // This sets the namespace for Custom Metrics in AWS (see CloudWatch)
+}
+
+// Designed for high-frequency metrics, for example, 1000 per minute is about $450 per month
+final class ExpensiveMetrics(
+  val service: String
+)(implicit system: ActorSystem, ec: ExecutionContext) extends CloudWatch with LazyLogging {
+  logger.info("expensive - created instance")
+
+  val stage = Config.stage
+  val application = Config.applicationName
+
+  system.scheduler.schedule(0.seconds, 55.seconds)(publishAllMetrics())
+
+  import scala.collection.JavaConverters._
+  private val chm = new ConcurrentHashMap[String, Int]().asScala
+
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/monitoring/Metrics.scala
```diff
@@ -1,9 +1,40 @@
...
+
+  def countRequest(key: String): Unit =
+    chm.getOrElseUpdate(key, new AtomicInteger(1)).incrementAndGet()
+
+  private def resetCount(key: String): Unit =
+    chm.getOrElseUpdate(key, new AtomicInteger(0)).set(0)
+
+  private def publishAllMetrics(): Unit =
+    chm foreach { case (key, value) =>
+      put(key, value.doubleValue())
```
twrichards, line 37: Why `doubleValue()` and not `intValue()`?

membership-attribute-service/app/monitoring/Metrics.scala
```diff
@@ -1,9 +1,40 @@
...
+  import scala.collection.JavaConverters._
+  private val chm = new ConcurrentHashMap[String, AtomicInteger]().asScala // keep it first in the constructor
+
+  val stage = Config.stage
+  val application = Config.applicationName
+
+  system.scheduler.schedule(5.seconds, 60.seconds)(publishAllMetrics())
+
+  def countRequest(key: String): Unit =
+    chm.getOrElseUpdate(key, new AtomicInteger(1)).incrementAndGet()
```
twrichards, line 30: https://42ways.de/blog/2015/02/21/scala-pitfalls-concurrent-hashmap/

mario-galic: Not sure this is relevant anymore since Scala 2.12: https://github.com/scala/scala/pull/4323 I think implementation of [concurrent.Map#getOrElseUpdate](https://github.com/scala/scala/blob/cd8410d30427747bbcd1c94429ce1509e527a315/src/library/scala/collection/concurrent/Map.scala#L94) uses `putIfAbsent` which documentation states is [atomic](https://docs.scala-lang.org/overviews/collections/concrete-mutable-collection-classes.html#concurrent-maps).

membership-attribute-service/app/monitoring/Metrics.scala
mario-galic: membership-common requires it https://github.com/guardian/membership-common/blob/47b96d64c76caae1fddcfe0c3e0c0f64ef19202b/src/main/scala/com/gu/monitoring/CloudWatch.scala#L35

mario-galic:
This seems to have worked 

image

mario-galic:
No, the issue was initialisation order which was fixed by https://github.com/guardian/members-data-api/pull/464/commits/b9fa6835d96c4f6b91156e316f979f87fa11266c 

Because 

```
system.scheduler.schedule(0.seconds, 55.seconds)(publishAllMetrics())
```

was first in the constructor and delay was `0s`,  it called `publishAllMetrics` immediately however at that point `ConcurrentHashMap` might not have been initialised yet which caused NPE. The way I was greping logs and the fact that some instance were still reporting samples made me miss the NPE. I was able to hunt down the issue after the pairing session when we noticed only 2 instances are reporting samples.

PR #470, opened 2020-08-04
Add Zuora accountId to detailed me/mma response

mario-galic:
### Why do we need this?

This enables manage-frontend to send to Invoicing API an `accountId` to retrieve all invoices associated with the account. 

### The changes 

The field is added at the level of (Salesforce) `contactId` for example

```
[
  {
    "tier": "Guardian Weekly - Domestic",
    ...
    "subscription": {
      "contactId": "0033111111111QAP",
      "accountId": "2c9211111111111111111e3d"
...
```

### trello card

https://trello.com/c/ggqXe1rg/1492-goal-expose-invoices-for-the-billing-tab

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -198,7 +198,8 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       accountHasMissedRecentPayments = false,
       safeToUpdatePaymentMethod = true,
       isAutoRenew = isAutoRenew,
-      alertText = alertText
+      alertText = alertText,
+      accountId = accountSummary.id.get
     ).toJson).run.run.map {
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -198,7 +198,8 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       accountHasMissedRecentPayments = false,
       safeToUpdatePaymentMethod = true,
       isAutoRenew = isAutoRenew,
-      alertText = alertText
+      alertText = alertText,
+      accountId = accountSummary.id.get
```
mario-galic, line 202: This is for some reason defined as `case class AccountId(get: String) extends AnyVal` so that `get` is safe.

PR #471, opened 2020-09-19
Fix contributions cancellation effective date to be end of last invoice

mario-galic:
https://trello.com/c/IIlIyWzB/1636-self-service-cancellation-effective-date-bug

- Related membership-common PR: https://github.com/guardian/membership-common/pull/628
- The error will manifest for any recurring contributor who changes the amount they contribute, and then try to self-service cancellation. After the amount changes, fetching the latest version of Subscription in Zuora does not contain the "End of last invoice date" (internally called chargeThroughDate) and the current implementation defaults to "now" as the cancellation effective date as opposed to at the end of last invoice date
- Request to salesforce is kept to confirm user owns the subscription, however subscriptions is then retrieved from Zuora directly. (In the future we could perhaps confirm ownership using identityId directly in Zuora).

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -63,7 +63,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
   import commonActions._
   implicit val executionContext: ExecutionContext = controllerComponents.executionContext
 
-  def cancelSubscription[P <: SubscriptionPlan.AnyPlan : SubPlanReads](subscriptionNameOption: Option[memsub.Subscription.Name]) =
+  def cancelSubscription[P <: SubscriptionPlan.AnyPlan : SubPlanReads](subscriptionName: memsub.Subscription.Name) =
     AuthAndBackendViaAuthLibAction.async { implicit request =>
     val tp = request.touchpoint
     val cancelForm = Form { single("reason" -> nonEmptyText) }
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
   val awsDynamo = "com.amazonaws" % "aws-java-sdk-dynamodb" % awsClientVersion
   val awsSQS = "com.amazonaws" % "aws-java-sdk-sqs" % awsClientVersion
   val awsCloudWatch = "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsClientVersion
-  val membershipCommon = "com.gu" %% "membership-common" % "0.579"
+  val membershipCommon = "com.gu" %% "membership-common" % "0.1-SNAPSHOT"
```
kelvin-chappell, line 20: This will need updating

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -81,8 +81,10 @@ class AccountController(commonActions: CommonActions, override val controllerCom
     def retrieveZuoraSubscription(user: String): Future[ApiError \/ Subscription[P]] = {
       val getSubscriptionData = for {
         sfUser <- EitherT(tp.contactRepo.get(user).map(_.flatMap(_ \/> s"No Salesforce user: $user")))
-        zuoraSubscription <- EitherT(tp.subService.current[P](sfUser).map(subscriptionSelector(subscriptionNameOption, s"Salesforce user $sfUser")))
-      } yield zuoraSubscription
+        validationSub <- EitherT(tp.subService.current[P](sfUser).map(subscriptionSelector(Some(subscriptionName), s"Salesforce user $sfUser")))
+        zuoraCurrentSub <- OptionT(tp.subService.get[P](subscriptionName, isActiveToday = true)).toRight("notFound")
+        _ <- EitherT(Future.successful(if (validationSub.name == zuoraCurrentSub.name) \/-(zuoraCurrentSub) else -\/("Sub does not belong to Salesforce user")))
+      } yield zuoraCurrentSub
```
kelvin-chappell, line 87: This scalaz/cats code could be difficult to make sense of.  Are there existing tests for this code?  Although I think the difficulty is in parsing the types rather than the logic of it.

kelvin-chappell (approved):
👍

membership-attribute-service/app/controllers/AccountController.scala
mario-galic: Attempted to clarify logic in https://github.com/guardian/members-data-api/pull/471/commits/dcad758ac7c83b8719d35c2e129780a9066192b9 Regarding types, not much we can do without major refactoring in membership-common and/or members-data-api to have the types align.

mario-galic:
Tested in CODE with DEV touchpoint backend on A-S00078162

PR #476, opened 2020-10-02
Differentiate anniversary from termEndDate

mario-galic:
### Why do we need this?

MMA has a concept of annual holiday limits, so we need to be able to model anniversary correctly. `termEndDate` is not sufficient because some subscriptions currently could have it many years in the future.

### The changes 

Adds `anniversaryDate` field to `me/mma` response

      "nextPaymentDate": "2020-09-19",
      "start": "2020-09-19",
      "renewalDate": "2022-09-11",
      "anniversaryDate": "2021-09-19",

https://trello.com/c/PlKMGmbK/1652-unable-to-apply-holiday-stops-via-mma-if-termstartdate-is-far-enough-in-the-future

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -196,11 +196,10 @@ object AccountDetails {
     start: LocalDate,
     today: LocalDate = LocalDate.now()
   ): LocalDate = {
-    @tailrec def loop(left: LocalDate): LocalDate = {
-      val right = left.plusYears(1)
-      if (today.isEqual(left)) right
-      else if (today.isAfter(left) && today.isBefore(right)) right
-      else loop(right)
+    @tailrec def loop(current: LocalDate): LocalDate = {
+      val next = current.plusYears(1)
+      if (today.isBefore(next)) next
+      else loop(next)
     }
     loop(start)
   }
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -180,4 +181,27 @@ object AccountDetails {
...
+    start: LocalDate,
+    today: LocalDate = LocalDate.now()
+  ): LocalDate = {
+    @tailrec def loop(left: LocalDate): LocalDate = {
+      val right = left.plusYears(1)
+      if (today.isEqual(left)) right
+      else if (today.isAfter(left) && today.isBefore(right)) right
+      else loop(right)
+    }
+    loop(start)
```
twrichards, line 205: Is there a reason this wouldn't be sufficient?
```suggestion
    @tailrec def loop(current: LocalDate): LocalDate = {
      val next = current.plusYears(1)
      if (today.isBefore(next)) next
      else loop(next)
    }
    loop(start)
```

membership-attribute-service/test/models/AnniversaryDateTest.scala
```diff
@@ -0,0 +1,22 @@
+package models
+
+import org.joda.time.LocalDate
+import org.specs2.mutable.Specification
+
+class AnniversaryDateTest extends Specification {
+  "anniversaryDate" should {
+    "if today is equal left then right" in {
```
twrichards, line 8: again left/right confused me a little, could they be closer to what those dates mean?

twrichards (approved):

PR #477, opened 2020-10-12
Recognise Digital Subscription gift accounts

rupertbates:
In order to recognise Digital Subscription gift subs in members-data-api we need to query Zuora for subs which are active and have the custom field GifteeIdentityId set to the id of the current user.

We also need to ensure that the purchaser of the gift is not interpreted as having a digital subscription. 

As background the way that Digital subscription gifts are represented in Zuora is as a regular gift sub with 
- `readerType = Gift` 
- `IdentityId` custom field set to the identity id of the purchaser
- `GifteeIdentityId` custom field set to the identity id of the person who has redeemed the gift code.

More details [here](https://docs.google.com/document/d/1tey3Px090ZaBwwOOvlCK6lbPl7MW_3MCJjqYer3NSeY/edit#)

This PR depends on https://github.com/guardian/membership-common/pull/631

### [Trello card](https://trello.com/c/25rECTtP/3355-update-members-data-api-to-recognise-gift-subs-80)

I have now tested this end to end by creating a new Digital Subscription gift, redeeming it and then viewing the members-data-api output which was
```json
// 20201013134252
// https://members-data-api.thegulocal.com/user-attributes/me

{
  "userId": "200003991",
  "digitalSubscriptionExpiryDate": "2021-10-14",
  "showSupportMessaging": false,
  "contentAccess": {
    "member": false,
    "paidMember": false,
    "recurringContributor": false,
    "digitalPack": true,
    "paperSubscriber": false,
    "guardianWeeklySubscriber": false
  }
}
```

### DIFF
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -20,7 +20,7 @@ import scalaz.syntax.traverse._
 import scalaz.{Disjunction, EitherT, \/}
 import akka.actor.{ActorSystem, Scheduler}
 import com.gu.memsub.subsv2.ReaderType.Gift
-import shapeless.record
+import services.AttributesFromZuora.mergeDigitalSubscriptionExpiryDate
 import utils.FutureRetry._
 
 class AttributesFromZuora(implicit val executionContext: ExecutionContext, system: ActorSystem) extends LoggingWithLogstashFields {
@@ -70,13 +70,13 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
           if(userHasDigiSub(subscriptions)) Future.successful(\/.right[String, List[GiftSubscriptionsFromIdentityIdRecord]](Nil))
           else giftSubscriptionsForIdentityId(identityId)
         )
-        giftAttributes = giftSubscriptions
+        maybeGiftAttributes = giftSubscriptions
           .sortWith((first, second) => first.TermEndDate.isAfter(second.TermEndDate)).headOption
           .map(record => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(record.TermEndDate)))
-        maybeZAttributes <- EitherT(AttributesMaker.zuoraAttributes(identityId, subscriptions, paymentMethodForPaymentMethodId, forDate).map(\/.right[String, Option[ZuoraAttributes]]))
+        maybeRegularAttributes <- EitherT(AttributesMaker.zuoraAttributes(identityId, subscriptions, paymentMethodForPaymentMethodId, forDate).map(\/.right[String, Option[ZuoraAttributes]]))
       } yield {
-        val maybeAttributes = maybeZAttributes.orElse(giftAttributes).map(ZuoraAttributes.asAttributes(_))
-        attributesFromDynamo.foreach(maybeUpdateCache(maybeZAttributes, _, maybeAttributes))
+        val maybeAttributes = mergeDigitalSubscriptionExpiryDate(maybeRegularAttributes, maybeGiftAttributes).map(ZuoraAttributes.asAttributes(_))
+        attributesFromDynamo.foreach(maybeUpdateCache(maybeRegularAttributes, _, maybeAttributes))
         Future.successful("Zuora", maybeAttributes)
       }
 
@@ -189,5 +189,16 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
   }
 
   def queryToAccountIds(response: GetAccountsQueryResponse): List[AccountId] =  response.records.map(_.Id)
+}
+
+object AttributesFromZuora {
+  def mergeDigitalSubscriptionExpiryDate(regular: Option[ZuoraAttributes], gift: Option[ZuoraAttributes]) = {
+    val mergedDigisubExpiry = List(regular.map(_.DigitalSubscriptionExpiryDate), gift.map(_.DigitalSubscriptionExpiryDate))
+      .flatten.flatten
+      .sortWith((first, second) => first.isAfter(second))
+      .headOption
+
+    regular.map(_.copy(DigitalSubscriptionExpiryDate = mergedDigisubExpiry)).orElse(gift)
+  }
 }
 
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -20,6 +20,7 @@ import scalaz.syntax.traverse._
 import scalaz.{Disjunction, EitherT, \/}
 import akka.actor.{ActorSystem, Scheduler}
 import com.gu.memsub.subsv2.ReaderType.Gift
+import shapeless.record
 import utils.FutureRetry._
 
 class AttributesFromZuora(implicit val executionContext: ExecutionContext, system: ActorSystem) extends LoggingWithLogstashFields {
@@ -27,7 +28,7 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
      identityId: String,
      identityIdToAccounts: String => Future[String \/ GetAccountsQueryResponse],
      subscriptionsForAccountId: AccountId => SubPlanReads[AnyPlan] => Future[Disjunction[String, List[Subscription[AnyPlan]]]],
-     giftSubscriptionsForIdentityId: String => Future[Disjunction[String, List[GiftSubscriptionsFromIdentityIdRecord]]],
+     giftSubscriptionsForIdentityId: String => Future[\/[String, List[GiftSubscriptionsFromIdentityIdRecord]]],
      paymentMethodForPaymentMethodId: PaymentMethodId => Future[\/[String, PaymentMethodResponse]],
      dynamoAttributeService: AttributeService,
      forDate: LocalDate = LocalDate.now(),
@@ -69,7 +70,9 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
           if(userHasDigiSub(subscriptions)) Future.successful(\/.right[String, List[GiftSubscriptionsFromIdentityIdRecord]](Nil))
           else giftSubscriptionsForIdentityId(identityId)
         )
-        giftAttributes = giftSubscriptions.headOption.map(record => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(record.TermEndDate)))
+        giftAttributes = giftSubscriptions
+          .sortWith((first, second) => first.TermEndDate.isAfter(second.TermEndDate)).headOption
+          .map(record => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(record.TermEndDate)))
         maybeZAttributes <- EitherT(AttributesMaker.zuoraAttributes(identityId, subscriptions, paymentMethodForPaymentMethodId, forDate).map(\/.right[String, Option[ZuoraAttributes]]))
       } yield {
         val maybeAttributes = maybeZAttributes.orElse(giftAttributes).map(ZuoraAttributes.asAttributes(_))
@@ -154,7 +157,7 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
 
     def accountWithSubscriptions(account: AccountObject)(implicit reads: SubPlanReads[AnyPlan]): Future[Disjunction[String, AccountWithSubscriptions]] = {
       val nonDigipackGiftSubs = subscriptionsForAccountId(account.Id)(anyPlanReads).map(_.map(
-          _.filter(isNotDigipackGiftSub)
+          _.filter(isNotDigipackGiftSub) //Filter out digital pack gift subs where the current user is the purchaser rather than the recipient
       ))
       EitherT(nonDigipackGiftSubs)
         .map(AccountWithSubscriptions(account, _))
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -21,6 +21,7 @@ import scalaz.{Disjunction, EitherT, \/}
 import akka.actor.{ActorSystem, Scheduler}
 import com.gu.memsub.subsv2.ReaderType.Gift
 import services.AttributesFromZuora.mergeDigitalSubscriptionExpiryDate
+import services.AttributesMaker.getSubsWhichIncludeDigitalPack
 import utils.FutureRetry._
 
 class AttributesFromZuora(implicit val executionContext: ExecutionContext, system: ActorSystem) extends LoggingWithLogstashFields {
@@ -56,8 +57,10 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
           case Failure(e) => SafeLogger.error(scrub"Failed to retrieve attributes from cache for $identityId because $e")
         }
 
-    def userHasDigiSub(accountsWithSubscriptions: List[AccountWithSubscriptions]) =
-      accountsWithSubscriptions.flatMap(_.subscriptions).exists(_.asDigipack.isDefined)
+    def userHasDigiSub(accountsWithSubscriptions: List[AccountWithSubscriptions]) = {
+      val subs = accountsWithSubscriptions.flatMap(_.subscriptions)
+      subs.exists(_.asDigipack.isDefined) || getSubsWhichIncludeDigitalPack(subs, LocalDate.now()).nonEmpty
+    }
 
     lazy val getAttrFromZuora =
       for {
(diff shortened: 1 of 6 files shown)


membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -52,13 +55,24 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
...
-        subscriptions <- EitherT(if (accounts.size > 0) getSubscriptions(accounts, subscriptionsForAccountId) else Future.successful(\/.right[String, List[AccountWithSubscriptions]](Nil)))
+        subscriptions <- EitherT(
+          if (accounts.size > 0) getSubscriptions(accounts, subscriptionsForAccountId)
+          else Future.successful(\/.right[String, List[AccountWithSubscriptions]](Nil))
+        )
+        giftSubscriptions <- EitherT(
+          if(userHasDigiSub(subscriptions)) Future.successful(\/.right[String, List[GiftSubscriptionsFromIdentityIdRecord]](Nil))
+          else giftSubscriptionsForIdentityId(identityId)
+        )
+        giftAttributes = giftSubscriptions.headOption.map(record => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(record.TermEndDate)))
```
johnduffell, line 72: how do we know that the first one is the latest expiry date?

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -136,10 +150,16 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
     subscriptionsForAccountId: AccountId => SubPlanReads[AnyPlan] => Future[Disjunction[String, List[Subscription[AnyPlan]]]]
   ): Future[Disjunction[String, List[AccountWithSubscriptions]]] = {
 
-    def accountWithSubscriptions(account: AccountObject)(implicit reads: SubPlanReads[AnyPlan]): Future[Disjunction[String, AccountWithSubscriptions]] =
-      EitherT(subscriptionsForAccountId(account.Id)(anyPlanReads))
+    def isNotDigipackGiftSub(subscription: Subscription[AnyPlan]) = subscription.asDigipack.isEmpty || subscription.readerType != Gift
+
+    def accountWithSubscriptions(account: AccountObject)(implicit reads: SubPlanReads[AnyPlan]): Future[Disjunction[String, AccountWithSubscriptions]] = {
+      val nonDigipackGiftSubs = subscriptionsForAccountId(account.Id)(anyPlanReads).map(_.map(
```
johnduffell, line 156: minor - could have wrapped in the EitherT before the `map.map` which would then only be a single `.map`

project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
   val awsDynamo = "com.amazonaws" % "aws-java-sdk-dynamodb" % awsClientVersion
   val awsSQS = "com.amazonaws" % "aws-java-sdk-sqs" % awsClientVersion
   val awsCloudWatch = "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsClientVersion
-  val membershipCommon = "com.gu" %% "membership-common" % "0.586"
+  val membershipCommon = "com.gu" %% "membership-common" % "0.1-SNAPSHOT"
```
johnduffell, line 20: TBC

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -52,13 +55,24 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
...
+        )
+        giftSubscriptions <- EitherT(
+          if(userHasDigiSub(subscriptions)) Future.successful(\/.right[String, List[GiftSubscriptionsFromIdentityIdRecord]](Nil))
+          else giftSubscriptionsForIdentityId(identityId)
+        )
+        giftAttributes = giftSubscriptions.headOption.map(record => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(record.TermEndDate)))
         maybeZAttributes <- EitherT(AttributesMaker.zuoraAttributes(identityId, subscriptions, paymentMethodForPaymentMethodId, forDate).map(\/.right[String, Option[ZuoraAttributes]]))
       } yield {
-        val maybeAttributes = maybeZAttributes.map(ZuoraAttributes.asAttributes(_))
+        val maybeAttributes = maybeZAttributes.orElse(giftAttributes).map(ZuoraAttributes.asAttributes(_))
```
johnduffell, line 75: if they have another attributable product `maybeZAttributes` will come back as Some, so we wouldn't use the gift

johnduffell: aattributable products are a few https://github.com/guardian/members-data-api/blob/e48de2a0a3b4388ed86e5904f2e5a513de68c0eb/membership-attribute-service/app/services/AttributesMaker.scala#L98-L102

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -52,13 +55,24 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
           case Failure(e) => SafeLogger.error(scrub"Failed to retrieve attributes from cache for $identityId because $e")
         }
 
+    def userHasDigiSub(accountsWithSubscriptions: List[AccountWithSubscriptions]) =
+      accountsWithSubscriptions.flatMap(_.subscriptions).exists(_.asDigipack.isDefined)
```
johnduffell, line 59: it's sensible to take this small optimisation. There is an even smaller additional optimisation, if they have a Paper+Digital subscription.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -52,13 +55,24 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
...
     lazy val getAttrFromZuora =
       for {
         accounts <- EitherT(identityIdToAccounts(identityId))
-        subscriptions <- EitherT(if (accounts.size > 0) getSubscriptions(accounts, subscriptionsForAccountId) else Future.successful(\/.right[String, List[AccountWithSubscriptions]](Nil)))
+        subscriptions <- EitherT(
+          if (accounts.size > 0) getSubscriptions(accounts, subscriptionsForAccountId)
+          else Future.successful(\/.right[String, List[AccountWithSubscriptions]](Nil))
+        )
+        giftSubscriptions <- EitherT(
+          if(userHasDigiSub(subscriptions)) Future.successful(\/.right[String, List[GiftSubscriptionsFromIdentityIdRecord]](Nil))
```
johnduffell, line 69: I can't tell where `subscriptions` has cancelled subs or old fixed terms filtered out. It's not in the query https://github.com/guardian/membership-common/blob/7d9d4ab369d59a8c918f1469d2f32723f34ab0ef/src/main/scala/com/gu/zuora/rest/ZuoraRestService.scala#L375-L378 and I can't see it in Zuora attributes maker.  So I assume it's buried in the anyPlanReads so it probably only does Cancelled subs and not fixed terms that have ended.
As far as I know we never did fixed term digi subs and paper+digital.

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: Well we know that it is active because of the query but you're right that it would be good to pick the one with the latest expiry in the event that there is more than one

johnduffell: active just means the subscription version is the current version, not the dates are current, 
https://knowledgecenter.zuora.com/Billing/Subscriptions/Subscriptions/A_Viewing_a_Subscription#Active_Status_Subtypes:~:text=If%20you%20are%20using%20the%20Zuora,or%20equal%20to%20the%20current%20date.

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: Thanks for spotting this

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: Ah, good to know, thanks!

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: I'm filtering out gifts from this list but we probably need to filter expired corporate subs as well. I've added a card to Trello: https://trello.com/c/AO59JiJX/3138-make-sure-that-expired-corporate-subs-no-longer-give-benefits

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -55,8 +57,10 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
           case Failure(e) => SafeLogger.error(scrub"Failed to retrieve attributes from cache for $identityId because $e")
         }
 
-    def userHasDigiSub(accountsWithSubscriptions: List[AccountWithSubscriptions]) =
-      accountsWithSubscriptions.flatMap(_.subscriptions).exists(_.asDigipack.isDefined)
+    def userHasDigiSub(accountsWithSubscriptions: List[AccountWithSubscriptions]) = {
+      val subs = accountsWithSubscriptions.flatMap(_.subscriptions)
+      subs.exists(_.asDigipack.isDefined) || getSubsWhichIncludeDigitalPack(subs, LocalDate.now()).nonEmpty
```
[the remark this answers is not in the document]
johnduffell: so no need the original check any more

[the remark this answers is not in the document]
rupertbates: Of course!

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -69,11 +73,13 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
           if(userHasDigiSub(subscriptions)) Future.successful(\/.right[String, List[GiftSubscriptionsFromIdentityIdRecord]](Nil))
           else giftSubscriptionsForIdentityId(identityId)
         )
-        giftAttributes = giftSubscriptions.headOption.map(record => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(record.TermEndDate)))
-        maybeZAttributes <- EitherT(AttributesMaker.zuoraAttributes(identityId, subscriptions, paymentMethodForPaymentMethodId, forDate).map(\/.right[String, Option[ZuoraAttributes]]))
+        maybeGiftAttributes = giftSubscriptions
+          .sortWith((first, second) => first.TermEndDate.isAfter(second.TermEndDate)).headOption
```
johnduffell, line 77: sorting always confuses me, also not needed here
https://www.scala-lang.org/api/2.12.12/scala/collection/TraversableOnce.html#maxBy[B](f:A=%3EB):A

so something like this, although haven't checked the joda API to get the proper method name

```suggestion
        maybeGiftAttributes = if (giftSubscriptions.isEmpty) 
          None
        else
          Some(giftSubscriptions.maxBy(_.TermEndDate.toMillisLong))
```

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -69,11 +73,13 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
...
+          .map(record => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(record.TermEndDate)))
```
johnduffell, line 79: sorting always confuses me, also not needed here
https://www.scala-lang.org/api/2.12.12/scala/collection/TraversableOnce.html#maxBy[B](f:A=%3EB):A

so something like this, although haven't checked the joda API to get the proper method name

```suggestion
        maybeGiftAttributes = (if (giftSubscriptions.isEmpty) 
          None
        else
          Some(giftSubscriptions.map(_.TermEndDate).maxBy(_.toMillisLong)))
          .map(termEndDate => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(termEndDate)))
```

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -188,3 +194,14 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
   def queryToAccountIds(response: GetAccountsQueryResponse): List[AccountId] =  response.records.map(_.Id)
 }
 
+object AttributesFromZuora {
+  def mergeDigitalSubscriptionExpiryDate(regular: Option[ZuoraAttributes], gift: Option[ZuoraAttributes]) = {
+    val mergedDigisubExpiry = List(regular.map(_.DigitalSubscriptionExpiryDate), gift.map(_.DigitalSubscriptionExpiryDate))
+      .flatten.flatten
+      .sortWith((first, second) => first.isAfter(second))
```
johnduffell, line 202: can use the same maxBy trick here mainly for clarity

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -188,3 +194,14 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
   def queryToAccountIds(response: GetAccountsQueryResponse): List[AccountId] =  response.records.map(_.Id)
 }
 
+object AttributesFromZuora {
+  def mergeDigitalSubscriptionExpiryDate(regular: Option[ZuoraAttributes], gift: Option[ZuoraAttributes]) = {
+    val mergedDigisubExpiry = List(regular.map(_.DigitalSubscriptionExpiryDate), gift.map(_.DigitalSubscriptionExpiryDate))
+      .flatten.flatten
```
johnduffell, line 201:
```suggestion
    val mergedDigisubExpiry = List(regular, gift).flatten
      .map(_.DigitalSubscriptionExpiryDate).flatten
```

membership-attribute-service/app/services/AttributesFromZuora.scala
johnduffell: thanks, I don't know if it's safer to just have a general TermEndDate check somewhere

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -74,9 +74,10 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
           if(userHasDigiSub(subscriptions)) Future.successful(\/.right[String, List[GiftSubscriptionsFromIdentityIdRecord]](Nil))
           else giftSubscriptionsForIdentityId(identityId)
         )
-        maybeGiftAttributes = giftSubscriptions
-          .sortWith((first, second) => first.TermEndDate.isAfter(second.TermEndDate)).headOption
-          .map(record => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(record.TermEndDate)))
+        maybeGiftAttributes = Option(giftSubscriptions)
+          .filter(_.nonEmpty)
+          .map(_.maxBy(_.TermEndDate.toDateTimeAtStartOfDay.getMillis).TermEndDate)
+          .map(date => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(date)))
```
johnduffell, line 80: great, nice idea using map/filter rather than an if statement.  much easier to follow than before

Option converts null to None, so it suggests a Java->scala conversion.
Let's do the term end date once up front?
as follows:
```suggestion
        maybeGiftAttributes = Some(giftSubscriptions.map(_.TermEndDate))
          .filter(_.nonEmpty)
          .map(_.maxBy(_.toDateTimeAtStartOfDay.getMillis))
          .map(date => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(date)))
```

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -197,12 +198,15 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
...
-      .sortWith((first, second) => first.isAfter(second))
-      .headOption
+    val maybeExpiryDates = Option(
+      List(regular, gift).flatten.flatMap(_.DigitalSubscriptionExpiryDate)
+    ).filter(_.nonEmpty)
 
-    regular.map(_.copy(DigitalSubscriptionExpiryDate = mergedDigisubExpiry)).orElse(gift)
+    val maybeLatestDate = maybeExpiryDates.map(
+      _.maxBy(_.toDateTimeAtStartOfDay.getMillis)
+    )
```
johnduffell, line 207: we're getting in to diminishing returns by now, but I have no idea why maxBy doesn't return an Option, as we seem to have more lines of code than using sortWith (although clearer)
do you think it would be clearer to implement a version of maxBy that returns an option, and using that in both places?

johnduffell (approved):
yeah I  have added some comments but I don't want to keep us going round in ever decreasing circles, good to go if you want to move on 👍 
nice work on managing to retro fit this withough adding too many extra zuora calls, and without doing too much reinterpreting the existing code.
Do we have a plan to monitor things once this goes out?

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: Tbh I don't think it's worth it, we've been through about 3 different formulations of this code now let's stick with what we've got.

johnduffell (approved):

PR #480, opened 2020-10-29
Validate contribution amount updates so must be >= 2

kelvin-chappell:
It's currently possible to update a contribution to 0.
This change stops that from happening by imposing more stringent serverside validation.
The rule is that no contribution < 2 currency units is allowed.

A lot of code has been reformatted because of [the scalafmt configuration](https://github.com/guardian/members-data-api/blob/main/.scalafmt.conf).
So best to concentrate on [this key commit](https://github.com/guardian/members-data-api/commit/aa2ec404aa9b895a8627a5ce37ee098cbc4a74c7).

I've moved [the validation code](https://github.com/guardian/members-data-api/commit/aa2ec404aa9b895a8627a5ce37ee098cbc4a74c7#diff-51be3bf42924f91138aeb210e429dd677197759ced0c609d68cda120318c2052R508-R515) out to make it easier to understand as the controller action is very complicated.
This also makes it easier [to test](https://github.com/guardian/members-data-api/commit/aa2ec404aa9b895a8627a5ce37ee098cbc4a74c7#diff-9521b1b5fcf0eb02c98324a2e715105ab7f9b9305797fc0de7397c226a96018dR1-R35).

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,1 +1,3 @@
-maxColumn = 100
+version = 2.7.5
+
+maxColumn = 150
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -4,12 +4,11 @@ import actions._
 import com.gu.memsub
 import services.PaymentFailureAlerter._
 import com.gu.memsub._
-import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
 import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.memsub.subsv2.services.SubscriptionService
-import com.gu.memsub.subsv2.{PaidChargeList, PaidSubscriptionPlan, Subscription, SubscriptionPlan}
+import com.gu.memsub.subsv2.{PaidChargeList, Subscription, SubscriptionPlan}
 import com.gu.monitoring.SafeLogger
 import com.gu.monitoring.SafeLogger._
 import com.gu.salesforce.SimpleContactRepository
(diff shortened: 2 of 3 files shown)


twrichards:
glad to see some server side validation being added, however the client side (in `manage-frontend`) also has validation (which perhaps needs reviewing first as its obviously got a bug @rBangay ).

The upper and lower thresholds are different for each currency and often change (as a result of the findings of a/b tests by the contributions team). When the client-side contributions changer component was re-written by @rBangay we discussed with @tomrf1 and others the idea of a shared config file (probs in S3) which could drive the 'control' version of `support-frontend` and the client-side validation in `manage-frontend` but it was never implemented. Perhaps now is the time to do that and also make use of it in this new server-side validation in `members-data-api`?!

tomrf1:
£1 contributions are very much allowed!
Screen Shot 2020-10-29 at 10 40 04

tomrf1:
I'm not sure what case this is solving, is it for modifying an existing recurring contribution?

kelvin-chappell:
@twrichards Thanks for the info.  I wasn't aware of the shared config idea but I'll be looking at how the clientside validation works next

kelvin-chappell:
@tomrf1 There was a case raised in AU where a contributor changed the amount of their recurring contribution to 0.  Presumably as a way of cancelling it.  I'll change the min value to 1. Thanks.

twrichards:
> @twrichards Thanks for the info. I wasn't aware of the shared config idea but I'll be looking at how the client-side validation works next

Perhaps best to look at it first - otherwise you'll get errors on the frontend without a nice explainer message.

kelvin-chappell:
@twrichards Don't worry! I won't merge this until it's working nicely on the frontend

kelvin-chappell:
Clientside work is now done in guardian/manage-frontend#516 so this can go out without giving a bad user experience.

kelvin-chappell:
Added shared validation config idea to our board for future health discussion [here](https://trello.com/c/L6f8dcV5/1727-shared-validation-between-clientside-and-serverside-manage-frontend).

PR #481, opened 2020-11-04
Expose Digital Subscription giftees to MMA

rupertbates:
### Why do we need this? 
Once we start selling Digital Subscription gifts we will want users to be able to view the details of these subs in the Manage My Account pages. 
Gift subs will already show up for the _purchaser_ of the subscription but not for the _giftee_, this PR adds that functionality.

### The changes 
Digital Subscription gifts are represented in Zuora as a single subscription for both purchaser and redeemer (giftee) with a custom field `gifteeIdentityId` recording the identity id of the giftee. 
Therefore to work out whether a user is a DS giftee we need to query Zuora for subscriptions where the `gifteeIdentityId` matches the identity id of the currently logged in user.
This PR relies on [changes in membership-common](https://github.com/guardian/membership-common/pull/634)

https://trello.com/c/vtkGm5Wm/3392-gifting-add-giftee-and-gifter-digital-subs-to-mma-members-data-api-changes

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -13,7 +13,7 @@ import com.gu.memsub.subsv2.services.SubscriptionService
 import com.gu.memsub.subsv2.{PaidChargeList, PaidSubscriptionPlan, Subscription, SubscriptionPlan}
 import com.gu.monitoring.SafeLogger
 import com.gu.monitoring.SafeLogger._
-import com.gu.salesforce.SimpleContactRepository
+import com.gu.salesforce.{Contact, SimpleContactRepository}
 import com.gu.services.model.PaymentDetails
 import com.gu.stripe.{Stripe, StripeService}
 import com.gu.zuora.api.RegionalStripeGateways
@@ -28,7 +28,7 @@ import models.{AccountDetails, ApiError, CancelledSubscription, ContactAndSubscr
 import org.joda.time.{LocalDate, LocalTime}
 import play.api.data.Form
 import play.api.data.Forms._
-import play.api.libs.json.{Format, JsObject, Json}
+import play.api.libs.json.{Format, JsObject, Json, __}
 import play.api.mvc.{Action, AnyContent, BaseController, ControllerComponents}
 import scalaz.std.option._
 import scalaz.std.scalaFuture._
@@ -36,6 +36,7 @@ import scalaz.syntax.monad._
 import scalaz.syntax.std.option._
 import scalaz.syntax.traverse._
 import scalaz.{-\/, EitherT, OptionT, \/, \/-}
+import utils.OptionEither.FutureEither
 import utils.{ListEither, OptionEither}
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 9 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -236,24 +241,53 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
           }
-        ) // TODO are we happy with an empty list in case of error ?!?!
-      filteredIfApplicable = filter match {
+        )
+
+    contactAndSubscriptions <- checkForGiftSubscription(
+        userId,
+        subService,
+        zuoraRestService,
+        nonGiftContactAndSubscriptions,
```
johnduffell, line 287: not so keen on passing these in just to have them passed back.  is it worth returning an option and then doing something like  `List( nonGiftContactAndSubscriptions, giftContactAndSubscriptions).flatten` to make the higher level structure clearer?

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -269,23 +303,72 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+            s"with account id ${contactAndSubscription.subscription.accountId}. Reason: $x")
+        })
+      effectiveCancellationDate <- ListEither.liftList(
+        tp.zuoraRestService.getCancellationEffectiveDate(contactAndSubscription.subscription.name).recover {
+          case x => \/.left(s"Failed to fetch effective cancellation date: ${contactAndSubscription.subscription.name} " +
+            s"with account id ${contactAndSubscription.subscription.accountId}. Reason: $x")
+        })
+      stripeService = accountSummary.billToContact.country.map(RegionalStripeGateways.getGatewayForCountry)
+        .flatMap(tp.stripeServicesByPaymentGateway.get)
+        .getOrElse(tp.ukStripeService)
```
johnduffell, line 371: this is very hard to read! I can see you have tidied it up a little which is great.  monad transfomers plus inline recover functions really add a lot of clutter.  I assume there are no substantial changes here.

project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
   val awsDynamo = "com.amazonaws" % "aws-java-sdk-dynamodb" % awsClientVersion
   val awsSQS = "com.amazonaws" % "aws-java-sdk-sqs" % awsClientVersion
   val awsCloudWatch = "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsClientVersion
-  val membershipCommon = "com.gu" %% "membership-common" % "0.589"
+  val membershipCommon = "com.gu" %% "membership-common" % "0.1-SNAPSHOT"
```
johnduffell, line 20: TODO

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -269,23 +303,72 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
+    remainingTrialLength = 0
+  )
+
+  def getPaymentDetailsForSub(
+    maybeUserId: Option[String],
+    freeOrPaidSub: Subscription[SubscriptionPlan.Free] \/ Subscription[SubscriptionPlan.Paid],
+    paymentService: PaymentService
+  ): Future[PaymentDetails] = freeOrPaidSub match {
+      case \/-(paidSub) if paidSub.gifteeIdentityId != maybeUserId =>
+        paymentService.paymentDetails(freeOrPaidSub, defaultMandateIdIfApplicable = Some(""))
```
johnduffell, line 332: so the original functionality is only called if the giftee identity id is not the same as the user logged in.
I think this will work, but it is a bit obscured given that this is protecting someone else's payment information.

Ideally I'd prefer to make the decision what kind of payment info to insert before combining the giftee subscription with the self-owned subscriptions.  So move this decision into the `allCurrentSubscriptions`.  Then there is no mixing of the code paths for getting payment details and getting giftee subscriptions.

Failing that we can make the pattern match check more explicitly that it's a gift ReaderType and also add a test to make sure the overall logic doesn't get broken.

johnduffell:
good work getting to the bottom of that - I've added some comments - I think it probably needs at least one test if possible, and see comments about the payment details.

membership-attribute-service/app/controllers/AccountController.scala
rupertbates: It's really tough to work through the types in the parent method, the comprehension is already using an `OptionT[FutureEither, A]` so we'd end up having to return a `OptionT[FutureEither, Option[Subscription]` somehow which is just horrendous. I've had a quick go at it and it doesn't seem to line up properly anyway which was why I decided to just encapsulate the whole mess inside one method and return the type we actually want.

membership-attribute-service/app/controllers/AccountController.scala
rupertbates: No only formatting changes

membership-attribute-service/app/controllers/AccountController.scala
rupertbates: Ok I've swapped things around a bit to make it more obvious, I've also added a check on the ReaderType and some tests.

johnduffell (approved):
looks great although you already have a conflict from Kelvin's PR by the looks
👍

PR #484, opened 2020-12-01
MMA fix for Digital Subscription gift subscriptions

rupertbates:
### Why do we need this? 
#481 had a bug which was that if a user both purchases and then redeems a DS gift subscription their `/mma` data will redact payment details from both the sub that represents them as a purchaser as well as one which represents them as a gifter. This is incorrect as we should retain the payment details for the purchaser sub.

This PR fixes this bug.

### MMA page after the fix:
Screen Shot 2020-12-01 at 10 54 22

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -31,7 +31,7 @@ import models._
 import org.joda.time.LocalDate
 import play.api.data.Form
 import play.api.data.Forms._
-import play.api.libs.json.Json
+import play.api.libs.json.{Json, __}
 import play.api.mvc.{Action, AnyContent, BaseController, ControllerComponents, Request}
 import scalaz.std.option._
 import scalaz.std.scalaFuture._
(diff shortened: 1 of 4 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -300,14 +300,30 @@ class AccountController(commonActions: CommonActions, override val controllerCom
...
 
+  def reuseAlreadyFetchedSubscriptionIfAvailable(
+    giftRecords:  List[ZuoraRestService.GiftSubscriptionsFromIdentityIdRecord],
+    nonGiftSubs: List[ContactAndSubscription],
+    subscriptionService: SubscriptionService[Future],
+  ) = {
+    val subscriptionName = Name(giftRecords.head.Name)
+    OptionEither.liftFutureOption(
+      // If the current user is both the gifter and the giftee we will have already retrieved their
+      // subscription so we can reuse it and avoid a call to Zuora
```
johnduffell, line 319: is this an important optimisation considering the likeihood of it happening, multiplied by the chance of the person going into MMA regularly enough to slow things down?

membership-attribute-service/app/controllers/PaymentDetailMapper.scala
```diff
@@ -33,11 +33,11 @@ object PaymentDetailMapper {
   )
 
   def paymentDetailsForSub(
-    maybeUserId: Option[String],
+    isGiftRedemption: Boolean,
     freeOrPaidSub: Subscription[SubscriptionPlan.Free] \/ Subscription[SubscriptionPlan.Paid],
     paymentService: PaymentService
   ): Future[PaymentDetails] = freeOrPaidSub match {
-    case \/-(giftSub) if giftSub.gifteeIdentityId == maybeUserId && giftSub.readerType == Gift =>
+    case \/-(giftSub) if giftSub.readerType == Gift && isGiftRedemption =>
```
johnduffell, line 40:
```suggestion
    case \/-(giftSub) if isGiftRedemption =>
```

johnduffell: wouldn't that be enough - we already know it's a gift as it has the giftee identity id filled in? https://github.com/guardian/membership-common/blob/74ec1966b3393d3eb2b86b8c247ac875ffc4b254/src/main/scala/com/gu/zuora/rest/ZuoraRestService.scala#L405
`select name, id, termEndDate from subscription where GifteeIdentityId__c = '${identityId}' and status = 'Active' and termEndDate >= '$today'`

johnduffell (approved):
great, the only question about the optimisation, if it's adding complexity is that worthwhile
👍

membership-attribute-service/app/controllers/AccountController.scala
rupertbates: If it wasn't for the fact that we are constantly over quota on our requests to Zuora during the day I'd say no, but given that we are I think it's worth reducing the number of calls we make wherever possible.

PR #486, opened 2020-12-11
UAT stage is named CODE in instance count

mario-galic:
### Why do we need this? 

PROD logs were filled with false positive error message

```
ERROR: There should be at least one healthy in-service instance. Failing to default value 6
```

This error message actually pertained to CODE (as we switch "touchpoint" backends in PROD), and was caused by config `stage` set to `UAT`, however in AWS it is known as `CODE`. PROD was unaffected.

### DIFF
--- a/membership-attribute-service/app/limit/InstanceCountOnSchedule.scala
+++ b/membership-attribute-service/app/limit/InstanceCountOnSchedule.scala
@@ -26,7 +26,7 @@ class InstanceCountOnSchedule(stage: String)(implicit ec: ExecutionContext, syst
       .getAutoScalingGroups
       .asScala
       .toList
-      .filter(_.getAutoScalingGroupName.startsWith(s"Memb-Attributes-$stage"))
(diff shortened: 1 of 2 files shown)


kelvin-chappell (approved):
👍

membership-attribute-service/app/limit/InstanceCountOnSchedule.scala
```diff
@@ -26,7 +26,7 @@ class InstanceCountOnSchedule(stage: String)(implicit ec: ExecutionContext, syst
       .getAutoScalingGroups
       .asScala
       .toList
-      .filter(_.getAutoScalingGroupName.startsWith(s"Memb-Attributes-$stage"))
+      .filter(_.getAutoScalingGroupName.startsWith(s"Memb-Attributes-${if (stage == "UAT") "CODE" else stage}"))
```
johnduffell, line 29: actually in UAT stage it is hit by both prod and code (and dev) so to be strictly correct this should get the total for PROD and CODE, and total them up.  i.e. 7.
However, we don't really care about overloading the UAT env, and it's unlikely to happen, so this fix would not cause any problem.

PR #488, opened 2021-01-11
write giftee attributes to dynamo, as well as direct

johnduffell:
This PR updates the attributes code so that a change in giftee status will update the dynamo attributes.

This means that giftees will get their benefits whether or not they are served from the cache or direct from zuora.  Previously they only came through from zuora.

https://trello.com/c/pWBlEksl/3561-members-data-api-gifting-intermittent-cache-issue

### DIFF
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -20,6 +20,7 @@ import scalaz.syntax.traverse._
 import scalaz.{Disjunction, EitherT, \/}
 import akka.actor.{ActorSystem, Scheduler}
 import com.gu.memsub.subsv2.ReaderType.Gift
+import monitoring.ExpensiveMetrics
 import services.AttributesFromZuora.mergeDigitalSubscriptionExpiryDate
 import services.AttributesMaker.getSubsWhichIncludeDigitalPack
 import utils.FutureRetry._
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -74,19 +78,21 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
...
-        Future.successful("Zuora", maybeAttributes)
+        val maybeGifteeAndNonGifteeAttributes = mergeDigitalSubscriptionExpiryDate(maybeNonGifteeAttributes, maybeGifteeAttributes)
+        val maybeAttributesFromZuora = maybeGifteeAndNonGifteeAttributes.map(ZuoraAttributes.asAttributes(_))
+        attributesFromDynamo.foreach(maybeUpdateCache(maybeGifteeAndNonGifteeAttributes, _, maybeAttributesFromZuora))
+        Future.successful("Zuora", maybeAttributesFromZuora)
       }
 
     lazy val fallbackToCache = (zuoraError: String) => {
       SafeLogger.error(scrub"Failed to retrieve attributes from Zuora for $identityId because $zuoraError so falling back on cache.")
+      expensiveMetrics.countRequest("cache-fallback")
```
johnduffell, line 95: if we were sticking with this implementation, we would need an alarm on this - as we have previously had situations where large portions of traffic were getting silently served from the cache

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -35,10 +36,13 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
      forDate: LocalDate = LocalDate.now(),
   ): Future[(String, Option[Attributes])] = {
 
+    lazy val expensiveMetrics = new ExpensiveMetrics("AttributesFromZuora")
+
     def maybeUpdateCache(zuoraAttributes: Option[ZuoraAttributes], dynamoAttributes: Option[DynamoAttributes], attributes: Option[Attributes]): Unit = {
       def twoWeekExpiry = forDate.toDateTimeAtStartOfDay.plusDays(14)
       if(dynamoUpdateRequired(dynamoAttributes, zuoraAttributes, identityId, twoWeekExpiry)) {
         log.info(s"Attempting to update cache for $identityId ...")
+        expensiveMetrics.countRequest("cache-update")
```
johnduffell, line 45: this would be higher if the cache code is not adding people to the cache.  There is already a known bug with cancelled  memberships not being removed from the cache.  Not sure if it would be possible to alarm on that though.  Also this code is likely to change completely soon.

rupertbates (approved):
Looks good to me, well done for spotting this bug!

johnduffell:
new metrics are working okay

image

PR #489, opened 2021-01-12
only send metric once per period, not on every call

johnduffell:
This fixes https://github.com/guardian/members-data-api/pull/488

The metric was getting sent on every request rather than at intervals! 🙈

### DIFF
--- a/membership-attribute-service/app/services/AttributesFromZuora.scala
+++ b/membership-attribute-service/app/services/AttributesFromZuora.scala
@@ -26,6 +26,9 @@ import services.AttributesMaker.getSubsWhichIncludeDigitalPack
 import utils.FutureRetry._
 
 class AttributesFromZuora(implicit val executionContext: ExecutionContext, system: ActorSystem) extends LoggingWithLogstashFields {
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -26,6 +26,9 @@ import services.AttributesMaker.getSubsWhichIncludeDigitalPack
 import utils.FutureRetry._
 
 class AttributesFromZuora(implicit val executionContext: ExecutionContext, system: ActorSystem) extends LoggingWithLogstashFields {
```
johnduffell, line 28: there is only one AttributesFromZuora created, here https://github.com/guardian/members-data-api/blob/44e10cc002f3a4b8211200c89b7268088d80c7f6/membership-attribute-service/app/wiring/AppLoader.scala#L57

johnduffell:
this cost us about 150$ and would have cost the same every day until fixed
image

PR #490, opened 2021-01-21
Use t4g.small instance type

tomrf1:
Which is cheaper than the current t3.small:
Screen Shot 2021-01-21 at 12 22 22

The cpu use of members-data-api instances on t3.small is very flat:
Screen Shot 2021-01-21 at 11 12 59
Screen Shot 2021-01-21 at 11 13 40
Screen Shot 2021-01-21 at 11 14 14

We just need to ensure that credit use does not increase significantly with t4g.

We had planned on running t3 and t4g at the same time within the same ASG, for a fairer comparison. But this gets messy because they need a separate AMI, and riffraff does not support this.

### DIFF
--- a/membership-attribute-service/conf/riff-raff.yaml
+++ b/membership-attribute-service/conf/riff-raff.yaml
@@ -7,7 +7,7 @@ deployments:
     parameters:
       templatePath: membership-attribute-service.yaml
       amiTags:
-        Recipe: xenial-membership
+        Recipe: bionic-membership-ARM
         AmigoStage: PROD
       amiEncrypted: true
       amiParameter: AmiId
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -12,6 +12,7 @@ Parameters:
     AllowedValues:
     - t2.small
     - t3.small
+    - t4g.small
     ConstraintDescription: must be a valid EC2 instance type.
   VpcId:
     Description: ID of the VPC onto which to launch the application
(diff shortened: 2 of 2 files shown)


membership-attribute-service/conf/riff-raff.yaml
```diff
@@ -7,12 +7,12 @@ deployments:
     parameters:
       templatePath: membership-attribute-service.yaml
       amiTags:
-        Recipe: xenial-membership
```
tomrf1, line 10: there was an issue with the ami build on xenial, but there's no reason not to upgrade

johnduffell:
given CPU is so light, why do we have 6 instances? Is it because of the Network limits?  Would it be worth checking the comparison about network and other resources (I assume those and RAM are the same, but worth asking the question)

tomrf1:
> given CPU is so light, why do we have 6 instances? Is it because of the Network limits? Would it be worth checking the comparison about network and other resources (I assume those and RAM are the same, but worth asking the question)

Yes, @TomPretty and I were also wondering this.
It's definitely worth exploring, though I think it's a separate task

johnduffell:
agree, it looks like the network bandwidth is the same, so this change won't affect it

johnduffell (approved):
great work, let's see what happens! 👏

paulbrown1982:
The main reason for 6 is to support push notification traffic. That 6 was set a couple of years ago though now, and the once-per-day caching mechanism on frontend (incl DCR) and liveblogs has improved since then. So definitely worth considering if we can support just 3 instances going forward.

tomrf1:
Follow up - no significant difference in CPU or credit use since changing to t4g.small:

Screen Shot 2021-01-22 at 08 33 36
Screen Shot 2021-01-22 at 08 33 42
Screen Shot 2021-01-22 at 08 33 51

PR #491, opened 2021-02-04
Log which Zuora environment is having auth errors

mario-galic:
Related PR: https://github.com/guardian/membership-common/pull/635

Before

```
ERROR: Scheduled Task: ZuoraPing failed to execute query: null
```

After

```
ERROR: Scheduled Task: CODE ZuoraPing failed to execute query: null
```


kelvin-chappell (approved):
👍

PR #492, opened 2021-02-24
Do not disable autoPay on cancellation if same account has multiple subscriptions

mario-galic:
### Why do we need this? 

After cancellation of one subscription we need to continue taking payments on other subscriptions that are still active within the account.

### The changes 

* [subscriptionsForAccountId](https://github.com/guardian/membership-common/blob/72db910ddbaf8f45920d05effdc6264dac86ef5f/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionService.scala#L198) should return all currently active subscriptions by accountId
* We disable auto-pay only if there exists a single subscription
* Tested end-to-end by acquiring multiple subscriptions within same billing account via Salesforce CSR functionality, cancelling one of them, and checking if autoPay remains on
  * fetch janus credentials
  * `sudo nginx`
  * `./contributions-platform/contributions-store/contributions-store-bastion/scripts/open_ssh_tunnel.sh -s CODE`
  *  MMA app `yarn watch`
  * MDAPI `./start-api.sh`
  * Enable cancellation temporarily for all countries via[ `SelfServiceCancellation.scala`](https://github.com/guardian/members-data-api/blob/444c861d404204c42f7e73a1385ac7264216d6d3/membership-attribute-service/app/models/SelfServiceCancellation.scala#L34)
  * Acquire one subscriptions via https://support.code.dev-theguardian.com
  * Make sure to not use magic string so that it targets DEV Zuora/Salesforce environments
  * Acquire another subscription within same billing account via Salesforce by going to the billing account (Zuora__CustomerAccount__c) and selecting `New Subscription`
  * Cancel one subscription via `Manage subscription` in https://manage.thegulocal.com 
  * Go to Zuora `Customer Accounts` and check if `Auto-Pay` is still on under `Billing and Payment Info`

https://trello.com/c/HMYQ06Gy/1962-self-service-cancellation-is-dangerous-for-users-who-have-multiple-active-products-in-1-billing-account-as-it-turns-off-auto-pay

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -101,6 +101,23 @@ class AccountController(commonActions: CommonActions, override val controllerCom
           }
       }
 
+      /**
+       * If user has multiple subscriptions within the same billing account, then disabling auto-pay
+       * on the account would stop collecting payments for all subscriptions including the non-cancelled ones.
+       * In this case debt would start to accumulate in the form of positive Zuora account balance, and if at
+       * any point auto-pay is switched back on, then payment for the entire amount would be attempted.
+       */
+      def disableAutoPayOnlyIfAccountHasOneSubscription(
+        accountId: memsub.Subscription.AccountId
+      ): EitherT[Future, String, Future[String \/ Unit]] = {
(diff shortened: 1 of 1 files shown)


kelvin-chappell (approved):
👍

PR #506, opened 2021-02-25
reduce instances to 3

johnduffell:
We increased to 6 a while ago due to some big event possible trump election, but now we want to see if 3 is ok for the reservations.
The orignial PR doesn't have much information https://github.com/guardian/members-data-api/pull/344


ripecosta (approved):
👍

PR #507, opened 2021-02-26
Add reminder status endpoint

TomPretty:
## What? 
Add an endpoint at `/user-attributes/me/reminders` for getting information about a users reminders. Currently we just return info about their recurring reminder which can be one of: `NotSet`, `Active`, and `Cancelled`. This is used by MMA to display/hide a consent on the email preferences page that allows a user to cancel/reactivate a recurring reminder.

## Images
mma email-prefs page:
Screenshot 2021-02-26 at 13 40 25

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,9 @@ dynamodb-local
 api/conf/keys.conf
 api/public/
 .bsp
+
+.bloop/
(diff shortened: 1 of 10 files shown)


tomrf1 (approved):
💯

PR #508, opened 2021-03-03
Return reminderCode from endpoint

TomPretty:
### Why do we need this? 

Return `recurringReminderCode` from the reminders endpoint for use by the MMA client to cancel/reactivate a reminder with the `support-reminders` service.

**DO NOT MERGE** until we update both the PROD databases with this new field.

### DIFF
--- a/membership-attribute-service/app/models/SupportReminders.scala
+++ b/membership-attribute-service/app/models/SupportReminders.scala
@@ -24,7 +24,8 @@ object RecurringReminderStatus {
 }
 
 case class SupportReminderDb (
-  isCancelled: Boolean,
+  is_cancelled: Boolean,
+  reminder_code: java.util.UUID,
 )
 
 object SupportReminderDb {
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/models/SupportReminders.scala
```diff
@@ -25,6 +25,7 @@ object RecurringReminderStatus {
 
 case class SupportReminderDb (
   isCancelled: Boolean,
+  reminder_code: String,
```
TomPretty, line 28: maybe I should camel case `isCancelled` for consistency?

TomPretty: I'm just gonna go ahead and do that, I can't face the mixed casing 😂

tomrf1: 👍

membership-attribute-service/app/models/SupportReminders.scala
```diff
@@ -24,7 +24,8 @@ object RecurringReminderStatus {
 }
 
 case class SupportReminderDb (
-  isCancelled: Boolean,
+  is_cancelled: Boolean,
+  reminder_code: String,
```
tomrf1, line 28: Should this be `Option[String]`?

tomrf1 (approved):
👍

PR #509, opened 2021-03-04
Supporter product data

rupertbates:
### Why do we need this? 

This is the first part of introducing the new supporter-product-data store into members-data-api

The ultimate aim is to replace any api calls to Zuora with requests to a DynamoDB table which is kept up to date with active subscriptions by a scheduled step function. 

This PR the code to request data from the Dynamo table and logic to check whether the results returned are the same as what would have been returned from Zuora, if they are not it fires an alert.

Once we have established that the new system is reliable then there will be another PR to switch over to using it exclusively and to remove the dependency on Zuora.


cloudformation/membership-attribute-service.yaml
```diff
@@ -556,6 +572,31 @@ Resources:
           ReturnData: false
     DependsOn:
       - LoadBalancer
+  AttributesMismatchAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdMonitoring
+    Properties:
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
```
johnduffell, line 580: do we even need an alarm or just keep an eye on the metric every few days?  Or just send the alarm to one or two of us.  I don't see why we need to notify the whole group.

membership-attribute-service/app/models/SupporterRatePlanItem.scala
```diff
@@ -0,0 +1,10 @@
+package models
+
+import org.joda.time.LocalDate
+
+case class SupporterRatePlanItem(
+    identityId: String, //Unique identifier for user
+    ratePlanId: String, //Unique identifier for this product purchase for this user
+    productRatePlanId: String, //Unique identifier for the product in this rate plan
+    termEndDate: LocalDate //Date that this subscription term ends
```
johnduffell, line 9:
```suggestion
  identityId: String, //Unique identifier for user
  ratePlanId: String, //Unique identifier for this product purchase for this user
  productRatePlanId: String, //Unique identifier for the product in this rate plan
  termEndDate: LocalDate //Date that this subscription term ends
```

membership-attribute-service/app/models/SupporterRatePlanItem.scala
```diff
@@ -0,0 +1,10 @@
...
+)
```
johnduffell, line 10: this is the dynamo format for the item I guess? it's not clear from the name

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -84,9 +86,13 @@ class AttributesFromZuora(implicit val executionContext: ExecutionContext, syste
           .map(_.maxBy(_.toDateTimeAtStartOfDay.getMillis))
           .map(date => ZuoraAttributes(identityId, DigitalSubscriptionExpiryDate = Some(date)))
         maybeNonGifteeAttributes <- EitherT(AttributesMaker.zuoraAttributes(identityId, subscriptions, paymentMethodForPaymentMethodId, forDate).map(\/.right[String, Option[ZuoraAttributes]]))
+        supporterProductDataAttributes <- EitherT(supporterProductDataService.getAttributes(identityId))
```
johnduffell, line 89: I guess dynamo is really quick, but this will happen after zuora has returned its results, so would reduce the overall concurrency.  Could put it in a val before the for comprehension to encourage it to run in parallel.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -215,5 +221,40 @@ object AttributesFromZuora {
...
+          fromZuora.map(_.RecurringContributionPaymentPlan) == fromSupporterProductData.map(_.RecurringContributionPaymentPlan) &&
+          datesAreEqualEnough(fromZuora.flatMap(_.DigitalSubscriptionExpiryDate), fromSupporterProductData.flatMap(_.DigitalSubscriptionExpiryDate)) &&
+          datesAreEqualEnough(fromZuora.flatMap(_.PaperSubscriptionExpiryDate), fromSupporterProductData.flatMap(_.PaperSubscriptionExpiryDate)) &&
+          datesAreEqualEnough(fromZuora.flatMap(_.GuardianWeeklySubscriptionExpiryDate), fromSupporterProductData.flatMap(_.GuardianWeeklySubscriptionExpiryDate))
+          )
+      )
+
+  def datesAreEqualEnough(fromZuora: Option[LocalDate], fromSupporterProductData: Option[LocalDate]) = {
+    //Currently dates from the supporter-product-data store will be a day after the zuora term end date to avoid
+    // worrying about time zones and when new subscriptions are created. This may well change.
```
johnduffell, line 250: this is because the zuora renewal job doesn't happen at midnight UTC.  That problem of offset will already hit us, but the main problem currently is that data can still be served from the cache for quite a few days after that.

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -215,5 +221,40 @@ object AttributesFromZuora {
 
     regular.map(_.copy(DigitalSubscriptionExpiryDate = maybeLatestDate)).orElse(gift)
   }
+
+  val metrics = Metrics("AttributesFromZuora") //referenced in CloudFormation
```
johnduffell, line 225: good, this is in an object, not created once per request! 👍  🙈

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -215,5 +221,40 @@ object AttributesFromZuora {
...
+    if (attributesDoNotMatch(fromZuora, fromSupporterProductData)) {
+      logger.warn(
+        "There was a mismatch between the attributes returned by Zuora and the supporter-product-data data store\n" +
+        s"Zuora returned: $fromZuora\n" +
+        s"supporter-product-data returned: $fromSupporterProductData"
+      )
+      metrics.put("AttributesMismatch", 1) //referenced in CloudFormation
+    } else logger.info("attributes returned from Zuora match those returned from supporter-product-data")
+
+  def attributesDoNotMatch(fromZuora: Option[Attributes], fromSupporterProductData: Option[Attributes]) =
```
johnduffell, line 237: why not change this to attributesMatch, then flip the order of the if statment, just to get rid of the negative

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -215,5 +221,40 @@ object AttributesFromZuora {
...
+
+  def datesAreEqualEnough(fromZuora: Option[LocalDate], fromSupporterProductData: Option[LocalDate]) = {
+    //Currently dates from the supporter-product-data store will be a day after the zuora term end date to avoid
+    // worrying about time zones and when new subscriptions are created. This may well change.
+    (fromZuora.isEmpty && fromSupporterProductData.isEmpty) || (for{
+        zuora <- fromZuora
+        supporter <- fromSupporterProductData
+      } yield {
+      zuora == supporter || zuora.plusDays(1) == supporter
+    }).getOrElse(false)
```
johnduffell, line 256: I think it's the same as this
```suggestion
    fromZuora == fromSupporterProductData || 
      fromZuora.map(_.plusDays(1)) == fromSupporterProductData
```

membership-attribute-service/app/services/SupporterProductDataService.scala
```diff
@@ -0,0 +1,45 @@
+package services
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsync
+import com.gu.scanamo.error.DynamoReadError.describe
+import com.gu.scanamo.syntax._
+import com.gu.scanamo.{ScanamoAsync, Table, _}
```
johnduffell, line 6: that's a strange import, woudln't `_` on its own be the same?

membership-attribute-service/app/services/SupporterProductDataService.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.joda.time.{Instant, LocalDate}
+import scalaz.\/
+
+import scala.concurrent.ExecutionContext
+
+class SupporterProductDataService(client: AmazonDynamoDBAsync, table: String, mapper: SupporterRatePlanToAttributesMapper)
+  (implicit executionContext: ExecutionContext) extends LazyLogging {
+
+  implicit val jodaNumberFormat: DynamoFormat[LocalDate] =
+    DynamoFormat.coercedXmap[LocalDate, Long, IllegalArgumentException](epochSeconds => Instant.ofEpochSecond(epochSeconds).toDateTime.toLocalDate)(
```
johnduffell, line 18: I'm sure it does, but does this work also in BST as well as GMT?

membership-attribute-service/app/services/SupporterProductDataService.scala
```diff
@@ -0,0 +1,45 @@
...
+      _.toDateTimeAtStartOfDay.getMillis / 1000
+    )
+
+  def getAttributes(identityId: String) =
+    for {
+      futureDynamoResult <- getSupporterRatePlanItems(identityId)
+      futureErrors = futureDynamoResult.collect { case Left(error) =>
+        error
+      }
+      _ = futureErrors.foreach(error => logger.error(error.toString))
```
johnduffell, line 29: I think we also decided to alarm on failed fetches so we are aware of possible silent missing data

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
```diff
@@ -0,0 +1,321 @@
...
+
+object SupporterRatePlanToAttributesMapper {
+
+  val productRatePlanMappings: Map[String, Map[List[String], (Attributes, SupporterRatePlanItem) => Attributes]] =
+    Map(
+    "PROD" -> Map(
+      List(
+        "2c92a0fb4edd70c8014edeaa4eae220a",
+        "2c92a0fb4edd70c8014edeaa4e972204",
+        "2c92a00d71c96bac0171df3a5622740f",
```
johnduffell, line 33: we must remember to add to this file every time we add a new product

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
```diff
@@ -0,0 +1,321 @@
...
+        case (ids, transformFunction) if ids.contains(ratePlanItem.productRatePlanId) =>
+          Some(transformFunction(maybeAttributes.getOrElse(Attributes(identityId)), ratePlanItem))
+      }
+      .getOrElse(maybeAttributes)
+
+}
+
+object SupporterRatePlanToAttributesMapper {
+
+  val productRatePlanMappings: Map[String, Map[List[String], (Attributes, SupporterRatePlanItem) => Attributes]] =
```
johnduffell, line 27: this type is a little hard to digest - might be worth making the `(Attributes, SupporterRatePlanItem) => Attributes` part of the type into a single abstract method trait, and possibly adding type aliases for Stage and ProductRatePlanId

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
```diff
@@ -0,0 +1,321 @@
...
+      ) -> memberTransformer("Patron")
+    )
+  )
+
+  def digitalSubTransformer(attributes: Attributes, supporterRatePlanItem: SupporterRatePlanItem) =
+    attributes.copy(
+      DigitalSubscriptionExpiryDate = Some(
+        getLatestDate(attributes.DigitalSubscriptionExpiryDate, supporterRatePlanItem.termEndDate)
+      )
+    )
```
johnduffell, line 278: the SAM thing would be add a trait like this 
```scala
  trait SubscriptionTransformer {
    def transform(attributes: Attributes, supporterRatePlanItem: SupporterRatePlanItem): Attributes
  }
```
then make it a val and add the trait type to the val
```suggestion
  val digitalSubTransformer: SubscriptionTransformer = (attributes: Attributes, supporterRatePlanItem: SupporterRatePlanItem) =>
    attributes.copy(
      DigitalSubscriptionExpiryDate = Some(
        getLatestDate(attributes.DigitalSubscriptionExpiryDate, supporterRatePlanItem.termEndDate)
      )
    )

```

johnduffell (approved):
looks good, some comments to consider, but the only real thing is we would probably need an alarm for the Dynamo read failures, since they may not cause any 5xx errors.

cloudformation/membership-attribute-service.yaml
rupertbates: Fair point, I've changed it to notify me only

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: Good suggestion

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: I preferred `if (attributesDoNotMatch(fromZuora, fromSupporterProductData))` in the main logic flow but it's marginal

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: Yeah I'm in two minds about whether this is necessary now. I think I originally put the extra day in when I thought we would be running this job once a day but now we are running it every few minutes I'm not sure that it really adds anything and it does complicate things conceptually. What do you think?

membership-attribute-service/app/services/AttributesFromZuora.scala
johnduffell: I mean don't use `!` at all - you can just flip the if and else parts of the statement.

membership-attribute-service/app/models/SupporterRatePlanItem.scala
rupertbates: Have updated the name to make it clearer

membership-attribute-service/app/services/AttributesFromZuora.scala
rupertbates: Oh I see, I wasn't actually going to have an else clause originally because it could be very noisy in the logs, but then I wanted to be sure that the code was executing successfully so added it at the last moment

membership-attribute-service/app/services/SupporterProductDataService.scala
rupertbates: Yup, Intellij having fun

membership-attribute-service/app/services/AttributesFromZuora.scala
johnduffell: personally I would prefer not to have it, as it adds extra complexity.  We can always add it in later if we decide it's important.

membership-attribute-service/app/services/AttributesFromZuora.scala
johnduffell: fair enough, in that case leave it as it is.  This code won't be here long anyway.

PR #511, opened 2021-03-11
Add missing product rate plan ids to SupporterRatePlanToAttributesMapper

rupertbates:
### Why do we need this? 
The SupporterRatePlanToAttributesMapper class is missing some product rate plan ids which are still in use by active subscriptions. This PR adds them in.

### DIFF
--- a/membership-attribute-service/test/services/SupporterRatePlanToAttributesMapperTest.scala
+++ b/membership-attribute-service/test/services/SupporterRatePlanToAttributesMapperTest.scala
@@ -23,14 +23,14 @@ class SupporterRatePlanToAttributesMapperTest extends Specification {
       mapper.attributesFromSupporterRatePlans(
         identityId,
         List(ratePlanItem("2c92a0fc5aacfadd015ad24db4ff5e97"))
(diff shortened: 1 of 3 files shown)


membership-attribute-service/test/services/SupporterRatePlanToAttributesMapperTest.scala
```diff
@@ -242,47 +274,85 @@ object SupporterRatePlanToAttributesMapperTest {
   // This is a list of all active product rate plans from an extract taken
   // using the support-frontend/supporter-product-data project on 26th Feb 2021
```
johnduffell, line 275: or 11th of march?

johnduffell (approved):
looks good 👍 
probably still easier to maintain a long hard coded list like this, than maintain the lots of implicit memsub code in membership-common!

rupertbates:
> looks good 👍
> probably still easier to maintain a long hard coded list like this, than maintain the lots of implicit memsub code in membership-common!

Yes much easier I think and also much quicker to understand for incoming devs

PR #512, opened 2021-03-12
Update kinesis-logback-appender

tomrf1:
https://github.com/guardian/kinesis-logback-appender/pull/39

Tested in CODE

### DIFF
--- a/membership-attribute-service/app/loghandling/Logstash.scala
+++ b/membership-attribute-service/app/loghandling/Logstash.scala
@@ -1,16 +1,17 @@
 package loghandling
 
-import com.amazonaws.auth.AWSCredentialsProvider
+//import com.amazonaws.auth.AWSCredentialsProvider
 import com.amazonaws.util.EC2MetadataUtils
 import configuration.Config
 import play.api.Configuration
 import com.gu.aws.CredentialsProvider
 import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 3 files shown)


ripecosta (approved):
👍

membership-attribute-service/app/loghandling/Logstash.scala
```diff
@@ -1,20 +1,24 @@
...
 case class LogStashConf(enabled: Boolean,
   stream: String,
   region: String,
-  awsCredentialsProvider: AWSCredentialsProvider,
+  awsCredentialsProvider: AwsCredentialsProvider,
   customFields: Map[String, String])
 
 object Logstash extends StrictLogging {
 
+  private val CredentialsProvider = AwsCredentialsProviderChain.builder
```
tomrf1, line 17: we were previously getting this from the `membership-common` library.
But `kinesis-logback-appender` now expects the new aws sdk, so we need to create it differently.
Elsewhere we're still using the old aws sdk, but I don't want to start migrating that yet

johnduffell: yes, this would be good to get upgraaded as for one thing it would reduce the jackson databind issues, but I agree not in this PR!

johnduffell (approved):
thanks for updating this! 👍

PR #513, opened 2021-03-30
Make attributes lookup use the new datastore for 5% of traffic

rupertbates:
### Why do we need this? 
This PR Switches 5% of live attribute traffic to the new supporter product data datastore.

I have spent the last week or so ironing out discrepancies between the content access reported from Zuora vs the supporter product data Dynamo table and the next step is to start serving some real users from the new datastore to check for issues while running in production. 

### The changes 

- Remove old code related to comparing results
- Add a new function to randomly choose between Zuora and supporter product data when looking up attributes
- I have added a new integration test which will run a comparison between the two stores if given a list of identity ids - this is ignored by default as it requires aws credentials and so can't be run on TeamCity.

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -115,7 +115,7 @@ class AttributeController(
   private lazy val random = new Random
 
   private def getZuoraAttributes(identityId: String)(implicit request: AuthenticatedUserAndBackendRequest[AnyContent]) = {
-    if(random.nextInt(100) > 5) {
+    if(random.nextInt(100) >= 5) {
       log.info(s"Fetching attributes from Zuora for user $identityId")
       getAttributesWithConcurrencyLimitHandling(identityId)
     } else {


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -112,6 +111,19 @@ class AttributeController(
     )
   }
 
+  private lazy val random = new Random
+
+  private def getZuoraAttributes(identityId: String)(implicit request: AuthenticatedUserAndBackendRequest[AnyContent]) = {
+    if(random.nextFloat > 0.05) {
```
johnduffell, line 117: do you want it to be random each time or consistent across a given identity id?  Might help if someone is reporting issues as we know if they're "in" our "out"- but would lose the wider coverage.
```scala
    if(List("00", "01", "02", "03", "04").contains(identityId.substring(-2))) {
```

Also I think it will be okay, but in general I've had issues with float/double in java in the past so I default to int where possible
```suggestion
    if(random.nextInt(100) >= 5) {
```

membership-attribute-service/app/services/SupporterProductDataService.scala
```diff
@@ -17,11 +17,9 @@ import scala.concurrent.ExecutionContext
 class SupporterProductDataService(client: AmazonDynamoDBAsync, table: String, mapper: SupporterRatePlanToAttributesMapper)
   (implicit executionContext: ExecutionContext) {
 
-  implicit val jodaNumberFormat: DynamoFormat[LocalDate] =
-    DynamoFormat.coercedXmap[LocalDate, Long, IllegalArgumentException](epochSeconds => Instant.ofEpochSecond(epochSeconds).toDateTime.toLocalDate)(
-      // We need epoch time in seconds https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html
-      _.toDateTimeAtStartOfDay.getMillis / 1000
-    )
+  implicit val jodaStringFormat: AnyRef with DynamoFormat[LocalDate] = DynamoFormat.coercedXmap[LocalDate, String, IllegalArgumentException](LocalDate.parse)(
```
johnduffell, line 20: why AnyRef?

membership-attribute-service/app/services/AttributesFromZuora.scala
```diff
@@ -226,34 +218,5 @@ object AttributesFromZuora extends LazyLogging {
...
-        (fromZuora.map(_.Tier) == fromSupporterProductData.map(_.Tier) &&
-          fromZuora.map(_.RecurringContributionPaymentPlan) == fromSupporterProductData.map(_.RecurringContributionPaymentPlan) &&
-          datesAreEqualEnough(fromZuora.flatMap(_.DigitalSubscriptionExpiryDate), fromSupporterProductData.flatMap(_.DigitalSubscriptionExpiryDate)) &&
-          datesAreEqualEnough(fromZuora.flatMap(_.PaperSubscriptionExpiryDate), fromSupporterProductData.flatMap(_.PaperSubscriptionExpiryDate)) &&
-          datesAreEqualEnough(fromZuora.flatMap(_.GuardianWeeklySubscriptionExpiryDate), fromSupporterProductData.flatMap(_.GuardianWeeklySubscriptionExpiryDate))
-          )
-      )
-
-  def datesAreEqualEnough(fromZuora: Option[LocalDate], fromSupporterProductData: Option[LocalDate]) = {
-    //Currently dates from the supporter-product-data store will be a day after the zuora term end date to avoid
```
johnduffell, line 253: I think we removed the "free" day in the end did we?

membership-attribute-service/test/services/SupporterProductDataIntegrationTest.scala
```diff
@@ -0,0 +1,113 @@
...
+  implicit private val actorSystem: ActorSystem = ActorSystem()
+  lazy val touchpoint = new TouchpointComponents("PROD")
+  lazy val attributesFromZuora = new AttributesFromZuora()
+
+  args(skipAll = true) // This test requires credentials so won't run on CI, change skipAll to false to run locally
+
+  "SupporterProductData" should {
+    "match Zuora" in {
+      val excluded = List()
+      val allIds: List[String] = Source.fromURL(getClass.getResource("/identityIds.txt")).mkString.split("\n").toList.filter(!excluded.contains(_))
```
johnduffell, line 40: it's not risk of the century, but as a habit I'd rather keep the file outside the VCS root to avoid a chance of checking it in.
```suggestion
      val allIds: List[String] = Source.fromFile(new File("~/identityIds.txt")).mkString.split("\n").toList.filter(!excluded.contains(_))
```

membership-attribute-service/app/models/DynamoSupporterRatePlanItem.scala
```diff
@@ -3,8 +3,9 @@ package models
 import org.joda.time.LocalDate
 
 case class DynamoSupporterRatePlanItem(
+  subscriptionName: String, //Unique identifier for the subscription
   identityId: String, //Unique identifier for user
-  ratePlanId: String, //Unique identifier for this product purchase for this user
   productRatePlanId: String, //Unique identifier for the product in this rate plan
-  termEndDate: LocalDate //Date that this subscription term ends
+  termEndDate: LocalDate, //Date that this subscription term ends
+  contractEffectiveDate: LocalDate //Date that this subscription started
```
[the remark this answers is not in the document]
rupertbates: Yes I had to change the structure of the query and the data a bit

membership-attribute-service/app/services/SupporterProductDataService.scala
rupertbates: Not sure I think Intellij added that and I didn't notice. I've removed it now.

membership-attribute-service/test/services/SupporterProductDataIntegrationTest.scala
rupertbates: I've just removed it actually, I don't think it will be particularly useful going forward.

membership-attribute-service/app/services/AttributesFromZuora.scala
johnduffell: that's helpful to know, thanks

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -114,7 +115,7 @@ class AttributeController(
   private lazy val random = new Random
 
   private def getZuoraAttributes(identityId: String)(implicit request: AuthenticatedUserAndBackendRequest[AnyContent]) = {
-    if(random.nextFloat > 0.05) {
+    if(random.nextInt(100) > 5) {
```
johnduffell, line 118:
```suggestion
    if(random.nextInt(100) >= 5) {
```

johnduffell: it would be 0-99 inclusive

johnduffell (approved):
great, this will be the first use in real life which is our end goal.. Good idea to roll it out in proportions. 👍   Very good on ironing out all the discrepencies as that is definitiely an uphill battle.

PR #565, opened 2021-05-20
Increase Supporter Product Data use to 10%

rupertbates:
Increase the percentage of requests which are using the SupporterProductData Dynamo table to 10%


johnduffell (approved):
go for it!  let's crank it up regularly and gradually 👍

PR #567, opened 2021-05-21
Add some metrics to track the time it takes to fetch user attributes

rupertbates:
### Why do we need this? 
I want to get an idea of the average time it takes to serve a user attributes response from either Zuora or the SupporterProductData data store, this PR adds some metrics to track that.


johnduffell (approved):
looks good, although just to be sure, make sure this is not creating a fresh instnce of anything every request - I did this once by putting the metrics object in the method rather than the class, and blew the metrics quota and cost us a load of money!

PR #571, opened 2021-05-25
bump identity and scanamo for snyk http4s issues and cats incompatibility

johnduffell:
This pulls in the new identity lib to prevent the http4s issue hitting us.
Also I had to bump scanamo due to a cats binary incompatibility issue.  That was probably overdue anyway.

PR #582, opened 2021-06-03
Update identity-auth-play

tomrf1:
to get a more recent version of the http4s dependency. However, this will not solve all current snyk warnings for that dependency


johnduffell:
@tomrf1 I hve a PR for this aand I have been working with identity to get a new version of the lib released https://github.com/guardian/members-data-api/pull/571
So this PR may not be needed after all.

PR #590, opened 2021-06-10
Serve all Zuora attribute requests from the Supporter Product Data store

rupertbates:
### Why do we need this? 
Following on from #586 this PR moves the remaining traffic over from direct calls to Zuora to the supporter product data dynamo table.


paulbrown1982 (approved):
🥳

PR #633, opened 2021-10-15
Give recurring contributors access to digital pack

kelvin-chappell:


kelvin-chappell:
This is just an example of what we might do.

PR #635, opened 2021-10-27
Avoid double idapi calls from AttributeController

pvighi:
### Why do we need this? 
Members data api sets the `X-Gu-Identity-Id` and `X-Gu-Membership-Test-User` headers on every response with an http filter and some error handlers.

It seems this was added for fastly logging a while ago: https://github.com/guardian/members-data-api/pull/157/commits/9bc912142d4db67ce68d44bb5495c2030a475b8e. The problem is that since then, identity changed and now this requires an identity API call to get the user id and username.

The code that sets the response headers doesn't know if the user data was already retrieved during the execution of the call so  it always makes a new IDAPI call. This means that in many cases (as in calls to the most called endpoint:  `user-attributes/me`) the same Identity API endpoint is hit twice to fulfil a single members-data-api call.

In the identity team we are trying to reduce the amount of Identity api calls to `/user/me` as the current number will be too large when we move to Okta.

This PR just adds some code to make sure the identity headers are set whenever the user data is retrieved in the attribute controller so that in the http filter we can skip the extra identity call.

But currently, as far as I can tell any call to members-data-api made from an authenticated client will result in at least one identity api call to set the headers, this includes 404, 401 or other error responses.

### Further work after merging this pr
 - still some double api calls remain when the user is retrieved from other places, like the accounts controller. This PR only addresses the attribute controller since it is the most frequently used.
 - This quick fix will only remove some (hopefully most) unnecessary second calls to identity api,  but we could analyse how important these response headers are and if they are worth it now that they require an identity api call. 
 Maybe we could omit the one or both headers in some cases, or get the data from a more lightweight Identity call rather than getting all the user data (although getting the username could be a challenge)

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -222,7 +222,7 @@ class AttributeController(
 
       val userHasValidatedEmail = request.user.flatMap(_.statusFields.userEmailValidated).getOrElse(false)
 
-      if (userHasValidatedEmail) {
+      val futureResult: Future[Result] = if (userHasValidatedEmail) {
         request.user.map(_.id) match {
           case Some(identityId) =>
             contributionsStoreDatabaseService.getAllContributions(identityId).map {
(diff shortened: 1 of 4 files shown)


pvighi:
Ok after a few talks last week it seems like the header is being used in at least one place (to recognise test accounts in MMA).
So maybe the best way to go ahead is to merge this PR as an incremental improvement before until we are sure we can safely remove the code that adds the header from all responses. Another option would be to only return the headers when we already have the data, but not trigger an IDAPI call just to populate the headers

kelvin-chappell (approved):
👍

PR #637, opened 2021-11-02
Increasing threshold of 5xx error alarm

jambo1:
Increasing the 5XX alarm to greater than 2 periods of 50 errors per 5 minutes during a 20 minute period to reduce noise alerting.

5 minutes is the minimum period for "basic metrics" https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#:~:text=CloudWatch%20statistics%20definitions.-,Alarms,-You%20can%20use


kelvin-chappell (approved):
👍

PR #643, opened 2021-11-15
Give more useful error message on missing payment gateway

kelvin-chappell:
Instead of throwing a `NoSuchElementException` when a payment method has no payment gateway, this change gives us a more meaningful error which is easier to find in a log and easier to trace.

### DIFF
--- a/membership-attribute-service/app/controllers/PaymentUpdateController.scala
+++ b/membership-attribute-service/app/controllers/PaymentUpdateController.scala
@@ -112,7 +112,10 @@ class PaymentUpdateController(commonActions: CommonActions, override val control
       createPaymentMethod = CreatePaymentMethod(
         accountId = subscription.accountId,
         paymentMethod = bankTransferPaymentMethod,
-        paymentGateway = account.paymentGateway.get, // this will need to change to use this endpoint for 'payment method' SWITCH
+        paymentGateway = account.paymentGateway
+          .getOrElse(
+            throw new RuntimeException(s"Unrecognised payment gateway for account ${subscription.accountId} for user $maybeUserId")
+          ), // this will need to change to use this endpoint for 'payment method' SWITCH
         billtoContact = billToContact,
         invoiceTemplateOverride = None
       )
--- a/membership-attribute-service/app/controllers/PaymentUpdateController.scala
+++ b/membership-attribute-service/app/controllers/PaymentUpdateController.scala
@@ -109,16 +109,17 @@ class PaymentUpdateController(commonActions: CommonActions, override val control
         lastName = billToContact.lastName,
         countryCode = "GB"
       )
-      createPaymentMethod = CreatePaymentMethod(
-        accountId = subscription.accountId,
-        paymentMethod = bankTransferPaymentMethod,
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/controllers/PaymentUpdateController.scala
```diff
@@ -112,7 +112,10 @@ class PaymentUpdateController(commonActions: CommonActions, override val control
       createPaymentMethod = CreatePaymentMethod(
         accountId = subscription.accountId,
         paymentMethod = bankTransferPaymentMethod,
-        paymentGateway = account.paymentGateway.get, // this will need to change to use this endpoint for 'payment method' SWITCH
+        paymentGateway = account.paymentGateway
+          .getOrElse(
+            throw new RuntimeException(s"Unrecognised payment gateway for account ${subscription.accountId} for user $maybeUserId")
```
tomrf1, line 117: I appreciate this was already throwing an exception here, but is there any value in making it produce a `Left` instead?

johnduffell: I feel like the value would be that we could potentially handle it later, e.g. add the actual JSON/payment gateway raw value, to the error message further up the chain.
Also that the function signature of the function fully represents the possible results of running it, so easier to reason about/refactor in future without stray exceptions.

membership-attribute-service/app/controllers/PaymentUpdateController.scala
```diff
@@ -112,7 +112,10 @@ class PaymentUpdateController(commonActions: CommonActions, override val control
...
+          ), // this will need to change to use this endpoint for 'payment method' SWITCH
```
johnduffell, line 118: I don't quite understand this comment nd I wonder if it's been separated from the relevant code at some point?

johnduffell (approved):
it's a +1 for me on the basis of it's definitely better than before, but since we're already in an EitherT context anyway, I am strongly tempted by Tom's suggestion at your discretion

tomrf1 (approved):
👍 
side note - it would be nice to remove scalaz one day!

kelvin-chappell:
> 👍 side note - it would be nice to remove scalaz one day!

Hear hear!  But that could start a holy war.  Cats or ZIO or vanilla Scala or something else?

tomrf1:
> > 👍 side note - it would be nice to remove scalaz one day!
> 
> Hear hear! But that could start a holy war. Cats or ZIO or vanilla Scala or something else?

Using scala's `Either` would be a good start! It could get quite scary once you start unpicking...

kelvin-chappell:
> Using scala's `Either` would be a good start! It could get quite scary once you start unpicking...

That does seem feasible 🤔

johnduffell:
using scala's Either makes sense, but since this is EitherT we would still need a cats or something more exotic.
Regarding ZIO it sounds like it's very technically well thought out, but given our focus on inclusion I would want to be careful from a people point of view.  The scala community often seems to be a bit of a let-down in terms of being a safe space so I would prefer to sail on the better side of the wind where possible.

PR #674, opened 2022-01-12
Make digital subscription gifts appear in mdapi

rupertbates:
### Why do we need this? 
https://github.com/guardian/membership-common/pull/646 enables membership-common to load digital subscription gifts from Zuora. This PR updates members-data-api to use this version and also makes a small change to how we request  subscriptions from Zuora to ensure that they are returned in the first place.

https://trello.com/c/lsQSGoD0/163-investigate-gift-redemptions-not-showing-up-in-members-data-api

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -333,7 +333,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       // subscription so we can reuse it and avoid a call to Zuora
       nonGiftSubs.find(_.subscription.name == subscriptionName) match {
         case Some(contactAndSubscription) => Future.successful(Some(contactAndSubscription.subscription))
-        case _ => subscriptionService.get[AnyPlan](subscriptionName, isActiveToday = false)
+        case _ => subscriptionService.get[AnyPlan](subscriptionName, isActiveToday = false) //set isActiveToday to false so that we find digisub plans with a one time charge
       }
     }
     Future.sequence(all).map(_.flatten) // failures turn to None, and are logged, so just ignore them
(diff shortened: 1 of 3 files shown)


start-api-debug.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 cd $(dirname $0)
-sbt -Djava.awt.headless=true -jvm-debug 9997 "project membership-attribute-service" "devrun"
+sbt -mem 4096 -Djava.awt.headless=true -jvm-debug 9997 "project membership-attribute-service" "devrun"
```
rupertbates, line 4: Give sbt a bit more memory as it was running out

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -333,7 +333,7 @@ class AccountController(commonActions: CommonActions, override val controllerCom
       // subscription so we can reuse it and avoid a call to Zuora
       nonGiftSubs.find(_.subscription.name == subscriptionName) match {
         case Some(contactAndSubscription) => Future.successful(Some(contactAndSubscription.subscription))
-        case _ => subscriptionService.get[AnyPlan](subscriptionName, isActiveToday = true)
+        case _ => subscriptionService.get[AnyPlan](subscriptionName, isActiveToday = false)
```
kelvin-chappell, line 336: Would it be worth adding a comment to explain why this is false rather than true?

kelvin-chappell (approved):
👍

membership-attribute-service/app/controllers/AccountController.scala
rupertbates: Maybe, there is an extensive Javadoc comment on the method itself:
Screen Shot 2022-01-12 at 14 49 00

Do you think I should add something else?

rupertbates: Maybe, there is an extensive Javadoc comment on the method itself:
Screen Shot 2022-01-12 at 14 49 00

Do you think I should add something else?

rupertbates: I added something in.

kelvin-chappell: Thanks.  That helps to explain why it works for gifts, which I don't think is clear in the javadoc.  Although I hadn't read the javadoc anyway.

PR #691, opened 2022-02-25
Log exception fetching payment details

kelvin-chappell:
There seems to be a timeout when fetching payment details from Zuora for complicated subscriptions with many invoices and amendments.
It appears to happen in membership-common, either [here](https://github.com/guardian/membership-common/blob/main/src/main/scala/com/gu/memsub/services/PaymentService.scala#L34) or [here](https://github.com/guardian/membership-common/blob/main/src/main/scala/com/gu/memsub/services/PaymentService.scala#L47).
This is an attempt to isolate it before delving into membership-common.


graham228221 (approved):
Thanks Kelvin (no probs if you want approval from a scala dev before deploying)

graham228221:
[CSR bug report (severity: 3) - User sees "Oops" error from billing info in MMA](https://trello.com/c/U9T3qWd6/2910-csr-bug-report-severity-3-user-sees-oops-error-from-billing-info-in-mma)

PR #703, opened 2022-04-01
Increase read timeout for Zuora calls

kelvin-chappell:
The change is to increase the read timeout from default 10 s to 30 s.

This is to solve the problem where reading complex invoice histories and subscriptions with many amendments from Zuora can take a longer time and we're seeing timeouts in the logs.

The logging we've put in place should prove whether this works or not, in
* #691 
* guardian/membership-common#649

https://trello.com/c/U9T3qWd6/2910-userhelp-bug-report-severity-3-user-sees-oops-error-from-billing-info-in-mma


rupertbates (approved):
👍

PR #716, opened 2022-05-11
Update identity-auth-play to 3.254

rupertbates:
### Why do we need this? 
Upgrade identity-auth-play to 3.254 to mitigate a critical vulnerability in http4s see https://github.com/guardian/identity/pull/2106 for more info.


TomPretty (approved):
👍

PR #725, opened 2022-05-24
Include Guardian Patrons in mdapi user attributes response

rupertbates:
### Why do we need this? 
We want to be able to recognise Guardian Patrons on the website and apps so that we can give them the digital subscription benefits they are entitled to. 

This PR builds on [TODO: add support-frontend PR link]() which adds Patron data into the supporter-product-data DynamoDb table by making that data available through the user attributes endpoint.


paulbrown1982 (approved):
👍  LGTM. So much simpler now that we have the Supporter Product Data Attributes table!

PR #727, opened 2022-05-30
Remove old user attributes mechanism (zuora + cache)

rupertbates:
### Why do we need this? 
Since moving to the SupporterProductData DynamoDB store in #509 we have had a lot of unused code related to the old mechanism for retrieving supporter data. This PR removes that code and also allows us to delete the associated legacy dynamo tables from AWS.


build.sbt
```diff
@@ -35,14 +35,6 @@ val commonSettings = Seq(
   Test / javaOptions += "-Dconfig.resource=TEST.public.conf"
 ) ++ buildInfoSettings
 
-lazy val dynamoDBLocalSettings = Seq(
```
rupertbates, line 38: Another benefit is that we no longer need to run the local DynamoDB tests which were quite flakey

PR #728, opened 2022-05-30
Remove some more unused bits and pieces

rupertbates:
### Why do we need this? 
I found some more unused code following on from #727


kelvin-chappell (approved):
👍

PR #753, opened 2022-06-20
Get Scalafmt working and run scalafmtAll

rupertbates:
### Why do we need this? 
This PR switches on [scalafmt](https://scalameta.org/scalafmt/) for members-data-api (it already had a .scalafmt file but the plugin wasn't added to the project so I'm not sure what was going on with that).

Scalafmt integrates with both [Intellij and VS Code (via Metals)](https://scalameta.org/scalafmt/docs/installation.html#vs-code) and provides reformatting capabilities (similar to prettier) as well as style checking. It is also the tool used on frontend, so this is good from a consistency point of view. There is a doc with some analysis on the two tools [here](https://docs.google.com/document/d/1u5Wk8htF9MJN2hHpnPdVEd-qykqvZ06c06LOAonI0zo/edit)

I have also run the formatter on the entire repo, this has affected 63 files so it would take a very long time to check thoroughly but might be worth skimming.

One style point to note is that I have turned on trailing commas so from now on the convention will be for code such as the following:
```scala
    GetAddressIOConfig(
      config.getString("get-address-io-api.url"),
      config.getString("get-address-io-api.key"),
    )
```
Instead of 
```scala
    GetAddressIOConfig(
      config.getString("get-address-io-api.url"),
      config.getString("get-address-io-api.key")
    )
```
This is consistent with our style rules for Typescript and it also matches with what the Support site and Frontend do for Scala.


kelvin-chappell (approved):
👍

PR #756, opened 2022-06-24
Integrate repo with snyk GHA

NovemberTang:
This PR reliably integrates the repository with the snyk GitHub action. This PR has only been raised on repos that have already been tested to make sure scanning will work out of the box. ‘reliably integrated’ means that this action compares the hash of the last commit on main to the one that snyk has, and makes sure that they match.


rupertbates (approved):
Nice, thanks!

PR #762, opened 2022-07-06
Remove code to check Zuora card details against Stripe

rupertbates:


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -173,7 +168,6 @@ class AccountController(
       val maybeUserId = request.user.map(_.id)
 
       (for {
-        identityId <- EitherT.fromEither(Future.successful(maybeUserId.toRight(unauthorized)))
```
rupertbates, line 176: This was unused

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -241,7 +193,7 @@ class AccountController(
       def getPaymentMethod(id: PaymentMethodId) = tp.zuoraRestService.getPaymentMethod(id.get).map(_.toEither)
       val maybeUserId = request.user.map(_.id)
 
-      logger.info(s"Attempting to retrieve payment details for identity user: ${maybeUserId.mkString}")
+      logger.info(s"Deprecated function called: Attempting to retrieve payment details for identity user: ${maybeUserId.mkString}")
```
rupertbates, line 196: Just putting this log message in so we can see if this deprecated method is still called

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -254,9 +206,6 @@ class AccountController(
         paymentDetails <- OptionEither.liftOption(tp.paymentService.paymentDetails(\/.fromEither(freeOrPaidSub)).map(Right(_)).recover { case x =>
           Left(s"error retrieving payment details for subscription: ${sub.name}. Reason: $x")
         })
-        upToDatePaymentDetails <- OptionEither.liftOption(getUpToDatePaymentDetailsFromStripe(sub.accountId, paymentDetails).map(Right(_)).recover {
-          case x => Left(s"error getting up-to-date card details for payment method of account: ${sub.accountId}. Reason: $x")
-        })
```
johnduffell, line 259: the stripe information is synced to zuora by the stripe-source-updated lambda in support-service-lambdas?
That is not 100% reliable so the extra check would provide some value.  Although perhaps the extra check was introduced before the lambda?

Are you saying that it's reliable enough compared with the impact of getting it wrong in mma, that the extra cost of the call is not worth it?  If so I probably agree.  But I just want to confirm the logic used here.

johnduffell (approved):
good idea, just inclided some questions/comments but other than that I don't see any issue.  Great to see members-data-api getting some love! 👏

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -406,47 +355,64 @@ class AccountController(
...
               s"with account id ${contactAndSubscription.subscription.accountId}. Reason: $x",
           )
-        },
+        }
+
+    for {
+      paymentDetails <- ListEither.liftList(paymentDetailsFuture)
+      accountSummary <- ListEither.liftList(accountSummaryFuture)
+      effectiveCancellationDate <- ListEither.liftList(effectiveCancellationDateFuture)
+    } yield (paymentDetails, accountSummary, effectiveCancellationDate)
```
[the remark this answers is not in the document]
rupertbates: Yes, it should be fine now that we're not calling Zuora from the AttributeController any more

membership-attribute-service/app/controllers/AccountController.scala
rupertbates: I believe the check was put in place before the lambda existed (@paulbrown1982 could confirm). Yes to your second paragraph, the performance cost is not worth the benefit of picking up any inconsistencies.

PR #770, opened 2022-07-15
Ensuring the GoCardless gateway is set as default when updating Direct Debit details

paulbrown1982:
### Why do we need this? 
Before, the payment method update API was not changing the account's payment gateway, causing a misalignment between the default payment method and the gateway.

### The changes 
Ensure the GoCardless gateway is set as default when updating Direct Debit details.


jonflynng (approved):

johnduffell (commented):
thanks 👍 
there used to be a comment explaining that, but looks like you found what to do anyway https://github.com/guardian/members-data-api/commit/b03a35738fed752ae1e76de94676be8ff8b69841#diff-51be3bf42924f91138aeb210e429dd677197759ced0c609d68cda120318c2052R134

PR #774, opened 2022-07-20
The calculation for the end date in the /me/mma response was nonsence…

paulbrown1982:
### Why do we need this? 
Some situations (e.g. cancelling a Guardian Weekly sub before fulfilment is due to start) lead to a cancelled sub's start date being after the end date. See screenshot.

### The changes 
The calculation for the end date in the /me/mma response was nonsensical - it used the chargedThroughDate which has no relevance to the sub contract mechanic - and led to the end date being before the start date. I've changed it to use the same logic as the start date, maintaining the original fallback to the term end date.

I'll fix the duplicate entry on this page in a separate PR, whereupon I'll add a test too. This fix is a dependency.
179944347-37476330-060d-4e54-9be6-4fd70066f337

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -43,8 +43,6 @@ object AccountDetails {
 
       val mmaCategory = mmaCategoryFrom(product)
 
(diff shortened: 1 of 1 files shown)


rupertbates (approved):
👍

PR #780, opened 2022-08-08
Give recurring contributors SupporterPlus on apps

tomrf1:
Recurring contribution amount + currency [is being added to the supporter product data](https://github.com/guardian/support-frontend/pull/4189).

We have a requirement to exclude high-contributors from apps metering.
The apps will disable metering if the `supporterPlus` field is true in the response from MDAPI.

This PR changes MDAPI to set `supporterPlus` to true if:
- the request user-agent indicates it's from a mobile app
- the user has a recurring contribution with an amount above a certain threshold

### Low contributor:
Screenshot 2022-09-15 at 13 21 46

### High contributor:
Screenshot 2022-09-15 at 13 20 54

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -89,6 +89,12 @@ class AttributeController(
       .map(maybeAttributes => ("supporter-product-data", maybeAttributes.getOrElse(None)))
   }
 
+  private def isLiveApp(ua: String): Boolean = ua.matches("""^Guardian(News)?\/.*""")
+  private def upgradeRecurringContributorsOnApps(userAgent: Option[String], attributes: Attributes): Attributes =
+    if (attributes.isRecurringContributor && userAgent.exists(isLiveApp)) {
+      attributes.copy(SupporterPlusExpiryDate = Some(LocalDate.now().plusDays(1)))
+    } else attributes
+
   private def lookup(
       endpointDescription: String,
       onSuccessMember: Attributes => Result,
@@ -118,9 +124,9 @@ class AttributeController(
               supporterAttributes,
               user.id,
             )
-            allProductAttributes: Option[Attributes] = supporterOrStaffAttributes.map(
-              addOneOffAndMobile(_, latestOneOffDate, latestMobileSubscription),
-            )
+            allProductAttributes: Option[Attributes] = supporterOrStaffAttributes
+              .map(addOneOffAndMobile(_, latestOneOffDate, latestMobileSubscription))
+              .map(upgradeRecurringContributorsOnApps(request.headers.get(USER_AGENT), _))
           } yield {
 
             def customFields(supporterType: String): List[LogField] = List(
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -4,7 +4,6 @@ import actions._
 import akka.actor.ActorSystem
 import com.gu.i18n.Currency
 import com.gu.identity.model.User
-import com.gu.memsub.Price
 import com.gu.memsub.subsv2.SubscriptionPlan.AnyPlan
 import filters.AddGuIdentityHeaders
 import loghandling.LoggingField.{LogField, LogFieldString}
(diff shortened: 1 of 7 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -89,6 +89,12 @@ class AttributeController(
       .map(maybeAttributes => ("supporter-product-data", maybeAttributes.getOrElse(None)))
   }
 
+  private def isLiveApp(ua: String): Boolean = ua.matches("""^Guardian(News)?\/.*""")
+  private def upgradeRecurringContributorsOnApps(userAgent: Option[String], attributes: Attributes): Attributes =
+    if (attributes.isRecurringContributor && userAgent.exists(isLiveApp)) {
+      attributes.copy(SupporterPlusExpiryDate = Some(LocalDate.now().plusDays(1)))
```
tomrf1, line 95: Note - the apps don't yet use `supporterPlus`, so if we want this out quickly then change it to `digitalPack`

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -89,6 +91,23 @@ class AttributeController(
       .map(maybeAttributes => ("supporter-product-data", maybeAttributes.getOrElse(None)))
   }
 
+  private def isLiveApp(ua: String): Boolean = ua.matches("""^Guardian(News)?\/.*""")
+  private def isHighContributor(price: Price, isMonthly: Boolean): Boolean = {
```
tomrf1, line 95: TODO - use this function once we have the amount/currency available

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -89,6 +90,26 @@ class AttributeController(
       .map(maybeAttributes => ("supporter-product-data", maybeAttributes.getOrElse(None)))
   }
 
+  private def isLiveApp(ua: String): Boolean = ua.matches("""^Guardian(News)?\/.*""")
+  private def isHighContributor(contribution: ContributionAmount, isMonthly: Boolean): Boolean = {
+    val threshold = contribution.currency match {
+      case Currency.GBP | Currency.USD | Currency.EUR =>
```
rupertbates, line 96: Are these values finalised?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -89,6 +90,26 @@ class AttributeController(
...
+        if (isMonthly) 15 else 140
+    }
+    contribution.amount >= threshold
+  }
+  private def upgradeRecurringContributorsOnApps(userAgent: Option[String], attributes: Attributes): Attributes =
+    if (
+      userAgent.exists(isLiveApp) &&
+      attributes.RecurringContributionAmount.exists(amount => isHighContributor(amount, attributes.isMonthlyRecurringContributor))
+    ) {
+      attributes.copy(SupporterPlusExpiryDate = Some(LocalDate.now().plusDays(1)))
```
rupertbates, line 110: Are we expecting the effect of this date to be that the apps will check daily? Do you know how that compares to the current check frequency?

membership-attribute-service/app/controllers/AttributeController.scala
tomrf1: yes I got them from BRR

membership-attribute-service/app/controllers/AttributeController.scala
tomrf1: actually I'm not convinced this date matters at all. E.g. the android app [just checks a bool](https://github.com/guardian/android-news-app/blob/09f754ddb5872c0f6823f3262c03687358ae1647/android-news-app/src/main/java/com/guardian/feature/money/readerrevenue/SyncMembersDataApi.kt#L87) in the MDAPI response.

Only MDAPI cares about this date, when setting that bool: https://github.com/guardian/members-data-api/blob/main/membership-attribute-service/app/models/Attributes.scala#L53

rupertbates (approved):
👍

tomrf1:
I've pushed up a refactor after realising that the fields are not nested in dynamodb.
The check for `isHighContributor` is now done when mapping from `DynamoSupporterRatePlanItem` -> `Attributes`.
This simplifies the `Attributes` serialization. But we do still need a `highContributor` field in `Attributes`, because the controller needs to check the user-agent before setting `supporterPlus`

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -92,9 +93,10 @@ object Attributes {
     (__ \ "userId").write[String] and
       (__ \ "tier").writeNullable[String] and
       (__ \ "recurringContributionPaymentPlan").writeNullable[String] and
+      JsPath.writeNullable[Boolean].contramap[Option[Boolean]](_ => None) and // do not serialize highContributor
       (__ \ "oneOffContributionDate").writeNullable[LocalDate] and
       (__ \ "membershipJoinDate").writeNullable[LocalDate] and
-      (__ \ "supporterPlusExpiryDate").writeNullable[LocalDate] and
+      JsPath.writeNullable[LocalDate].contramap[Option[LocalDate]](_ => None) and // do not serialize supporterPlusExpiryDate
```
tomrf1, line 99: `supporterPlusExpiryDate` is not used, and we decided it's best to not include it in the response

PR #794, opened 2022-09-16
Pass "new card" Stripe key to Manage-Frontend in all cases

paulbrown1982:
### Why do we need this? 
Manage-frontend should ideally be told which public Stripe key to use for this customer rather than having to work it out for itself (which it is currently trying to do by evaluating the Delivery Country, which is incorrect). This change allows Manage-Frontend to defer to MDAPI like it does in other instances (e.g. payment failure).

### The changes 
Always include `stripePublicKeyForCardAddition`.
 
https://trello.com/c/X0ulSbI9 and https://github.com/guardian/manage-frontend/pull/874


rupertbates (approved):
👍

paulbrown1982:
Closing this as it's not safe as this property is used in other ways in manage-frontend. https://github.com/guardian/members-data-api/pull/795 deployed instead.

PR #795, opened 2022-09-16
Adding Billing Country to Account Details

paulbrown1982:
### Why do we need this? 
Adding Billing Country to Account Details as manage-frontend needs it to determine the correct Stripe Public Key to use, as it's currently only using the Delivery Country, which is wrong.

### The changes 
 Adding Billing Country value, or empty string, to Account Details.

https://trello.com/c/X0ulSbI9 and https://github.com/guardian/manage-frontend/pull/874

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -153,7 +153,6 @@ object AccountDetails {
           "shouldDisplayEmail" -> selfServiceCancellation.shouldDisplayEmail,
           "phoneRegionsToDisplay" -> selfServiceCancellation.phoneRegionsToDisplay,
         ),
-        "billingCountry" -> billingCountry.iterator.mkString,
       ) ++
         regNumber.fold(Json.obj())({ reg => Json.obj("regNumber" -> reg) }) ++
         billingCountry.fold(Json.obj())({ bc => Json.obj("billingCountry" -> bc.name)}) ++
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -153,14 +153,15 @@ object AccountDetails {
           "shouldDisplayEmail" -> selfServiceCancellation.shouldDisplayEmail,
           "phoneRegionsToDisplay" -> selfServiceCancellation.phoneRegionsToDisplay,
         ),
+        "billingCountry" -> billingCountry.iterator.mkString,
       ) ++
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -153,9 +153,11 @@ object AccountDetails {
           "shouldDisplayEmail" -> selfServiceCancellation.shouldDisplayEmail,
           "phoneRegionsToDisplay" -> selfServiceCancellation.phoneRegionsToDisplay,
         ),
+        "billingCountry" -> billingCountry.iterator.mkString,
       ) ++
         regNumber.fold(Json.obj())({ reg => Json.obj("regNumber" -> reg) }) ++
-        Json.obj(
+        billingCountry.fold(Json.obj())({ bc => Json.obj("billingCountry" -> bc.name)}) ++
```
rupertbates, line 158: Are we adding billiingCountry twice, once as a string and once as a json object?

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -153,9 +153,11 @@ object AccountDetails {
           "shouldDisplayEmail" -> selfServiceCancellation.shouldDisplayEmail,
           "phoneRegionsToDisplay" -> selfServiceCancellation.phoneRegionsToDisplay,
         ),
+        "billingCountry" -> billingCountry.iterator.mkString,
```
rupertbates: I think this will rely on `toString` won't it? Might be better to explicitly state which field we want eg.
```scala
"billingCountry" -> billingCountry.map(_.alpha2).getOrElse(""),
```

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -155,6 +155,7 @@ object AccountDetails {
         ),
       ) ++
         regNumber.fold(Json.obj())({ reg => Json.obj("regNumber" -> reg) }) ++
+        billingCountry.fold(Json.obj())({ bc => Json.obj("billingCountry" -> bc.name)}) ++
```
rupertbates, line 158: Is name definitely what we want here, not `alpha2`?

paulbrown1982: Yes, this makes it similar to `deliveryAddress.country` pulled in from Salesforce. Manage-frontend uses the country name to evaluate the Stripe token: https://github.com/guardian/manage-frontend/blob/42a23b000f95c4d5f273e3d807cfce2c7c1c6f5d/client/stripe.ts#L10

rupertbates (approved):
👍

PR #799, opened 2022-09-28
Simplify User model

kelvin-chappell:
This is enabling work for the movement of the authentication mechanism from Idapi to Okta.
The [User](https://github.com/guardian/identity/blob/main/identity-model/src/main/scala/com/gu/identity/model/Model.scala#L51-L83) model is very complicated and uses all sorts of fields that we don't intend to make available in Okta Access tokens.
So, here we replace it with a simpler local model: [AccessClaims](https://github.com/guardian/members-data-api/compare/kc/user?expand=1#diff-590be6f5b78842588c6a40afa9eebf29506732d21acaa2d05ecefb064fb3a4a0R14-R19).

Open to better naming of course.

https://trello.com/c/NLwIqfd7/4163-create-new-minimal-user-type-in-mdapi

Tested read and write endpoints in Code, signed in and signed out.

### DIFF
--- a/membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
+++ b/membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
@@ -1,8 +1,8 @@
 package actions
 
-import com.gu.identity.model.User
 import components.{TouchpointBackends, TouchpointComponents}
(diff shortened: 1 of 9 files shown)


kelvin-chappell:
Superseded by #801

PR #801, opened 2022-10-05
Enable authorisation either by Idapi or by Okta

pvighi:
This PR integrates [v4.3](https://github.com/guardian/identity/releases/tag/v4.3) and [v4.4](https://github.com/guardian/identity/releases/tag/v4.4) of the Identity libraries, which enable authorisation by Okta or, if an Okta access token is not available, falls back to Idapi.

It will have no immediate effect but will allow us to slowly move over, client by client, to authorisation by Okta.

The main changes in this PR are:
1. The introduction of a new [AccessClaims](https://github.com/guardian/members-data-api/pull/801/files#diff-590be6f5b78842588c6a40afa9eebf29506732d21acaa2d05ecefb064fb3a4a0R17-R22) class that replaces the previous `User` class.  And a [means of constructing](https://github.com/guardian/members-data-api/pull/801/files#diff-590be6f5b78842588c6a40afa9eebf29506732d21acaa2d05ecefb064fb3a4a0R24-R45) `AccessClaims`.
1. [Configuration for verifying Okta tokens](https://github.com/guardian/members-data-api/pull/801/files#diff-3b27f4a16a1c9f6bf5743f7e9447a8bf6ab1ec384bf8db1e2724a4013ef00dd7R56-R61).
1. Each endpoint that's authorised by `AuthAndBackendViaAuthLibAction` now has a set of [required scopes](https://github.com/guardian/members-data-api/pull/801/files#diff-97a990dc96d61a4ee519db5b8d3d767219005e2119cdcbea79d82607101bb599R11) matching the agreed scopes in [this sheet](https://docs.google.com/spreadsheets/d/17Zo1aVBV6yzOcVp3hC_nnYcwrt5PMsSrgInGY9D7i_A/edit#gid=0).  These scopes are only applied when an Okta access token is in use.
1. We [log](https://github.com/guardian/members-data-api/pull/801/files#diff-6bedd5db474b88212b823ca31365d6dc4e3dad606c968bb29c86a3ee9e1edc47R59) the method of authorisation to see progress.

Endpoints authorised by `AuthAndBackendViaIdapiAction` are going to be dealt with separately.


membership-attribute-service/app/services/IdentityAuthService.scala
```diff
@@ -1,41 +1,93 @@
...
   }
 
-  private def getUser(requestHeader: RequestHeader): Future[User] =
-    identityPlayAuthService
-      .getUserFromRequest(requestHeader)
-      .map { case (_, user) => user }
-      .unsafeToFuture()
-
+  private def getUser(requestHeader: RequestHeader): Future[Option[User]] = {
+    val scopes = List("profile", "email")
```
pvighi, line 48: this is just an example in the real version we would have members-data-api specific scopes here, and we probably will have to make different calls with different scopes for different endpoints

build.sbt
```diff
@@ -27,6 +27,7 @@ val commonSettings = Seq(
     "Guardian Github Releases" at "https://guardian.github.io/maven/repo-releases",
     "Guardian Github Snapshots" at "https://guardian.github.io/maven/repo-snapshots",
     Resolver.sonatypeRepo("releases"),
+    Resolver.sonatypeRepo("snapshots") //TODO remove before merging!
```
pvighi, line 30: like the comment says we should remove this when we are ready to merge. This was just added in order to be able to use a snapshot version of the identity-play-auth library but we don't want to risk leaving snapshot dependencies in prod

membership-attribute-service/app/models/AccessClaims.scala
```diff
@@ -0,0 +1,41 @@
...
+  *   User name, a synonym for display name. Also used to determine if this is a test user.
+  * @param userEmailValidated
+  *   true iff the user has validated their email address
+  */
+case class AccessClaims(
+                         primaryEmailAddress: String,
+                         identityId: String,
+                         username: Option[String] = None,
+                         userEmailValidated: Option[Boolean] = None,
+) extends UserClaims
```
pvighi, line 22: It might be that `userEmailValidated` in needed in multiple places and it makes sense to it to the default access claims defined in the library.  I added them here because 1) it allows me to demonstrate how to use custom claims defined outside the library and 2) it's a mutable field that could get out of date so maybe we should avoid getting it from the claims as much as possible.

as a semi related comment we should really think about the consequences of reading a stale field in each case. It might make sense to force the user to renew their access token after updating their email or setting their username or things like that. Although it probably won't be possible or at least easy to do across devices

project/Dependencies.scala
```diff
@@ -7,7 +7,7 @@ object Dependencies {
   val awsClientV2Version = "2.16.86"
 
   val sentryLogback = "io.sentry" % "sentry-logback" % "1.7.5"
-  val identityAuth = "com.gu.identity" %% "identity-auth-play" % "3.255"
+  val identityAuth = "com.gu.identity" %% "identity-auth-play" % "3.256.03-SNAPSHOT"
```
pvighi, line 10: this is just using the snapshot version of the library for testing in `CODE` (the `.03` at the end of the library version is because I released 3 additional snapshot versions to test but that should not affect the version number of the library so I didn't push the version change in the identity PR as the released version should just be `3.256`)

membership-attribute-service/app/services/IdentityAuthService.scala
```diff
@@ -1,41 +1,55 @@
...
       }
   }
 
-  private def getUser(requestHeader: RequestHeader): Future[User] =
-    identityPlayAuthService
-      .getUserFromRequest(requestHeader)
-      .map { case (_, user) => user }
-      .unsafeToFuture()
-
+  private def getUser(requestHeader: RequestHeader): Future[Option[AccessClaims]] = {
```
kelvin-chappell, line 47: Will need scopes to be sent through from client app:
```suggestion
  private def getUser(requestHeader: RequestHeader, requiredScopes: List[String] = List.empty): Future[Option[AccessClaims]] = {
```

membership-attribute-service/app/services/IdentityAuthService.scala
kelvin-chappell: These will be passed through from client apps

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -110,6 +112,11 @@ class TouchpointComponents(stage: String)(implicit system: ActorSystem, executio
   lazy val paymentService = new PaymentService(zuoraService, catalogService.unsafeCatalog.productMap)
 
   lazy val idapiService = new IdapiService(tpConfig.idapi, RequestRunners.futureRunner)
-  lazy val identityAuthService = new IdentityAuthService(tpConfig.idapi)
+  //todo maybe this should be initialized elsewhere
```
kelvin-chappell, line 115: I agree.  Perhaps a new `IdentityComponents` or something

membership-attribute-service/app/models/AccessClaims.scala
```diff
@@ -0,0 +1,41 @@
...
+  * @param primaryEmailAddress
+  *   primary email address
+  * @param identityId
+  *   Identity ID
+  * @param username
+  *   User name, a synonym for display name. Also used to determine if this is a test user.
+  * @param userEmailValidated
+  *   true iff the user has validated their email address
+  */
+case class AccessClaims(
```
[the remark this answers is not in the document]
kelvin-chappell: Can leave as-is for now I think.

membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
```diff
@@ -1,21 +1,22 @@
...
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class AuthAndBackendViaAuthLibAction(touchpointBackends: TouchpointBackends)(implicit ex: ExecutionContext)
+class AuthAndBackendViaAuthLibAction(touchpointBackends: TouchpointBackends, requiredScopes: List[AccessScope])(implicit ex: ExecutionContext)
     extends ActionRefiner[Request, AuthenticatedUserAndBackendRequest] {
   override val executionContext = ex
 
   override protected def refine[A](request: Request[A]): Future[Either[Result, AuthenticatedUserAndBackendRequest[A]]] = {
     // On each request via this action, we make a call to IDAPI and see if we can authenticate the user.
```
rupertbates, line 16: Is this comment still accurate?

membership-attribute-service/app/models/AccessClaims.scala
```diff
@@ -0,0 +1,45 @@
...
```
rupertbates, line 17: I'm not sure about calling this class `AccessClaims` it seems very confusing for a dev coming to this code without background knowledge of how Okta works. Also we have variables scattered around the codebase now like `val user: AccessClaims` which is non-idiomatic and distracting. Could we call this `User` now that we don't have a user class from `com.gu.identity.model` anymore?

Mark-McCracken (approved):
Looks good to me!

rupertbates (changes requested):
I've added a comment about the naming of the `AccessClaims` class

membership-attribute-service/app/models/AccessClaims.scala
pvighi: I'm not opposed to a rename but I think it's important to make it clear that this is data included within the token itself when it was created not fresh data coming from the server side. It might not be up to date depending on what happened since then

rupertbates: Yeah I see what you mean, maybe we can call it something like `UserTokenData` or `UserFromToken`?

kelvin-chappell: `UserFromToken` is a good suggestion.  Let's go with that.

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -37,13 +35,11 @@ class AttributeController(
   lazy val metrics = Metrics("AttributesController")
   lazy val expensiveMetrics = new ExpensiveMetrics("AttributesController")
 
-  private def getLatestOneOffContributionDate(identityId: String, user: User)(implicit
-      executionContext: ExecutionContext,
+  private def getLatestOneOffContributionDate(identityId: String, user: UserFromToken)(implicit
+                                                                                       executionContext: ExecutionContext,
```
rupertbates, line 39: ?

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -242,16 +248,16 @@ class AttributeController(
   /** Allow all validated guardian.co.uk/theguardian.com email addresses access to the digipack
     */
   private def maybeAllowAccessToDigipackForGuardianEmployees(
-      maybeUser: Option[User],
-      maybeAttributes: Option[Attributes],
-      identityId: String,
+                                                              maybeUser: Option[UserFromToken],
```
rupertbates, line 251: more formatting weirdness

rupertbates (approved):
Looks great, just a few formatting oddities that have crept in

kelvin-chappell:
> Looks great, just a few formatting oddities that have crept in

Thanks Rupert.  It looks like bulk renaming had some kind of inconsistent effect in Intellij.  But I thought bad formatting would fail CI.  I'll have a look at that.

PR #815, opened 2022-11-01
Return isRecurringSupporter field

tomrf1:
The existing `showSupportMessaging` field no longer serves the needs of RRCP targeting.
We need to separate single contributors from other types of support.

The new `isRecurringSupporter` enables us to do this. It's different from `showSupportMessaging` in that it does not include single contributors (unless they currently support us in some other way)

### DIFF
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -94,7 +94,7 @@ case class Attributes(
       || isGuardianWeeklySubscriber
       || isPremiumLiveAppSubscriber
       || isGuardianPatron
-    )
+  )
 
 }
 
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/models/DynamoSupporterRatePlanItem.scala
```diff
@@ -5,7 +5,7 @@ import org.joda.time.LocalDate
 import play.api.libs.json.{Writes, __}
 
 object DynamoSupporterRatePlanItem {
-    implicit val currencyWrite: Writes[Currency] = __.write[String].contramap(_.iso)
+  implicit val currencyWrite: Writes[Currency] = __.write[String].contramap(_.iso)
```
tomrf1, line 8: it insisted on a bunch of formatting changes which have nothing to do with my actual change...

PR #817, opened 2022-11-07
Rename secure auth scopes

kelvin-chappell:
We've updated some of the scope names to make it easier to determine which need special treatment during the authorisation process.  The scope names available to the API have also been brought down into this repo rather than being read from the Identity library to avoid having to release new versions of the Identity library whenever a new scope is added to an API.


pvighi (approved):
looks good, but I still think it might be nicer to have different traits for secure and non secure scopes so that we can match when deciding on how to authenticate instead of having to look into the string

PR #826, opened 2022-11-10
Currency and Currency ISO objects added to subscription body 

charleycampbell:
In this commit new currency and currency iso objects added to subscription body. This is the first step to removing the plan object from the account details

### Why do we need this? 
This is the first step to removing the plan object from the API response body to MMA

### The changes 
Currency and CurrencyISO objects have been copied into the subscription body.

https://trello.com/c/Jb7VYDSc/3592-remove-plan-property-from-members-data-api-response


KentGrm:
Closing this pull request as after reviewing the surrounding code, removing plan from the API isn't solving a problem but moving the problem deeper into the api codebase. i.e., somewhere a membership plan needs to be mapped to a currencyGlyph/ISO value for the front end. It's better to leave this thread and remove once memberships are removed as a whole.

PR #832, opened 2022-11-15
Filter out cancelled Patrons subs when generating user attributes

rupertbates:
### Why do we need this? 
Up until recently Guardian Patrons which were cancelled were not removed from the SupporterProductData dynamo table until the end of the current term meaning that they would continue to receive benefits even after they had cancelled. Since [this PR](https://github.com/guardian/support-frontend/pull/4298) however cancelled Patrons have a cancellation date set in the data store. 

This PR filters out cancelled Patrons from the user attributes endpoint at `/user/me`

### DIFF
--- a/membership-attribute-service/app/services/SupporterProductDataService.scala
+++ b/membership-attribute-service/app/services/SupporterProductDataService.scala
@@ -5,7 +5,7 @@ import com.gu.i18n.Currency
 import com.typesafe.scalalogging.LazyLogging
 import models.{Attributes, DynamoSupporterRatePlanItem}
 import monitoring.Metrics
-import org.joda.time.LocalDate
+import org.joda.time.{DateTimeZone, LocalDate}
 import org.scanamo.DynamoReadError.describe
 import org.scanamo.{DynamoReadError, _}
 import org.scanamo.generic.semiauto._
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/services/SupporterProductDataService.scala
```diff
@@ -25,8 +25,12 @@ class SupporterProductDataService(client: DynamoDbAsyncClient, table: String, ma
     DynamoFormat.xmap[Currency, String](s => Currency.fromString(s).toRight(TypeCoercionError(new Throwable("Invalid currency"))), _.iso)
   implicit val dynamoSupporterRatePlanItem: DynamoFormat[DynamoSupporterRatePlanItem] = deriveDynamoFormat
 
-  def getAttributes(identityId: String): Future[Either[String, Option[Attributes]]] =
-    getSupporterRatePlanItems(identityId).map(ratePlanItems => mapper.attributesFromSupporterRatePlans(identityId, ratePlanItems)).value
+  def getNonCancelledAttributes(identityId: String): Future[Either[String, Option[Attributes]]] = {
+    getSupporterRatePlanItems(identityId).map { ratePlanItems =>
+      val nonCancelled = ratePlanItems.filter(_.cancellationDate.isEmpty)
```
paulbrown1982, line 30: I think it would be safer w/r potential future changes, to test that the cancellationDate is today or in the past.

paulbrown1982 (approved):
👍 LGTM

PR #835, opened 2022-11-17
Bump Identity library

kelvin-chappell:
This brings in the auth server validation of sensitive endpoints in guardian/identity#2220


kelvin-chappell:
> looks good to me as long as this has been tested in CODE. Might be worth checking it works with apps too

Tested in Code and I'll check with apps after the next PR is on Code so they don't have to keep going round the same testing loop.

PR #847, opened 2022-12-01
Give correct response for different types of authorisation failure

kelvin-chappell:
For different types of authorisation failure.


membership-attribute-service/app/actions/CommonActions.scala
```diff
@@ -32,7 +33,7 @@ class CommonActions(touchpointBackends: TouchpointBackends, bodyParser: BodyPars
 class BackendRequest[A](val touchpoint: TouchpointComponents, request: Request[A]) extends WrappedRequest[A](request)
 
 class AuthenticatedUserAndBackendRequest[A](
-    val user: Option[UserFromToken],
+    val user: Either[AuthenticationFailure, UserFromToken],
```
pvighi, line 36: I'm a bit confused as to why the authenticated request has an optional user here

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -86,7 +86,7 @@ class AccountController(
     AuthAndBackendViaAuthLibAction(requiredScopes = List(readSelf, updateSelf)).async { implicit request =>
       val tp = request.touchpoint
       val cancelForm = Form { single("reason" -> nonEmptyText) }
-      val maybeUserId = request.user.map(_.identityId)
+      val maybeUserId = request.user.toOption.map(_.identityId)
```
pvighi, line 89: also confused why we can discard the error message here

pvighi (approved):
looks good. 
Might be good to add some comments to the places we have the user data in options just for idapi authentication

membership-attribute-service/app/actions/CommonActions.scala
kelvin-chappell: This is just preserving the type that's returned from the Identity lib so as not to break any other clients that are calling it.

membership-attribute-service/app/controllers/AccountController.scala
kelvin-chappell: At this point, authorisation has already been given so the `user` instance is just being used for the purposes of the endpoint.  If it wants to throw an error later it can.

PR #850, opened 2022-12-05
Revert: Give recurring contributors SupporterPlus on apps

tomrf1:
A [previous PR](https://github.com/guardian/members-data-api/pull/780) gave recurring contributors with an amount above a certain threshold Supporter+ access in the app.
We no longer want this feature.

I've kept the contributionAmount and currency in the `DynamoSupporterRatePlanItem` model, though they're no longer being used.

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -86,12 +86,6 @@ class AttributeController(
       .map(maybeAttributes => ("supporter-product-data", maybeAttributes.getOrElse(None)))
   }
 
-  private def isLiveApp(ua: String): Boolean = ua.matches("""^Guardian(News)?\/.*""")
-  private def upgradeRecurringContributorsOnApps(userAgent: Option[String], attributes: Attributes): Attributes =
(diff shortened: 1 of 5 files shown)


membership-attribute-service/app/models/DynamoSupporterRatePlanItem.scala
```diff
@@ -15,6 +9,4 @@ case class DynamoSupporterRatePlanItem(
     termEndDate: LocalDate, // Date that this subscription term ends
     contractEffectiveDate: LocalDate, // Date that this subscription started
     cancellationDate: Option[LocalDate], // If this subscription has been cancelled this will be set
-    contributionAmount: Option[BigDecimal],
```
rupertbates, line 18: I wonder if we should leave this in? The values exist in the data store so it would make sense to make them available here even if we are no longer using them for the current purpose. Or do you think it would be better to remove them from the data store as well?

tomrf1: good point, I don't think it hurts to leave the fields in even if they're unused. Maybe we'll use them one day...

rupertbates (approved):
👍

PR #851, opened 2022-12-07
Support Okta authorisation for all endpoints

kelvin-chappell:
This finishes off the remaining endpoints that use the `AuthAndBackendViaIdapiAction` for authorisation so that all endpoints now support Okta authorisation.  While we're in transition, Idapi cookies are also supported for all endpoints.

The strategy of this change is to delegate requests that have an Okta token to the `AuthAndBackendViaAuthLibAction`.  This is safe to do because the lifecycle of the Okta token will be managed in the client app rather than in this API so the check for the age of the token is redundant.  For example, MMA will have short-lived non-refreshable tokens and will respond to a 401 response by re-authenticating and requesting a new token.

As requests to this API authorised by Idapi don't use the `authorization` header, it seems reasonable to assume that as far as this API is concerned that header is only ever used for Okta tokens.  So the presence of the header is used as the indicator of an Okta token.

This change won't affect the existing behaviour of the API with Idapi cookies.

The expected behaviour is:
1. With an Idapi cookie, no change.
2. With an Okta token in the authorization header:
    1. If token is invalid: return 401
    2. If token doesn't have required scopes for endpoint: return 403
    3. Otherwise action goes ahead

The scopes for each endpoint are those agreed in [this sheet](https://docs.google.com/spreadsheets/d/17Zo1aVBV6yzOcVp3hC_nnYcwrt5PMsSrgInGY9D7i_A/edit#gid=0).

Tested on Code with Idapi cookies and Okta tokens.

### DIFF
--- a/membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
+++ b/membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
@@ -1,9 +1,10 @@
 package actions
 
 import com.gu.identity.auth.AccessScope
-import components.TouchpointBackends
+import components.{TouchpointBackends, TouchpointComponents}
 import filters.AddGuIdentityHeaders
-import play.api.mvc.{ActionRefiner, Request, Result, Results}
+import models.UserFromToken
+import play.api.mvc.{ActionRefiner, Request, Result, Results, WrappedRequest}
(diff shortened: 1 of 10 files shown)


kelvin-chappell:
Closing in favour of a new approach, which is on its way.

PR #863, opened 2022-12-16
AttributeController Acceptance Test

pawel-krupinski-guardian:
Test that:
- Starts the service programmatically (inside a Play Server)
- Mocks responses of external http calls (like identity service)
- Interacts with membership-data-api service by calling its REST endpoint(s) and asserting on the response

To make this test possible I replaced many references to configuration.Config fields, which is a singleton that loads config (and therefore doesn't allow overriding)  with usage of an instance of Typesafe Config (that comes from Play configuration).

TODO (in the future): Use embedded Postgres and Dynamo instead of mocks.


membership-attribute-service/test/acceptance/AttributeControllerAcceptanceTest.scala
```diff
@@ -0,0 +1,281 @@
...
+      (json \ "contentAccess" \ "paidMember").as[Boolean] shouldEqual false
+      (json \ "contentAccess" \ "recurringContributor").as[Boolean] shouldEqual false
+      (json \ "contentAccess" \ "supporterPlus").as[Boolean] shouldEqual false
+      (json \ "contentAccess" \ "digitalPack").as[Boolean] shouldEqual true
+      (json \ "contentAccess" \ "guardianWeeklySubscriber").as[Boolean] shouldEqual false
+      (json \ "contentAccess" \ "guardianPatron").as[Boolean] shouldEqual false
+    }
+  }
+
+  val identityResponse =
```
rupertbates, line 106: Do we get this full reponse from Identity currently?

project/Dependencies.scala
```diff
@@ -44,6 +46,8 @@ object Dependencies {
     "com.okta.jwt" % "okta-jwt-verifier" % oktaJwtVerifierVersion,
     "com.okta.jwt" % "okta-jwt-verifier-impl" % oktaJwtVerifierVersion,
   )
+  val unirest = "com.konghq" % "unirest-java" % "4.0.0-RC2" % Test
+  val okHttp = "com.squareup.okhttp3" % "okhttp" % "5.0.0-alpha.10"
```
rupertbates, line 50: Do we need to use an alpha version of okhttp?

membership-attribute-service/test/acceptance/AttributeControllerAcceptanceTest.scala
pawel-krupinski-guardian: Yeah I coped it from a real response.

project/Dependencies.scala
pawel-krupinski-guardian: It's to resolve a dependency conflict.
It's a transitive dependency and I think sbt might be already using that version.
Will double check.

pawel-krupinski-guardian: Ok, explicit dependency (override) wasn't needed, just an exclusion.

project/Dependencies.scala
```diff
@@ -44,6 +46,8 @@ object Dependencies {
     "com.okta.jwt" % "okta-jwt-verifier" % oktaJwtVerifierVersion,
     "com.okta.jwt" % "okta-jwt-verifier-impl" % oktaJwtVerifierVersion,
   )
+  val unirest = "com.konghq" % "unirest-java" % "4.0.0-RC2" % Test
+//  val okHttp = "com.squareup.okhttp3" % "okhttp" % "5.0.0-alpha.10"
```
rupertbates, line 50: Can you remove the comment please?

rupertbates (approved):
Just one little nitpick, but good work! 👍

project/Dependencies.scala
pawel-krupinski-guardian: Fixed now

PR #865, opened 2022-12-19
Removing configuration.Config

pawel-krupinski-guardian:
Removing all dependencies on configuration.Config singleton that loads config independently of Play.
Replacing them with usage of config loaded by Play.

Removing configuration.Config as it increases complexity and makes things harder to test (unoverridable side-effect of loading config from files in a singleton).


rupertbates (approved):
👍

PR #872, opened 2023-01-04
Refactor names of response properties

jonflynng:
Refactors the names of API response properties to be more in-line with Zuora's naming conventions (specifically the conventions in their REST API). That is, we are changing the property `amount` to be named `price`, and the property `interval` to be named `billingPeriod`. Refer to some examples of Zuora rest API responses [here](https://github.com/guardian/price-migration-engine/blob/main/lambda/src/test/resources/GuardianWeekly/CappedPriceIncrease/Subscription.json) and [here](https://github.com/guardian/price-migration-engine/blob/main/lambda/src/test/resources/GuardianWeekly/CappedPriceIncrease/Catalogue.json).

PR for the frontend [here](https://github.com/guardian/manage-frontend/pull/947).

### DIFF
--- a/membership-attribute-service/app/models/ContributionData.scala
+++ b/membership-attribute-service/app/models/ContributionData.scala
@@ -17,7 +17,7 @@ object ContributionData {
     override def writes(o: ContributionData): JsValue = Json.obj(
       "created" -> o.created,
       "currency" -> o.currency.toString,
-      "amount" -> o.amount,
+      "price" -> o.amount,
       "status" -> o.status,
     )
   }
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -111,10 +111,10 @@ object AccountDetails {
         case paidPlan: PaidSubscriptionPlan[_, _] =>
           Json.obj(
             "chargedThrough" -> paidPlan.chargedThrough,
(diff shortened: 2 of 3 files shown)


membership-attribute-service/app/models/ContributionData.scala
```diff
@@ -17,7 +17,7 @@ object ContributionData {
     override def writes(o: ContributionData): JsValue = Json.obj(
       "created" -> o.created,
       "currency" -> o.currency.toString,
-      "amount" -> o.amount,
+      "price" -> o.amount,
```
jonflynng, line 20: could keep this as `amount` as we use that convention in manage.theguardian.com for contributions

charleycampbell: I'd be happy for it to remain as is but for consistency it might make sense to change it?

charleycampbell (commented):
Looks good, I've replied to your open comment!

jonflynng:
[Rename object properties to be in-line with Zuora's naming](https://trello.com/c/ZfJMciLT/3737-rename-object-properties-to-be-in-line-with-zuoras-naming)

membership-attribute-service/app/models/ContributionData.scala
jonflynng: yep will keep it changed

charleycampbell (approved):

PR #875, opened 2023-01-04
Add user details to /me/mma

pawel-krupinski-guardian:
Adding user details (first name, last name and email) to /me/mma and /me/mma/:subscriptionName responses.
Moving array from top level to a field in the response object.
Some refactorings.

Delivering this task:
https://trello.com/c/DqAqj1bn/3630-add-user-details-to-members-data-api-response


membership-attribute-service/app/actions/CommonActions.scala
```diff
@@ -22,24 +22,30 @@ class CommonActions(touchpointBackends: TouchpointBackends, bodyParser: BodyPars
...
     override val executionContext = ex
     def invokeBlock[A](request: Request[A], block: (Request[A]) => Future[Result]) = block(request).map(f)
   }
 }
 
-class BackendRequest[A](val touchpoint: TouchpointComponents, request: Request[A]) extends WrappedRequest[A](request)
-
 class AuthenticatedUserAndBackendRequest[A](
-    val user: Either[AuthenticationFailure, UserFromToken],
+    val user: UserFromToken,
```
kelvin-chappell, line 46: You seem to be losing some information here about the authentication result

membership-attribute-service/app/actions/CommonActions.scala
```diff
@@ -22,24 +22,30 @@ class CommonActions(touchpointBackends: TouchpointBackends, bodyParser: BodyPars
...
+  def AuthAndBackendViaIdapiAction(howToHandleRecencyOfSignedIn: HowToHandleRecencyOfSignedIn): ActionBuilder[AuthAndBackendRequest, AnyContent] =
     NoCacheAction andThen new AuthAndBackendViaIdapiAction(touchpointBackends, howToHandleRecencyOfSignedIn, isTestUser)
 
+  def AuthAndBackendViaBothAction(
+      howToHandleRecencyOfSignedIn: HowToHandleRecencyOfSignedIn,
+      requiredScopes: List[AccessScope],
+  ): ActionBuilder[AuthenticatedUserAndBackendRequest, AnyContent] =
+    NoCacheAction andThen
+      new AuthAndBackendViaIdapiAction(touchpointBackends, howToHandleRecencyOfSignedIn, isTestUser) andThen
+      new AuthAndBackendViaAuthLibAction(touchpointBackends, requiredScopes, isTestUser)
```
kelvin-chappell, line 36: Not sure why this new action is necessary

kelvin-chappell (commented):
It seems like quite a large and complex change for a single PR.  I'm also worried that some auth info is getting lost along the way.

Perhaps a simpler approach would be better.  Eg. take a single endpoint like `anyPaymentDetails` and do the bare minimum that's required to get the first and last names from Idapi into that response.  I think that would have far less risk and be possible to test all scenarios with more confidence.

membership-attribute-service/app/actions/CommonActions.scala
pawel-krupinski-guardian: If you look at where an instance of this class is created, there's no code path that returns Left.

pawel-krupinski-guardian: Instead the whole response is turned into Unauthorized or such.

membership-attribute-service/app/actions/CommonActions.scala
pawel-krupinski-guardian: What's the alternative when we need to check for both recency of login and scopes?

membership-attribute-service/app/actions/CommonActions.scala
pawel-krupinski-guardian: Here:
https://github.com/guardian/members-data-api/blob/main/membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala#L20-L32

pawel-krupinski-guardian:
Fair point, I could split it into two PR (this one is already the 3rd of a series I think).

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -85,7 +85,7 @@ class AccountController(
           single("reason" -> nonEmptyText)
         }
         // transforming to Option here because type of failure is no longer relevant at this point
```
jonflynng, line 87:
```suggestion
       
```

pawel-krupinski-guardian:
On the other hand, we can't really test it (in CODE) right now.

membership-attribute-service/app/controllers/AccountController.scala
pawel-krupinski-guardian: Nice catch

pawel-krupinski-guardian:
Moved the refactorings to another commit for simplicity.
Can someone review?

membership-attribute-service/app/actions/CommonActions.scala
kelvin-chappell: Scopes only apply when authorising with Okta so initially could just use `AuthAndBackendViaIdapiAction` for Idapi auth.  Okta support is coming to that action imminently, certainly within the next week.

membership-attribute-service/test/acceptance/data/TestPaidSubscriptionPlan.scala
```diff
@@ -21,8 +21,8 @@ object TestPaidSubscriptionPlan {
       features: List[Feature] = Nil,
       charges: C = TestPaidCharge(),
       chargedThrough: Option[LocalDate] = None, // this is None if the sub hasn't been billed yet (on a free trial)
-      start: LocalDate = LocalDate.now().minusDays(14),
-      end: LocalDate = LocalDate.now().minusDays(14).plusYears(1),
+      start: LocalDate = LocalDate.now().minusDays(13),
```
charleycampbell, line 24: Just curious why we need this change on the number of days?

membership-attribute-service/test/acceptance/data/TestSubscription.scala
```diff
@@ -13,9 +13,9 @@ object TestSubscription {
       name: memsub.Subscription.Name = memsub.Subscription.Name(randomId("subscriptionName")),
       accountId: memsub.Subscription.AccountId = memsub.Subscription.AccountId(randomId("accountId")),
       startDate: LocalDate = LocalDate.now().minusDays(7),
-      acceptanceDate: LocalDate = LocalDate.now().minusDays(6),
```
charleycampbell, line 16: Same comment as above

charleycampbell (commented):
looks good. I'v left a couple of comments around the tests. Happy to approve but also aware you may want Jon or Kelvin to take a look

membership-attribute-service/test/acceptance/data/TestPaidSubscriptionPlan.scala
pawel-krupinski-guardian: I think it may have been clashing with another date and ideally I want each date to be different, so they can't get swapped in the output without the test failing.

membership-attribute-service/test/acceptance/data/TestSubscription.scala
pawel-krupinski-guardian: Same as above.

charleycampbell (approved):

PR #876, opened 2023-01-05
Acceptance test for me/mma endpoint

pawel-krupinski-guardian:
Acceptance test for existing functionality of /user-attributes/me/mma in preparation for functionality changes.

This includes:
- New test data classes
- AccountController refactoring - extracting some code into methods for readability.
- Refactoring to reuse code between AttributeControllerFunctionalTest and AccountControllerFunctionalTest.
- Switch to mockito-scala to take advantage of scala mockito syntax. 
- Port randomization added in acceptance tests.


membership-attribute-service/app/components/TouchpointBackends.scala
```diff
@@ -1,27 +1,56 @@
...
+    zuoraServiceOverride,
+  )(
     actorSystem,
     executionContext,
   )
   private val testStage = Stage(config.getString("touchpoint.backend.test"))
   val test =
-    new TouchpointComponents(testStage, createMetrics, config, supporterProductDataServiceOverride)(actorSystem, executionContext)
+    new TouchpointComponents(
+      testStage,
```
rupertbates, line 46: This is a bit strange now because we're using the same services for normal and test which defeats the point of having the two sets of components

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -322,39 +322,46 @@ class AccountController(
...
-    nonGiftContactAndSubscriptions <-
-      OptionEither.liftEitherOption(
-        subService.current[SubscriptionPlan.AnyPlan](contact) map {
-          _ map { subscription =>
-            ContactAndSubscription(contact, subscription, isGiftRedemption = false)
-          }
-        },
-      )
+  ): OptionT[OptionEither.FutureEither, List[ContactAndSubscription]] = {
+    def nonGiftContactAndSubscriptionsFor(contact: Contact): OptionT[FutureEither, List[ContactAndSubscription]] = {
```
rupertbates, line 326: I like the tidying up in this file 👍

membership-attribute-service/app/components/TouchpointBackends.scala
pawel-krupinski-guardian: If it was up to me, I'd throw away the whole test user code path.
Which is not that hard, cause it's just deleting code.

pawel-krupinski-guardian: If we ever get to testing the test users code path (which is probably never), we can address this.
They are different cases though - acceptance / integration testing and test users.

membership-attribute-service/app/controllers/AccountController.scala
pawel-krupinski-guardian: Thanks!

membership-attribute-service/app/components/TouchpointBackends.scala
rupertbates: Fair enough

PR #878, opened 2023-01-05
Tidy up alarms

rupertbates:
### Why do we need this? 
This PR makes the logging and alarms slightly more useful by:
- including the appropriate identity id when logging a dynamo read failure 
- sending alarms to the reader-revenue-dev group rather than just fulfilment-dev - this is a wider group and contains devs from conversion who also work on mdapi


pawel-krupinski-guardian (approved):
Makes sense.

PR #881, opened 2023-01-10
AccountController refactorings and simplifications

pawel-krupinski-guardian:
Refactorings leading up to the changes in /user-attributes/me/mma.
Making it easier to make the code change and simplifying the code paths the endpoint uses.
Unifying the use of scalaz classes and extracting code to separate classes / services.
More details in each commit message in this PR.

New scalaz-based type aliases are introduced and used throughout:
type SimpleEitherT[A] = EitherT[String, Future, A]
type ListTEither[A] = ListT[SimpleEitherT, A]


membership-attribute-service/app/utils/SimpleEitherT.scala
```diff
@@ -0,0 +1,21 @@
+package utils
+
+import scalaz.{EitherT, \/}
```
rupertbates, line 3: Can we use cats.EitherT instead please? I'd like to remove scalaz if possible

pawel-krupinski-guardian: That's a massive change since it's used everywhere.
I haven't introduced it, just aliased and renamed an existing type.

pawel-krupinski-guardian: If I was doing it though I'd switch everything, since it is all interlinked anyway.

pawel-krupinski-guardian: Some services in membership-common use EitherT.
And cats has no replacement for ListT.
So it is a bigger change.

rupertbates (approved):

PR #886, opened 2023-01-13
Refactorings

pawel-krupinski-guardian:
More refactorings leading up to functionality change in AccountController.
Split to another PR to make the functional change easier to review.

Includes:
- Removing optionality of user passed in to AuthenticatedUserAndBAckendRequest
- EitherT and OptionT usages replaced by the new type aliases.

### DIFF
--- a/membership-attribute-service/app/actions/CommonActions.scala
+++ b/membership-attribute-service/app/actions/CommonActions.scala
@@ -37,7 +37,7 @@ class CommonActions(touchpointBackends: TouchpointBackends, bodyParser: BodyPars
 class BackendRequest[A](val touchpoint: TouchpointComponents, request: Request[A]) extends WrappedRequest[A](request)
 
 class AuthenticatedUserAndBackendRequest[A](
-    val user: Either[AuthenticationFailure, UserFromToken],
+    val user: UserFromToken,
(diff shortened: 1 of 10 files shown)


membership-attribute-service/app/actions/CommonActions.scala
```diff
@@ -37,7 +37,7 @@ class CommonActions(touchpointBackends: TouchpointBackends, bodyParser: BodyPars
 class BackendRequest[A](val touchpoint: TouchpointComponents, request: Request[A]) extends WrappedRequest[A](request)
 
 class AuthenticatedUserAndBackendRequest[A](
-    val user: Either[AuthenticationFailure, UserFromToken],
```
jonflynng, line 40: Is this only used down below where it used to be converted directly to Option?

pawel-krupinski-guardian: Well, it had to be handled all over the place.
But it's actually unnecessary - if request is set (right), this one is set (right) as well.

pawel-krupinski-guardian: So this is an actual refactoring - no functionality is changed or code paths lost (except unused ones).

PR #889, opened 2023-01-16
Add new method to get user and credentials from either Okta or Idapi

kelvin-chappell:
This is in preparation for modifying the [AuthAndBackendViaIdapiAction](https://github.com/guardian/members-data-api/blob/main/membership-attribute-service/app/actions/AuthAndBackendViaIdapiAction.scala) to support both Okta and Idapi credentials.

As the logic will depend on the source of the credentials, ie either an Okta token or Idapi cookies, we need a new method that will give us a set of access claims and tell us where those claims came from: either Idapi or Okta.

This change is to give us that new method: `IdentityAuthService.userAndCredentials`.

I have also refactored the IdentityAuthService constructor so that it takes an `IdentityPlayAuthService` rather than building it itself.  This is to decouple it slightly and make it easier to test.

https://trello.com/c/LzxN3jTi/4226-update-mdapi-endpoints-that-use-alternative-authentication-technique

Tested on Code.

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -4,7 +4,8 @@ import akka.actor.ActorSystem
 import com.gu.aws.ProfileName
 import com.gu.config
 import com.gu.identity.IdapiService
-import com.gu.identity.auth.OktaTokenValidationConfig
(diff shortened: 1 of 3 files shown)


coldlink (approved):
looks okay to me!

PR #892, opened 2023-01-17
Using units from allowed list in metrics to make them work again (metrics broke after the last change).

pawel-krupinski-guardian:
Using units from allowed list in metrics to make them work again (metrics broke after the last change).


rupertbates:
Do you know why they stopped working? If so can include details in the description please

rupertbates (approved):

PR #894, opened 2023-01-19
Support Okta auth in remaining endpoints

kelvin-chappell:
This adds support for authorisation by Okta tokens in the `AuthAndBackendViaIdapiAction`.  With this change, all the API endpoints can now be authorised either by Okta token or by Idapi cookies.

In `AuthAndBackendViaIdapiAction`, we split the `refine` method so that it first uses the Identity library to determine what type of credentials are in use in the current request, if any.  It then either goes to `oktaRefine` or `idapiRefine`.  The `idapiRefine` method is unchanged.  The `oktaRefine` method is new and it works in the same way: if `Return401IfNotSignedInRecently` it keeps the session going for a maximum of 30 mins and after that time returns a 401, giving a hint to the client that the user should reauthenticate.

The rest of the PR is just adding access scopes to all the endpoints that use the `AuthAndBackendViaIdapiAction`.  These have been previously agreed.

Tested on Code.

https://trello.com/c/LzxN3jTi/4226-update-mdapi-endpoints-that-use-alternative-authentication-technique


coldlink (approved):
Code looks okay to me, would be useful if someone else who's worked with members-data-api have a look

PR #903, opened 2023-01-28
Metrics for duration of BasicStripeService, ZuoraRestService and ZuoraSoapService calls

pawel-krupinski-guardian:
Adding more granular metrics (service / http call level) to diagnose possible bottlenecks of /me/mma and potentially other endpoint calls.

ZuoraRestService and ZuoraSoapService brought in from membership-common.

Wrappers that add metrics introduced for each of them.

Adding call duration metrics to BasicStripeService instances.


membership-attribute-service/app/services/subscription/ZuoraSubscriptionService.scala
```diff
@@ -22,8 +21,9 @@ import utils.SimpleEitherT.SimpleEitherT
 import scala.concurrent.{ExecutionContext, Future}
 import scala.language.higherKinds
 import scala.util.Try
+import _root_.services.zuora.rest.SimpleClient
```
rupertbates, line 24: This import looks odd, does it need the _root_ bit?

pawel-krupinski-guardian: Overall it's needed in certain situations - like when there is a package on top level with the same name as subpackage (of the package of class the import sits in) - scala then can't reconcile if we mean relative or absolute.
Also sometimes needed when we wildcard-imported some package.

pawel-krupinski-guardian: In this case there's
import com.gu.memsub.subsv2._
higher up and subsv2 package has services subpackage.

At some point I'll restructure the packages so we won't have this problem.

membership-attribute-service/app/services/PaymentService.scala
```diff
@@ -23,7 +24,7 @@ import scalaz.{MonadTrans, OptionT, \/}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-class PaymentService(zuoraService: ZuoraService, planMap: Map[ProductRatePlanChargeId, Benefit])(implicit ec: ExecutionContext)
+class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePlanChargeId, Benefit])(implicit ec: ExecutionContext)
     extends api.PaymentService {
 
   implicit val monadTrans = MonadTrans[OptionT] // it's the only one we use here, really
```
rupertbates, line 30: Would love to get rid of this!

rupertbates (approved):

membership-attribute-service/app/services/PaymentService.scala
pawel-krupinski-guardian: The monad?

pawel-krupinski-guardian: Or the SoapService?

PR #910, opened 2023-02-01
Migrating membership-common classes to members-data-api

pawel-krupinski-guardian:
Migrating all the classes and tests used by members-data-api from membership-common, so we can retire membership-common going forward.

Process:
- Finding all com.gu class dependencies that come from membership-common (by looking at imports) and moving them to membership-data-api.
- Updating imports.
- Moving config files used.
- Moving tests.
- Removing unused classes, rationalizing package structure.
- Removing the dependency on membership-common


rupertbates (approved):
Lets keep a close eye on this one when it goes out

PR #931, opened 2023-02-22
WIP: Remove redundant zuora call

undergroundquizscene:
When finding existing payment methods, the code currently calls zuora twice for account details: the first time to turn a contact ID into an account ID (throwing away the rest of the account details in the process), and the second time to get the account details again.

This change surfaces the account details from the first call within ExistingPaymentOptionsController, and then removes the second call: the hope is that removing an unnecessary call to zuora will speed up the endpoint.

A couple of other places in the code also turn a contact ID into an account ID in the same way: I haven’t yet checked whether they can avoid a later call. For at least one of them, the later call is to zuora’s rest API, which seems to return slightly more info – if any of that info is used this would mean the call can’t be avoided, though perhaps the soap call could be replaced by an up-front rest call.

### DIFF
--- a/membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
+++ b/membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
@@ -8,6 +8,7 @@ import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.memsub.subsv2.{Subscription, SubscriptionPlan}
 import com.gu.memsub.{GoCardless, PayPalMethod, PaymentCard, PaymentCardDetails, PaymentMethod}
+import com.gu.zuora.soap.models.Queries.Account
 import com.typesafe.scalalogging.LazyLogging
 import components.TouchpointComponents
 import models.AccessScope.completeReadSelf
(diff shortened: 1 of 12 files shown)


membership-attribute-service/app/services/subscription/SubscriptionTransform.scala
```diff
@@ -97,6 +147,22 @@ object SubscriptionTransform {
...
+  def getSubscriptionsActiveOnOrAfterWithAccounts[P <: AnyPlan: SubPlanReads](
+      onOrAfter: LocalDate,
+  )(catalog: CatalogMap, pids: ProductIds)(
+      subJsons: List[(Account, JsValue)],
+  ): Disjunction[String, NonEmptyList[(Account, Subscription[P])]] =
+    Sequence(
+      subJsons
+        .map { case (account, subJson) =>
+          getSubscription[P](catalog, pids)(subJson).map(s => (account, s))
+        } // why can't I call map or traverse on tuples here to avoid manual (un)packing?
```
undergroundquizscene, line 159: I haven’t used scalaz a huge amount before – I’m assuming what’s missing here is the right kind of import or magic to get the relevant implicits into scope, but I can’t work it out. Let me know if you know!

membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
```diff
@@ -43,12 +44,12 @@ class ExistingPaymentOptionsController(
...
       contactRepository: ContactRepository,
       subscriptionService: SubscriptionService,
-  ): SimpleEitherT[Map[AccountId, List[Subscription[SubscriptionPlan.AnyPlan]]]] =
+  ): SimpleEitherT[Map[Account, List[(Account, Subscription[SubscriptionPlan.AnyPlan])]]] =
     (for {
       user <- ListTEither.fromOption(maybeUserId)
       contact <- ListTEither.fromFutureOption(contactRepository.get(user))
       subscription <- ListTEither.fromFutureList(subscriptionService.since[SubscriptionPlan.AnyPlan](date)(contact))
-    } yield subscription).toList.map(_.groupBy(_.accountId))
+    } yield subscription).toList.map(_.groupBy { case (account, sub) => account })
```
undergroundquizscene, line 52: I changed this groupBy to use the account rather than the ID, so that I could access the account later in place of the account ID. However, assuming the account comparison is a full equality test, this is inefficient: I’d rather group by their account ID but keep the first encountered account in each group (effectively assuming uniqueness of zuora’s account IDs, which seems fair). I suppose if the grouping could produce a non-empty list type this would be straightforward.

undergroundquizscene:
I’ve made a spreadsheet with some performance testing in CODE: https://docs.google.com/spreadsheets/d/1pdheqZYh391aT2xVKe9iVousw--r_sXSeVSaoK8CI1g/edit?usp=sharing

It’s a bit limited: only 20 tests each pre- and post-change, and only tested on an account with one existing payment method. However, I think it shows an improvement in runtime.

PR #934, opened 2023-03-02
Recognise Supporter Plus V2 subscriptions in /user-attributes/me

pawel-krupinski-guardian:
Making /user-attributes/me recognise Supporter Plus V2 subscriptions and return appropriate attribute values (same as for old supporter plus plans).

Creating an aws alarm for when there's an unsupported product rate plan id introduced in the future.
Fixing existing log-based alarms (they weren't working).

Setting user-friendly names (most of the resources are now named members-data-api-$stage) on CloudFormation resources so it's easier to find them.

Trello task:
https://trello.com/c/rN98TBpV/3871-make-members-data-api-recognise-supporter-plus-v2-subs

### DIFF
--- a/membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
+++ b/membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
@@ -74,7 +74,7 @@ object SupporterRatePlanToAttributesMapper {
     ) -> supporterPlusTransformer,
     List(
       "8a128ed885fc6ded018602296ace3eb8",
-      "8a128ed885fc6ded01860228f77e3d5a"
+      "8a128ed885fc6ded01860228f77e3d5a",
     ) -> supporterPlusV2Transformer,
     List(
       "2c92a0fb4edd70c8014edeaa4eae220a",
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
```diff
@@ -61,297 +72,320 @@ object SupporterRatePlanToAttributesMapper {
       GuardianPatronExpiryDate = Some(supporterRatePlanItem.termEndDate),
     )
   val guardianPatronProductRatePlanId = "guardian_patron"
-  val productRatePlanMappings: Map[Stage, Map[List[ProductRatePlanId], AttributeTransformer]] =
+
+  private val prodMappings: Map[List[ProductRatePlanId], AttributeTransformer] = Map(
```
rupertbates, line 76: I don't see that this refactor is giving us any value, we now just have to map from stage to a named variable below

pawel-krupinski-guardian: It's just splitting it into smaller, more manageable chunks rather than map of maps declared at once.
There's still one big map.

rupertbates (approved):
👍

PR #935, opened 2023-03-03
Upgrade to ubuntu jammy

tomrf1:
Ubuntu Bionic is approaching end of life. This PR updates the metering api to use an AMI based on Jammy.

Tested in CODE


paulbrown1982 (approved):
👍

PR #945, opened 2023-03-20
Log path + user-agent for OktaValidationErrors

tomrf1:
We've seen a big increase in 4XX responses caused by this error. Each has the cause:
`Caused by: io.jsonwebtoken.ExpiredJwtException: JWT expired at 2023-03-19T22:37:33Z`


kelvin-chappell (approved):
👍

PR #951, opened 2023-03-24
Disable 4xx alarm

tomrf1:
From SR Dev chat:

>> It seems that it's down to a difference in the treatment of access tokens between Android and iOS.  iOS refresh tokens on each cold start of the app and Android rely on the response from the API to decide when to refresh.  So in either case there shouldn't be any noticeable effect on users as Android will just get a new token and repeat the Mdapi call

>> The Android team are going to adopt the same behaviour as the iOS app so that should significantly reduce the number of 401s but not sure how long that will take so there might continue to be an alarm going off for a while


kelvin-chappell (approved):
👍

PR #952, opened 2023-03-24
Fix alarm cloudformation

tomrf1:
it seems it has to be in the Conditions block!


kelvin-chappell (approved):
I suppose could just disable actions instead but if this works 👍

PR #960, opened 2023-04-01
Serving first and last name from SalesForce rather than Identity in /me/mma

pawel-krupinski-guardian:
Serving first and last name from SalesForce rather than Identity in /me/mma, since first and last name are not yet available in identity response.

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -7,6 +7,7 @@ import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads
 import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.memsub.subsv2.{Subscription, SubscriptionPlan}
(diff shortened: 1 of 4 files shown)


pawel-krupinski-guardian:
[Serving first and last name from SalesForce rather than Identity in /me/mma](https://trello.com/c/l1FtdLPn/4018-serving-first-and-last-name-from-salesforce-rather-than-identity-in-me-mma)

graham228221:
@pawel-krupinski-guardian will this make any additional API requests to Salesforce? If so, can we get some estimates of how many requests will be made? We will want to be confident that we won't hit a Salesforce rate limit (particularly the 24 hour rolling API call limit).

If it is just using data from a callout that would have been made anyway then that is not a problem, but just wanted to check.

cc @guardian/salesforce-developers

pawel-krupinski-guardian:
@graham228221 No extra calls - I just moved the existing call to top level.

graham228221:
> @graham228221 No extra calls - I just moved the existing call to top level.

Perfect, thanks for confirming!

PR #1024, opened 2023-06-22
add new endpoint for updating subscription cancellation reason in Zuora

jonflynng:
This PR introduces a new endpoint in the AccountController that updates the cancellation reason in Zuora. This is part of the new cancellation saves work, which allows customers to optionally select the reason for cancelling at the end of the journey, after the cancellation is complete.

Previously, customers had to select the cancellation reason before cancelling. When confirming the cancellation, the reason would be sent to the API, which would perform both the reason update and cancellation. Now, these two actions must be done separately.

[frontend PR here](https://github.com/guardian/manage-frontend/pull/1123)

- [x] Tested in CODE


jonflynng:
[Membership Saves - Update reason](https://trello.com/c/p4JrQJyp/4174-membership-saves-update-reason)

michaelbjacobson (approved):
Nice! 👍🏼

PR #1028, opened 2023-07-13
Fix supporter plus v2 charge reads

rupertbates:
### Why do we need this? 
Since the move from supporter plus V1 to V2 members data api has been unable to load supporter plus subs because they now have two charges instead of one and this violates an assumption in membership-common that non-newspaper subs will only have one charge. This PR fixes this.


shtukas (approved):

membership-common/src/test/scala/com/gu/memsub/subsv2/services/CatalogServiceTest.scala
```diff
@@ -21,7 +21,7 @@ class CatalogServiceTest extends Specification {
       val dev = ConfigFactory.parseResources("touchpoint.CODE.conf")
       val ids = SubsV2ProductIds(dev.getConfig("touchpoint.backend.environments.CODE.zuora.productIds"))
       val cats = new CatalogService[Id](ids, FetchCatalog.fromZuoraApi(CatalogServiceTest.client("rest/Catalog.json")), identity, "CODE")
-      cats.catalog.map(_ => true) mustEqual \/.right(true)
+      cats.catalog.map(catalog => catalog.supporterPlus.plans.size) mustEqual \/.right(2)
```
paulbrown1982, line 24: I don't get why you needed to do this?

paulbrown1982 (approved):
👍  LGTM

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/ChargeListReads.scala
```diff
@@ -179,7 +180,17 @@ object ChargeListReads {
     }
   }
 
-  implicit def readSingle[B <: Benefit : ChargeReads]: ChargeListReads[ChargeList with SingleBenefit[B]] =
+  implicit def readSupporterPlusV2ChargeList: ChargeListReads[SupporterPlusCharges] = new ChargeListReads[SupporterPlusCharges] {
+
+    override def read(cat: PlanChargeMap, charges: List[ZuoraCharge]): ValidationNel[String, SupporterPlusCharges] = {
+      val pricingSummaries = charges.map(_.pricing)
+      Validation.success[NonEmptyList[String], SupporterPlusCharges](
+        SupporterPlusCharges(Month, pricingSummaries),
```
paulbrown1982, line 188: @rupertbates this code path is running for pretty-much _all_ subs, not just S+, and so everything's being reported as monthly bill cycle in MMA.

PR #1030, opened 2023-07-14
Fix bug where all charges have monthly billing period

rupertbates:
### Why do we need this? 
#1028 had a bug which hard coded the billing period of the charge lists attached to subscriptions to be monthly. 
This meant that the billing period displayed in MMA was incorrect for annual subscriptions as shown here:
mma-bug
Due to the incorrect mdapi response here:
mdapi-bug

### The changes 
This PR changes the serialisation of charge lists to take the billing period from the catalog

### DIFF
--- a/membership-common/src/main/scala/com/gu/memsub/subsv2/reads/ChargeListReads.scala
+++ b/membership-common/src/main/scala/com/gu/memsub/subsv2/reads/ChargeListReads.scala
@@ -182,11 +182,24 @@ object ChargeListReads {
 
   implicit def readSupporterPlusV2ChargeList: ChargeListReads[SupporterPlusCharges] = new ChargeListReads[SupporterPlusCharges] {
 
-    override def read(cat: PlanChargeMap, charges: List[ZuoraCharge]): ValidationNel[String, SupporterPlusCharges] = {
+    def getBillingPeriod(charges: List[ZuoraCharge]): ValidationNel[String, BillingPeriod] = {
+      val billingPeriods = charges.flatMap(_.billingPeriod).distinct
+      billingPeriods match {
+        case Nil => Validation.failureNel("No billing period found")
+        case b :: Nil => Validation.success[NonEmptyList[String], BillingPeriod](b.toBillingPeriod)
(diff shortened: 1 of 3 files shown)


membership-common/src/main/scala/com/gu/memsub/subsv2/reads/ChargeListReads.scala
```diff
@@ -182,11 +182,24 @@ object ChargeListReads {
...
-      Validation.success[NonEmptyList[String], SupporterPlusCharges](
-        SupporterPlusCharges(Month, pricingSummaries),
-      )
+      Validation
+        .success[NonEmptyList[String], List[PricingSummary]](pricingSummaries)
+        .ensure("No pricing summaries found".wrapNel)(_.nonEmpty)
+    }
+
+    override def read(cat: PlanChargeMap, charges: List[ZuoraCharge]): ValidationNel[String, SupporterPlusCharges] = {
+      (getBillingPeriod(charges) |@| getPricingSummaries(charges)).apply(SupporterPlusCharges)
```
paulbrown1982, line 202: @rupertbates How does the code stop non SupporterPlus subs successfully validating via this code path?

rupertbates: I've switched the ordering in readPaidChargeList so that supporter plus is the last one attempted

```scala
implicit def readPaidChargeList: ChargeListReads[PaidChargeList] = new ChargeListReads[PaidChargeList] {
  def read(cat: PlanChargeMap, charges: List[ZuoraCharge]): ValidationNel[String, PaidChargeList] = {
    readPaperChargeList.read(cat, charges).map(identity[PaidChargeList]) orElse2
      readPaidCharge[Benefit, BillingPeriod](readAnyProduct, anyBpReads).read(cat, charges) orElse2
      readSupporterPlusV2ChargeList.read(cat, charges).map(identity[PaidChargeList])
  }.withTrace("readPaidChargeList")
}
```

I'm not 100% clear if it matters if we use the 'wrong' deserialiser if they're all producing a PaidChargeList

shtukas (approved):

PR #1031, opened 2023-07-14
Run scalafmtAll on the membership-common project

rupertbates:
### Why do we need this? 
When the membership-common code was included in this repo it wasn't formatted according to the scalafmt rules that are in place. 

This is a problem because it leads to messy pull requests as files get formatted when they are worked on and often include many formatting changes which make it hard to see the actual behavioural changes being made.

This PR is the result of running 
```shell
sbt "project membership-common" scalafmtAll
```


paulbrown1982 (approved):
👍  I think I trust scalafmt knows what it's doing!

PR #1033, opened 2023-07-20
Improved validation for readSupporterPlusV2ChargeList

rupertbates:
### Why do we need this? 
The `readSupporterPlusV2ChargeList` ChargeListReads object was quite loose in the type of charges it would match, this PR tightens it up so that it only matches charges with billing periods which are valid for supporter plus subscriptions.


johnduffell (approved):
nice, that means it's less likely to accidentally scoop up non supporter plus, also when it fails, it will fall through.
I still have some doubt about it's picking up some subscriptions it shouldn't, so we should delete all the FreeCharge related code.  but we also complain that we have to store all the IDs for all the products, so we can't have it both ways.
Pie in the sky - I wonder if the new engine team's catalog as code could be imported instead?

PR #1035, opened 2023-09-08
Remove deprecated routes

rupertbates:
### Why do we need this? 
These routes in AccountController have been deprecated for ages (years?). I've checked the cloudwatch metrics and they are never called any more so I'm deleting them


kelvin-chappell (approved):
👍

PR #1034, opened 2023-09-08
Add support for National Delivery product - Part 1

rupertbates:
### Why do we need this? 
To support National paper delivery (Paperround) we need to make sure that the `/user-attributes/me` endpoint can recognise customers of this product.

This PR adds mappings for the product rate plans of the new National Delivery product to the SupporterRatePlanToAttributesMapper class so that we can correctly identify subscriptions stored in the SupporterProductData Dynamo store

## [Trello card](https://trello.com/c/MmL9heDE/49-product-build-supporter-product-data-mdapi)

PR #1036, opened 2023-09-14
Add support for National Delivery product - Part 2

rupertbates:
### Why do we need this? 
Following on from #1034 this PR enables the endpoints used by MMA to load National Delivery subscriptions

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -168,67 +168,6 @@ class AccountController(
       }
     }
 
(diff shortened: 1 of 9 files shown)


johnduffell (approved):
nice and simple PR 👍

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -168,67 +168,6 @@ class AccountController(
       }
     }
 
-  @Deprecated
```
rupertbates, line 171: This is no longer used

PR #1041, opened 2024-01-22
Add Digipack product to self-service-cancellation

KaliedaRik:
### Why do we need this? 
We need to update MDAPI to support a self-service cancellation journey for Digipack products.

This work supports some related work in the member-frontend repo, as detailed in this PR: https://github.com/guardian/manage-frontend/pull/1291 

### The changes 
One word added to one file


andrewHEguardian (approved):
👍

PR #1042, opened 2024-01-23
Remove recency tests when handling Okta tokens

kelvin-chappell:
# Problem
At the moment, the API is inspecting access tokens for their age and rejecting them if they are above a certain age, even when the token is valid.  APIs shouldn't know anything more about an access token than whether it is valid or not.
This recency test is also causing a potential race condition with similar checks taking place in the MMA client app.  So it's both redundant and harmful.
(We do need to continue to check the age of Identity cookies though, when they are presented, as they have no similar checking process in client apps.)

# Solution implemented
I've removed the recency check for Okta tokens.  It does a validity check, which includes an expiry check, in the `IdentityAuthService`.


raphaelkabo (approved):
Nice and simple, this is great!

PR #1043, opened 2024-01-29
Testing `alertableProducts` functionality

michaelbjacobson:
This PR inverts the logic determining whether or not to produce an alert for products in payment failure. Instead of a hard-coded list of products which are opted _into_ alerting, payment failure alerts will be generated for all products unless they are explicitly opted _out_.

### DIFF
--- a/membership-attribute-service/app/services/PaymentFailureAlerter.scala
+++ b/membership-attribute-service/app/services/PaymentFailureAlerter.scala
@@ -89,7 +89,7 @@ object PaymentFailureAlerter extends LoggingWithLogstashFields with StrictLoggin
       expectedAlertText.map { someText => shouldShowAlert.option(someText).flatten }
     }
   }
-  val alertableProducts = List(Product.Membership, Product.Digipack)
+  val alertableProducts = List(Product.Membership, Product.Contribution, Product.Digipack, Product.SupporterPlus)
 
   def alertAvailableFor(
       account: AccountObject,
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/services/PaymentFailureAlerter.scala
```diff
@@ -89,15 +89,16 @@ object PaymentFailureAlerter extends LoggingWithLogstashFields with StrictLoggin
       expectedAlertText.map { someText => shouldShowAlert.option(someText).flatten }
     }
   }
-  val alertableProducts = List(Product.Membership, Product.Contribution, Product.Digipack)
+
+  val nonAlertableProducts: List[Product] = List()
```
tjmw, line 93: Oh cool! So there are no non-alertable products now? That's great.

tjmw (approved):
Great! 👍

membership-attribute-service/test/services/PaymentFailureAlerterTest.scala
```diff
@@ -71,6 +71,28 @@ class PaymentFailureAlerterTest(implicit ee: ExecutionEnv) extends Specification
         result must beSome(expectedActionText).await
       }
 
+      "return a message for a supporter plus who is in payment failure" in {
+        val result: Future[Option[String]] =
+          PaymentFailureAlerter.alertText(accountSummaryWithBalance, supporterPlus, paymentMethodResponseRecentFailure)
+
+        val attemptDateTime = DateTime.now().minusDays(1)
```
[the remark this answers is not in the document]
michaelbjacobson: I considered this myself, but I decided to stick with the existing pattern. There's definitely a case for revising the whole thing, but it feels beyond the scope of this task.

membership-attribute-service/app/services/PaymentFailureAlerter.scala
michaelbjacobson: Exactly. As far as I'm aware there aren't currently any products which we want to prevent alerting for.

PR #1045, opened 2024-02-09
Add new feast boolean to contentAccess as well as advice on which iOS subscription group to offer

paulbrown1982:
### Why do we need this? 

There is a request by the Feast team to allow access to the app or to a free trial IAP by various cohorts of the existing Supporter base (see https://docs.google.com/presentation/d/1y6Sxe5pLiq8bUSreLiPy-CsWZHf-0DmWlFSRh3Qycks/edit).

I created this PR (1) to collaborate with Ben Briggs on what the new fields will look like so that work across both teams can run in parallel and (2) to roughly demonstrate that the code changes were not extensive, and possible without requiring access to more [SupporterProduct] data within MDAPI.

### The changes 
- Added new `feast` boolean within the `contentAccess` object
- Added a new `feastIosSubscriptionGroup` field at the top level to provide advice to the app on which subscription group (SKU) to offer (this moves logic to the backend saving logic in the app having to make the decision).

https://trello.com/c/qZy2vvXn and https://trello.com/c/rPUKspl9

### DIFF
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -74,7 +74,7 @@ case class Attributes(
     paperSubscriber = isPaperSubscriber,
     guardianWeeklySubscriber = isGuardianWeeklySubscriber,
     guardianPatron = isGuardianPatron,
-    feast = isGuardianPatron || isPartnerTier || isPatronTier || isGuardianPatron ||
+    feast = isStaffTier || isPatronTier || isPartnerTier || isPatronTier || isGuardianPatron ||
       (isSupporterPlus && SupporterPlusAcquisitionDate.exists(isBeforeFeastLaunch))
   )
 
--- a/membership-attribute-service/app/models/Attributes.scala
+++ b/membership-attribute-service/app/models/Attributes.scala
@@ -74,7 +74,7 @@ case class Attributes(
     paperSubscriber = isPaperSubscriber,
     guardianWeeklySubscriber = isGuardianWeeklySubscriber,
     guardianPatron = isGuardianPatron,
-    feast = isStaffTier || isPatronTier || isPartnerTier || isPatronTier || isGuardianPatron ||
+    feast = isStaffTier || isPartnerTier || isPatronTier || isGuardianPatron ||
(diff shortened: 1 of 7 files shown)


membership-attribute-service/app/models/Attributes.scala
```diff
@@ -69,6 +74,8 @@ case class Attributes(
     paperSubscriber = isPaperSubscriber,
     guardianWeeklySubscriber = isGuardianWeeklySubscriber,
     guardianPatron = isGuardianPatron,
+    feast = isStaffTier || isPatronTier || isPartnerTier || isPatronTier || isGuardianPatron ||
```
rupertbates, line 77: You've got `isPatronTier` in here twice

rupertbates:
Can you add an explanation of what all this means into the description please?

paulbrown1982:
> Can you add an explanation of what all this means into the description please?

Done @rupertbates. This PR is an RFC on the approach rather than a production-ready change.

membership-attribute-service/app/models/Attributes.scala
```diff
@@ -85,6 +92,11 @@ case class Attributes(
       || isGuardianPatron
   )
 
+  lazy val feastIosSubscriptionGroup =
```
rupertbates, line 95: What is a subscription group?

paulbrown1982: image

rupertbates (approved):
This should be good to go out now pending final decision on whether we need to treat RCs differently based on acquisition data

membership-attribute-service/test/controllers/AttributeControllerTest.scala
```diff
@@ -242,7 +265,7 @@ class AttributeControllerTest extends Specification with AfterAll with Idiomatic
                  """.stripMargin)
   }
 
-  private def verifySuccessfullOneOfContributionsResult(result: Future[Result]) = {
+  private def verifySuccessfulOneOfContributionsResult(result: Future[Result]) = {
```
paulbrown1982, line 268: If you're gonna repair the two 'l's at the end of Successful you may as well also add the additional 'f' at the end of 'OneOf' or call it 'OneTime'!

PR #1049, opened 2024-02-27
fix high severity vulnerabilities and version bumps

johnduffell:
This PR fixes up several high severity vulnerabilities, plus a couple of incidental version bumps.

Tested locally and the server starts and serves a basic request

Tested in CODE and it seems to handle the main attributes endpoint fine.

### DIFF
--- a/membership-attribute-service/app/monitoring/SentryLogging.scala
+++ b/membership-attribute-service/app/monitoring/SentryLogging.scala
@@ -9,7 +9,6 @@ import monitoring.SentryLogging.SanitizedLogMessage
 import utils.SanitizedLogging
 import utils.Sanitizer.Sanitizer
 
-import scala.jdk.CollectionConverters._
 import scala.util.{Failure, Success, Try}
 
 object SentryLogging extends SanitizedLogging {
@@ -19,10 +18,12 @@ object SentryLogging extends SanitizedLogging {
       case Some(sentryDSN) =>
         logger.info(s"Initialising Sentry logging")
         Try {
-          val sentryClient = Sentry.init(sentryDSN)
+          Sentry.init(sentryDSN)
           val buildInfo: Map[String, String] = app.BuildInfo.toMap.view.mapValues(_.toString).toMap
           val tags = Map("stage" -> config.stage) ++ buildInfo
-          sentryClient.setTags(tags.asJava)
+          tags.foreach { case (key, value) =>
+            Sentry.setTag(key, value)
+          }
         } match {
           case Success(_) => logger.debug("Sentry logging configured.")
           case Failure(e) => logError(scrub"Something went wrong when setting up Sentry logging ${e.getStackTrace}")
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/monitoring/SentryLogging.scala
```diff
@@ -19,10 +18,12 @@ object SentryLogging extends SanitizedLogging {
...
         logger.info(s"Initialising Sentry logging")
         Try {
-          val sentryClient = Sentry.init(sentryDSN)
+          Sentry.init(sentryDSN)
           val buildInfo: Map[String, String] = app.BuildInfo.toMap.view.mapValues(_.toString).toMap
           val tags = Map("stage" -> config.stage) ++ buildInfo
-          sentryClient.setTags(tags.asJava)
+          tags.foreach { case (key, value) =>
+            Sentry.setTag(key, value)
+          }
```
johnduffell, line 26: this appeared to have changed in the later version of sentry lib, I just bumped it about 6 major versions (from 1.x.x. to 7.x.x).  I'm not sure we really use this functionality but I played around until it looked right in IntelliJ

project/Dependencies.scala
```diff
@@ -66,12 +67,11 @@ object Dependencies {
     specs2.exclude("org.specs2", "specs2-mock_2.13"),
     kinesis,
     logstash,
+    logback,
     anorm,
-    "com.amazonaws" % "aws-java-sdk-autoscaling" % awsClientVersion,
```
johnduffell, line 70: was unused, I think this was for when we used to set the request limit for attributes requests to zuora based on a fraction of the zuora request limit

project/Dependencies.scala
```diff
@@ -66,12 +67,11 @@ object Dependencies {
...
     netty,
     nettyHttp,
     "com.google.guava" % "guava" % "32.1.3-jre", // until https://github.com/playframework/playframework/pull/10874
-    akkaHttpCore,
```
johnduffell, line 74: was unused and we are moving to pekko now, this repo is on play 3 already

membership-attribute-service/app/monitoring/SentryLogging.scala
jamesgorrie: We do sent server side errors to sentry, but I am not sure we use it. I am not sure what we get that we don't get from `CloudWatch`- maybe it's worth adding to the noise reduction effort going on?

jamesgorrie (approved):
Fab - nice work.

membership-attribute-service/app/monitoring/SentryLogging.scala
johnduffell: I think sentry should remember all the exceptions only, group them together intelligently, and notify us somehow when something new goes wrong.  It sounds a lot better than cloudwatch's blunt metrics when used correctly.  But in reality I haven't used it, so it hasn't any value for me!

PR #1050, opened 2024-03-20
Exclude removed contribution plans from subscriptions

rupertbates:
### Why do we need this? 
Currently members-data-api doesn't filter out removed contribution plans (on subs which have switched to supporter plus) when loading them from Zuora, this PR fixes that. 

If we allow that then people who used to be on recurring contribution are still appearing to be on it, which means that if they try to switch the contribution amount, it will amend the old charge which is not used.

Now it returns an error, and hopefully it will also avoid that code path in the first place.  We will monitor the product-move-api logs also to check that people are not still trying to switch from supporter plus to supporter plus, which causes an error there.

There are lots more details in the [trello card](https://trello.com/c/uv5JMf9s/461-make-mdapi-filter-out-switched-contributions)


johnduffell (approved):
nice, wasn't as bad as I expected in the end!

PR #1051, opened 2024-03-22
Give full or trial access to Feast for paper and GW subscribers.

rupertbates:
### Why do we need this? 
Paper subscribers should be given access to the Feast app, and Guardian Weekly subscribers should be offered an extended free trial.


paulbrown1982 (approved):
👍 LGTM

paulbrown1982 (approved):
👍 still LGTM

PR #1053, opened 2024-03-28
fix: change SupporterTier to have free feast trial

jamesgorrie:
The `SupporterTier` should have a 6 month free trial for feast rather than full access forever. 

(Unit tests to follow)


johnduffell (approved):
looks good that will put them in the ExtendedTrial group instead of feast=true

PR #1056, opened 2024-04-22
make all logging use a consistent logger

johnduffell:
This is PR 1 of about 4, the idea is to keep them conceptually small.

This PR switches us from various custom logger wrappers to one single SafeLogging trait.
This gives us the correct logger class name (raterh than always SafeLogger), and I have also made it so the correct file name/line number appears.
WAS:
image
NOW:
image

It also removes the custom logstash (Kibana) properties as we haven't used ELK for 5? years or so.
 
TODO in later PRs:
- remove the other logstash stuff as if we were reintroducing ELK we would start from scratch.
- remove some excessive logging of Auth that's filling up the log files.
- add identity id to all log lines relevant to a request, so we can search more easily

Tested locally, everything looks great.
 tested in CODE and check cloudwatch looks fine


membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -148,7 +146,7 @@ class TouchpointComponents(
   lazy val futureCatalog: Future[CatalogMap] = catalogService.catalog
     .map(_.fold[CatalogMap](error => { println(s"error: ${error.list.toList.mkString}"); Map() }, _.map))
     .recover { case error =>
-      logError(scrub"Failed to load the product catalog from Zuora due to: $error")
+      logger.error(scrub"Failed to load the product catalog from Zuora due to: $error")
```
johnduffell, line 149: this is a more typical syntax for log statements, it's the same as the standard scala-logging way (apart from we "scrub" to avoid sending PII to sentry

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -126,46 +126,38 @@ class AttributeController(
...
-          def customFields(supporterType: String): List[LogField] = List(
-            LogFieldString("lookup-endpoint-description", endpointDescription),
-            LogFieldString("supporter-type", supporterType),
-            LogFieldString("data-source", fromWhere),
-          )
-
           val result = allProductAttributes match {
             case Some(attrs @ Attributes(_, Some(tier), _, _, _, _, _, _, _, _, _, _, _)) =>
-              logInfoWithCustomFields(
+              logger.info(
```
johnduffell, line 131: this was sending some extra datapoints to ELK for some purpose,but we haven't sent anything there for about 5 years, so I changed back to the normal logger.info

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -175,10 +167,10 @@ class AttributeController(
         }).recover { case e =>
           // This branch indicates a serious error to be investigated ASAP, because it likely means we could not
           // serve from either Zuora or DynamoDB cache. Likely multi-system outage in progress or logic error.
-          val errMsg = s"Failed to serve entitlements either from cache or directly. Urgently notify Retention team: $e"
+          val errMsg = scrub"Failed to serve entitlements either from cache or directly. Urgently notify Retention team: $e"
           metrics.incrementCount(s"$endpointDescription-failed-to-serve-entitlements")
-          log.error(errMsg, e)
-          InternalServerError(errMsg)
+          logger.error(errMsg, e)
+          InternalServerError("failed to serve entitlements")
```
johnduffell, line 173: shouldn't pipe actual exceptions into an InternalServerError response as it's a security bad practice

membership-attribute-service/app/wiring/AppLoader.scala
```diff
@@ -16,28 +19,24 @@ import play.api.mvc.EssentialFilter
...
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
 
 class AppLoader extends ApplicationLoader {
   def load(context: Context): Application = {
     LoggerConfigurator(context.environment.classLoader).foreach {
       _.configure(context.environment)
     }
+    val loggerPackageName = classOf[SafeLoggerImpl].getPackageName
+    LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext].getFrameworkPackages.add(loggerPackageName)
```
johnduffell, line 38: this is needed so that the logger doesn't add `SafeLogger.scala:49` into the logs, instead of the real file/line number

membership-common/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionService.scala
```diff
@@ -1,28 +1,28 @@
...
 import org.joda.time.{LocalDate, LocalTime}
 import play.api.libs.json.{Reads => JsReads, _}
-import scala.language.higherKinds
-import scala.util.Try
 import scalaz._
 import scalaz.syntax.all._
 import scalaz.syntax.std.either._
 import scalaz.syntax.std.option._
 
+import scala.language.higherKinds
```
johnduffell, line 23: this annotation is not needed in scala 2.13 is it's on by default.

membership-common/src/main/scala/com/gu/monitoring/SafeLogger.scala
```diff
@@ -1,21 +1,33 @@
 package com.gu.monitoring
 
-import com.typesafe.scalalogging.StrictLogging
-import org.slf4j.{Marker, MarkerFactory}
+import com.gu.monitoring.SafeLogger.LogMessage
+import com.typesafe.scalalogging.Logger
+import org.slf4j.{LoggerFactory, Marker, MarkerFactory}
 
-object SafeLogger extends StrictLogging {
+trait SafeLogging {
```
johnduffell, line 7: these are the changes to get the new syntax, this is just taken from support-frontend
https://github.com/guardian/support-frontend/blob/main/support-config/src/main/scala/com/gu/monitoring/SafeLogger.scala

membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
```diff
@@ -20,20 +20,20 @@ import play.api.mvc.{Action, AnyContent, BaseController, ControllerComponents}
 import scalaz.-\/
 import scalaz.std.scalaFuture._
 import scalaz.syntax.monadPlus._
-import services.salesforce.ContactRepository
-import services.subscription.SubscriptionService
-import services.zuora.rest.ZuoraRestService.ObjectAccount
+import _root_.services.salesforce.ContactRepository
+import _root_.services.subscription.SubscriptionService
+import _root_.services.zuora.rest.ZuoraRestService.ObjectAccount
```
jamesgorrie, line 25: Why have we had to do this here?

membership-attribute-service/app/loghandling/LogbackConfig.scala
```diff
@@ -2,14 +2,14 @@ package loghandling
...
 import com.gu.logback.appender.kinesis.KinesisAppender
 import net.logstash.logback.layout.LogstashLayout
-import org.slf4j.{Logger => SLFLogger, LoggerFactory}
+import org.slf4j.{LoggerFactory, Logger => SLFLogger}
 import play.api.{Logger => PlayLogger}
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 
 object LogbackConfig {
-
+//TODO delete all this code and config as we switched off kibana years ago
```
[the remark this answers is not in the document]
jamesgorrie: Worth creating an issue in Trello?

membership-attribute-service/app/wiring/AppLoader.scala
jamesgorrie: Nice! You could leave this comment in the code?

membership-attribute-service/test/services/SupporterProductDataIntegrationTest.scala
```diff
@@ -43,7 +42,7 @@ class SupporterProductDataIntegrationTest(implicit ee: ExecutionEnv) extends Spe
     "get attributes by identity id" in {
       supporterProductDataService.getNonCancelledAttributes("3355555").map {
         case Right(attributes) =>
-          logger.info(attributes.toString)
+          println(attributes.toString)
```
jamesgorrie, line 45: 👀 why not use `SafeLogger` here?

membership-common/src/main/scala/com/gu/monitoring/SafeLogger.scala
```diff
@@ -1,21 +1,33 @@
...
+
+  protected val logger: SafeLoggerImpl =
+    new SafeLoggerImpl(Logger(LoggerFactory.getLogger(getClass.getName)))
+
+  implicit class Sanitizer(val sc: StringContext) {
+    def scrub(args: Any*): LogMessage = {
```
jamesgorrie, line 13: For another PR - but I wonder if we should have `scrub` as the default, and you can have `unscrubbed` logs when you explicitly set it?

jamesgorrie (approved):
This looks great, glad it's been tested. Couple of questions, but nothing blocking!

membership-attribute-service/app/loghandling/LogbackConfig.scala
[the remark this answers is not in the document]
johnduffell: I'm just doing it all under this one https://trello.com/c/FKoWjBIi/246-clean-up-and-improve-mdapi-logs

membership-attribute-service/app/wiring/AppLoader.scala
johnduffell: good idea
```suggestion
    val loggerPackageName = classOf[SafeLoggerImpl].getPackageName
    // at the moment we get SafeLogger.scala:49 type things in the logs
    // adding to the FrameworkPackages makes logback skip over and find the real line of code
    LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext].getFrameworkPackages.add(loggerPackageName)
```

membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
johnduffell: no idea! I just did optimise imports on all the changed files as there were loads of unused imports
Let me work out what the issue is

membership-common/src/main/scala/com/gu/monitoring/SafeLogger.scala
johnduffell: I'm not sure what you mean - scrub is a string interpolator that returns both the literal and the interpolated string, and the SafeLogger  (for `error`) sends only the literal to sentry (i.e. no PII) and the actual full log statementn to cloudwatch.
So sentry is always scrubbed, and cloudwatch is never scrubbed.

membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
johnduffell: ok so the reason is on line 10 
`import com.gu.memsub._`
there is a `services` package in there which is then in scope.
So since it wants the services package at the root, it has to do that.

I'm inclined to just stick with what it wants, although ideally we wouldn't have two packages with the same name.

jamesgorrie: Yea, if auto-formatting says, I normally defer to to avoid having to think about it. Good to know the reason though, thanks!

membership-common/src/main/scala/com/gu/monitoring/SafeLogger.scala
jamesgorrie: Ah, OK - I think I misunderstood what was happening. 

Still feels like we do a lot of `logger.warn(scrub"message")` and wondered if scrubbing could just be the default.

membership-attribute-service/app/loghandling/LogbackConfig.scala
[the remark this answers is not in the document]
johnduffell: PR is here now https://github.com/guardian/members-data-api/pull/1057

membership-common/src/main/scala/com/gu/monitoring/SafeLogger.scala
johnduffell: yeah I know what you mean, but actually by default there are no `${...}` insertions in string literals.

The usual `s"..."` is syntactic sugar for calling a (scala library) function called `s` on `StringContext` that takes the "chunks" of string literal and a list of values to insert, and returns a plain string
https://www.scala-lang.org/api/current/scala/StringContext.html

So if we want to save the normal string (for cloudwatch) as well as a scrubbed one (for sentry), we need to call a different function that returns both (e.g. in a case class). This is our function `scrub`.

We could try to somehow overload `s` or use implicits even more cleverly to have both the built in `s` and an `s` that returns the scrubbed string, but I wonder if that would be more confusing?  I think normal overloading only works on parameters.

Another option is to stop sending errors to sentry from the server side, as as far as I know no one acts on the alerts.

Happy to pair or take suggestions!

PR #1058, opened 2024-04-22
reduce noisy logging

johnduffell:
this just removes a couple of log lines that are heavy in terms of quantity, and as far as I know not useful.


kelvin-chappell (approved):
No-one actively looks for these log entries and we can put in more useful logging if necessary in future so 👍

PR #1060, opened 2024-04-23
log identity id on error

johnduffell:
This PR passes around an identity id context to the logger everywhere so that we can just search by identity id in the logs and find absolutely everything relevant.  It also makes it easy to add things to the context as it's all contained in one place.

I decided to make it implicit because logging is clutter anyway, so making us pass it in explicitly seemed even more clutterful.  But I know implicits are a risk and we almost always incline away from them nowadays!

I investigated how other libraries work and basically implicits seem to be standard including in play framework itself
https://github.com/playframework/playframework/blob/b8304c897f473fb9d41a7c62a580e2122cc39e48/core/play/src/main/scala/play/api/Logger.scala#L133
(it makes it optional by having a default value if you don't override it
https://github.com/playframework/playframework/blob/b8304c897f473fb9d41a7c62a580e2122cc39e48/core/play/src/main/scala/play/api/Logger.scala#L369
)

I thought about other approaches but since we're using execution contexts rather than threads, things like ThreadLocal would not be reliable.  There's also an [MDC thing in logback](https://logback.qos.ch/manual/mdc.html) that also uses threads.  I'm not sure there are other viable approaches?

My inclination is to go for this (plus the followup PRs to do the same with info/warn, but this is a small demo so I'm happy to rejig before I get too far).  Thoughts?


membership-attribute-service/app/actions/CommonActions.scala
```diff
@@ -52,10 +53,18 @@ class AuthenticatedUserAndBackendRequest[A](
...
 
 class AuthAndBackendRequest[A](
     val redirectAdvice: RedirectAdviceResponse,
     val touchpoint: TouchpointComponents,
     request: Request[A],
-) extends WrappedRequest[A](request)
+) extends WrappedRequest[A](request) {
+  implicit val logPrefix: LogPrefix = new LogPrefix {
+    override def message: String = redirectAdvice.userId.getOrElse("no-identity-id")
+  }
```
johnduffell, line 69: whatever kind of incoming request we get, we try to provide a suitable implicit log prefix.
we can later add things here if we need more information.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -103,7 +103,7 @@ class TouchpointComponents(
 
   private lazy val mapper = new SupporterRatePlanToAttributesMapper(stage)
   private lazy val dynamoSupporterProductDataService =
-    new DynamoSupporterProductDataService(dynamoClientBuilder.build(), supporterProductDataTable, mapper, createMetrics)
+    new SupporterProductDataService(dynamoClientBuilder.build(), supporterProductDataTable, mapper, createMetrics)
```
johnduffell, line 106: this had a trait and a class but there was only one implementation so I collapsed them together for now

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -143,20 +143,20 @@ class TouchpointComponents(
   )
 
   lazy val futureCatalog: Future[CatalogMap] = catalogService.catalog
-    .map(_.fold[CatalogMap](error => { println(s"error: ${error.list.toList.mkString}"); Map() }, _.map))
+    .map(_.fold[CatalogMap](error => { logger.errorNoPrefix(scrub"error: ${error.list.toList.mkString}"); Map() }, _.map))
```
johnduffell, line 146: not sure where the println would have gone (probably not cloudwatch), it should be using the logger.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -143,20 +143,20 @@ class TouchpointComponents(
...
     .recover { case error =>
-      logger.error(scrub"Failed to load the product catalog from Zuora due to: $error")
+      logger.errorNoPrefix(scrub"Failed to load the product catalog from Zuora due to: $error")
```
johnduffell, line 148: if there's no relevant identity id I made a `...NoPrefix` method.

Another option (used by play framework) would be to provide a low priority implicit value of LogPrefix, however the drawback of that is if we forget to take the identity id as a parameter it would still compile but wouldn't log the identity id.  I'd rather have a compiler error unless we explicitly say it's NoPrefix.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -143,20 +143,20 @@ class TouchpointComponents(
...
       throw error
     }
 
   lazy val subscriptionService: SubscriptionService = {
-    lazy val zuoraSubscriptionService = new ZuoraSubscriptionService(productIds, futureCatalog, zuoraRestClient, zuoraSoapService.getAccountIds)
+    lazy val zuoraSubscriptionService = new ZuoraSubscriptionService(productIds, futureCatalog, zuoraRestClient, zuoraSoapService)
```
johnduffell, line 153: this wasn't trivial to get to compile with the new implicit parameter.  Rather than making a suitable trait, I just passed in the whole service.  It may be excessive to give it the whole thing, but there were no tests relying on it so it didn't cause any extra code there.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -143,20 +143,20 @@ class TouchpointComponents(
...
 
     subscriptionServiceOverride.getOrElse(
       new SubscriptionServiceWithMetrics(zuoraSubscriptionService, createFineMetrics),
     )
   }
-  lazy val paymentService: ZuoraPaymentService = new ZuoraPaymentService(zuoraSoapService, catalogService.unsafeCatalog.productMap)
+  lazy val paymentService: PaymentService = new PaymentService(zuoraSoapService, catalogService.unsafeCatalog.productMap)
```
johnduffell, line 159: this had a trait and a class but there was only one implementation, so I squashed them together for now.

membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
```diff
@@ -43,7 +44,7 @@ class ExistingPaymentOptionsController(
       maybeUserId: Option[String],
       contactRepository: ContactRepository,
       subscriptionService: SubscriptionService,
-  ): SimpleEitherT[Map[AccountId, List[Subscription[SubscriptionPlan.AnyPlan]]]] =
+  )(implicit logPrefix: LogPrefix): SimpleEitherT[Map[AccountId, List[Subscription[SubscriptionPlan.AnyPlan]]]] =
```
johnduffell, line 47: now we have to make it so everything that does logging based off a request takes an identity id as an implicit parameter, which is a bit invasive but hopefully worth it?

membership-attribute-service/app/services/zuora/payment/SetPaymentCard.scala
```diff
@@ -11,11 +12,13 @@ import utils.SimpleEitherT.SimpleEitherT
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class SetPaymentCard(paymentService: ZuoraPaymentService, zuoraService: ZuoraSoapService, stripeServiceEither: Either[String, StripeService])(implicit
```
johnduffell, line 14: paymentService is unused

membership-attribute-service/test/testdata/TestLogPrefix.scala
```diff
@@ -0,0 +1,11 @@
+package testdata
+
+import com.gu.monitoring.SafeLogger.LogPrefix
+
+object TestLogPrefix {
+
+  implicit val testLogPrefix: LogPrefix = new LogPrefix {
+    override def message: String = "TestLogPrefix"
```
johnduffell, line 8: in the tests we can just import the hard coded prefix to keep it happy

membership-common/src/main/scala/com/gu/memsub/subsv2/Subscription.scala
```diff
@@ -126,12 +126,14 @@ object GetCurrentPlans {
         case _ => ensureStarted(_.start <= dateToCheck)
       }
       val freePlanCancelled = alreadyStarted.ensure(DiscardedPlan(plan, "has a free plan which has been cancelled").wrapNel)(_)
-      val contributorPlanCancelled = alreadyStarted.ensure(DiscardedPlan(plan, "has a contributor plan which has been cancelled or removed").wrapNel)(_)
+      val contributorPlanCancelled =
```
johnduffell, line 129: these changes must have been scalafmt.  I don't know if membership-common tests are run of `scalafmtCheck`'ed as membership-commmon is not aggregaed (although they do all pass so they might be)

membership-common/src/main/scala/com/gu/monitoring/SafeLogger.scala
```diff
@@ -45,11 +49,16 @@ class SafeLoggerImpl(logger: Logger) {
...
-  def error(logMessage: LogMessage): Unit = {
+  def errorNoPrefix(logMessage: LogMessage): Unit = {
     logger.error(logMessage.withPersonalData)
     logger.error(SafeLogger.sanitizedLogMessage, logMessage.withoutPersonalData)
   }
 
+  def error(logMessage: LogMessage)(implicit logPrefix: LogPrefix): Unit = {
+    logger.error(logPrefix.message + ": " + logMessage.withPersonalData)
+    logger.error(SafeLogger.sanitizedLogMessage, logMessage.withoutPersonalData)
+  }
```
johnduffell, line 60: this is the new method for error.  I didn't add it to the sentry message as sentry is supposed to aggregate similar messages.
TODO in the next PR, add the equivalent for info, warn etc..

membership-common/src/main/scala/com/gu/zuora/soap/Client.scala
```diff
@@ -81,49 +78,39 @@ class Client(
...
-  }
 
   def parentFilter[C <: Query, P <: Query](child: C, foreignKey: (C) => String): SimpleFilter =
     SimpleFilter("Id", foreignKey(child))
 
   def childFilter[C <: Query, P <: Query with Identifiable](parent: P): SimpleFilter =
     SimpleFilter(parent.objectName + "Id", parent.id)
+
+  val noLogPrefix: LogPrefix = new LogPrefix {
+    override def message: String = "no-id"
```
johnduffell, line 113: for SOAP calls there's usually an identity id, except for health checks or scheduled jobs.  So I enforced by default in the compiler - rather than using the "NoPrefix" one I made a dummy prefix to pass in.

membership-attribute-service/test/acceptance/PaymentUpdateControllerAcceptanceTest.scala
```diff
@@ -393,13 +391,12 @@ class PaymentUpdateControllerAcceptanceTest extends AcceptanceTest {
       httpResponse.getStatus shouldEqual 200
 
       identityMockClientAndServer.verify(identityRequest)
-      subscriptionServiceMock.current[SubscriptionPlan.Contributor](contact)(any) was called
-      catalogServiceMock.unsafeCatalog was called
```
johnduffell, line 397: I don't know why it expected this to be called for updating payment details.  I think it was probably called as a side effect.

tomrf1 (approved):
looks good

johnduffell:
seems to work fine in CODE

PR #1061, opened 2024-04-30
delete unused membership-common code

johnduffell:
Membership-common was copied into members-data-api to make life easier now it's not used in many places.

There was a lot of redundant code, presumably mostly only used by the defunct subscribe and membership sites.

This PR deletes the code that's not needed.

It all seems to compilea and pass tests locally, tested in CODE and working.


jamesgorrie (approved):
Assuming if it compiles - the code wasn't used. Nice

johnduffell:
> Assuming if it compiles - the code wasn't used. Nice

yeah my only concern is when things like json serialisers come in. You can delete a statically "unused" field but it was actually used to write out a result in a response.
But I am relatively confident about this in principle, as membership-common would have had a lot of code only used by subscribe and membership.

PR #1062, opened 2024-05-01
delete more unused code

johnduffell:
This PR deletes more unused code

It looked like some of the membership-common code (that was deleted in https://github.com/guardian/members-data-api/pull/1061 ) had been copied into the main project (in order to instrument it for metrics).

A few of the methods were unused due to the similar reason - they were only used presumably by membership and subscribe.

This PR deletes all the unused code, all tests pass locally.  It also deletes the instrumentation that was added for timings (because it makes it easier to statically analyse for unused code). It could be added back in later if needed.  Or I could add it back now if it's important enough.

seems to work ok in CODE (/me and /me/mma)


rupertbates (approved):
Very satisfying to remove all this code!

PR #1064, opened 2024-05-02
add identity id to all log statements

johnduffell:
This PR follows on from https://github.com/guardian/members-data-api/pull/1060 which added this pattern to only logger.error statements.

This adds the identity ID to the rest of the log statements (i.e. info, warn)

All passes tests locally.

 tested in CODE and working /me/mma and /me
See here for how to search https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/members-data-api-CODE/log-events/members-data-api-CODE-application-i-0c83f2280aec43b54$3FfilterPattern$3D200235544

### DIFF
--- a/membership-attribute-service/app/actions/CommonActions.scala
+++ b/membership-attribute-service/app/actions/CommonActions.scala
@@ -54,9 +54,7 @@ class AuthenticatedUserAndBackendRequest[A](
     val touchpoint: TouchpointComponents,
     val request: Request[A],
 ) extends WrappedRequest[A](request) {
-  implicit val logPrefix: LogPrefix = new LogPrefix {
-    override def message: String = user.identityId
-  }
+  implicit val logPrefix: LogPrefix = user.logPrefix
 }
 
 class AuthAndBackendRequest[A](
--- a/membership-attribute-service/app/actions/CommonActions.scala
+++ b/membership-attribute-service/app/actions/CommonActions.scala
@@ -62,7 +62,5 @@ class AuthAndBackendRequest[A](
     val touchpoint: TouchpointComponents,
     request: Request[A],
 ) extends WrappedRequest[A](request) {
-  implicit val logPrefix: LogPrefix = new LogPrefix {
-    override def message: String = redirectAdvice.userId.getOrElse("no-identity-id")
-  }
+  implicit val logPrefix: LogPrefix = LogPrefix(redirectAdvice.userId.getOrElse("no-identity-id"))
 }
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -78,7 +78,7 @@ class AccountController(
 
   private def CancelError(details: String, code: Int): ApiError = ApiError("Failed to cancel subscription", details, code)
 
-  def extractCancellationReason(cancelForm: Form[String])(implicit request: play.api.mvc.Request[_]): Either[ApiError, String] =
+  def extractCancellationReason(cancelForm: Form[String])(implicit request: play.api.mvc.Request[_], logPrefix: LogPrefix): Either[ApiError, String] =
     cancelForm
       .bindFromRequest()
       .value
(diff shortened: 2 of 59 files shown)


membership-attribute-service/app/actions/CommonActions.scala
```diff
@@ -54,17 +54,13 @@ class AuthenticatedUserAndBackendRequest[A](
...
 }
 
 class AuthAndBackendRequest[A](
     val redirectAdvice: RedirectAdviceResponse,
     val touchpoint: TouchpointComponents,
     request: Request[A],
 ) extends WrappedRequest[A](request) {
-  implicit val logPrefix: LogPrefix = new LogPrefix {
-    override def message: String = redirectAdvice.userId.getOrElse("no-identity-id")
-  }
```
johnduffell, line 69: I previously made the LogPrefix a trait but that was totally unnecessary because it's just returning a fixed value, so now it's a case class.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -299,16 +299,17 @@ class AccountController(
...
   private def sendSubscriptionCancelledEmail(
       email: String,
       contact: Contact,
       plan: SubscriptionPlan.AnyPlan,
       cancellationEffectiveDate: Option[LocalDate],
-  ) =
+  )(implicit logPrefix: LogPrefix) =
     SimpleEitherT
-      .right(sendEmail(subscriptionCancelledEmail(email, contact, plan, cancellationEffectiveDate)))
+      .right(sendEmail.send(subscriptionCancelledEmail(email, contact, plan, cancellationEffectiveDate)))
```
johnduffell, line 312: this used to use an `apply` method, so I updated it to use a normal method as it's less confusing

membership-attribute-service/app/controllers/HealthCheckController.scala
```diff
@@ -36,7 +36,7 @@ class HealthCheckController(touchPointBackends: TouchpointBackends, override val
       if (failures.isEmpty) {
         Ok(Json.obj("status" -> "ok", "gitCommitId" -> app.BuildInfo.gitCommitId))
       } else {
-        failures.flatMap(_.messages).foreach(msg => logger.warn(msg))
+        failures.flatMap(_.messages).foreach(msg => logger.warnNoPrefix(msg))
```
johnduffell, line 39: we use `...NoPrefix` where there will never be an identity id available to log as it's not a user request.

membership-common/src/main/scala/com/gu/zuora/soap/Client.scala
```diff
@@ -109,8 +109,4 @@ object Client {
   def childFilter[C <: Query, P <: Query with Identifiable](parent: P): SimpleFilter =
     SimpleFilter(parent.objectName + "Id", parent.id)
 
-  val noLogPrefix: LogPrefix = new LogPrefix {
-    override def message: String = "no-id"
-  }
```
johnduffell, line 114: I moved this out to the LogPrefix object so it can be used from anywhere that it's needed, and is easier to discover.

membership-common/src/main/scala/com/gu/monitoring/SafeLogger.scala
```diff
@@ -37,18 +39,30 @@ class SafeLoggerImpl(logger: Logger) {
...
 
-  def info(logMessage: String): Unit = {
+  def infoNoPrefix(logMessage: String): Unit = {
     logger.info(logMessage)
   }
 
-  def warn(logMessage: String): Unit = {
+  def info(logMessage: String)(implicit logPrefix: LogPrefix): Unit = {
+    logger.info(logPrefix.message + ": " + logMessage)
+  }
```
johnduffell, line 48: this (and others) are the old method with the new parameter required for the prefix.  The new method has the "NoPrefix" suffix to make it clearer you only use them if there's no identity id available.

tomrf1 (approved):
👍 looks good

PR #1065, opened 2024-05-10
feat: make members-data-api user attributes work with new 3-tier product rate plans (supporter plus with guardian weekly)

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/c43i1tD0/754-check-that-members-data-api-understands-new-gw-digital-rate-plan-part-1-user-attributes)

This makes `members-data-api` user attributes work with the new 3-tier product rate plans:
- Add new attribute transformer for SupporterPlusWithGuardianWeekly
- Add Zuora product rate plan IDs for CODE and PROD
- Add test to check for correct user attributes
- Add test to check for correct handling of new IDs

## Why are you doing this?

At the moment 3-tier product purchases are handled by creating two subscriptions - one for Supporter Plus and one for Guardian Weekly.

This PR is part of the refactor to be able to just create one subscription containing the new rate plans.

## How to test

From the root of the repository, run `sbt` and then `test`.

## How can we measure success?

Subscribers to 3-tier products can get correct access to the relevant products.

## Next

Check that `members-data-api` understands new GW + Digital rate plan for MMA.

### DIFF
--- a/membership-attribute-service/test/controllers/AttributeControllerTest.scala
+++ b/membership-attribute-service/test/controllers/AttributeControllerTest.scala
@@ -41,6 +41,7 @@ class AttributeControllerTest extends Specification with AfterAll with Idiomatic
   private val userWithNewspaperPlusUserId = "6"
   private val userWithGuardianWeeklyUserId = "7"
   private val unvalidatedEmailUserId = "8"
+  private val userWithSupporterPlusWithGuardianWeeklyId = "9"
 
   private val testAttributes = Attributes(
     UserId = validUserId,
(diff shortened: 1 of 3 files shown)


membership-attribute-service/test/controllers/AttributeControllerTest.scala
```diff
@@ -543,6 +560,35 @@ class AttributeControllerTest extends Specification with AfterAll with Idiomatic
...
+             |  "userId": "$userWithSupporterPlusWithGuardianWeeklyId",
+             |  "guardianWeeklyExpiryDate":"${dateTimeInTheFuture.toLocalDate}",
+             |  "showSupportMessaging": false,
+             |  "contentAccess": {
+             |    "member": false,
+             |    "paidMember": false,
+             |    "recurringContributor": false,
+             |    "supporterPlus" : true,
+             |    "feast": true,
+             |    "digitalPack": true,
```
rupertbates, line 582: They shouldn't have digital pack should they?

rupertbates (approved):
As discussed we should merge this and then we can think about the S+ vs digisub booleans as a separate issue

PR #1066, opened 2024-05-14
filter out free/zero months from mma next payment date/amount

johnduffell:
This is part of the cancellation saves work - previously the mma response will tell the user they have a payment on (x date) of 0.00 if they are on a free month.
This PR changes it so it filters out all free months, and if there are any non-free months in the preview, it will display that date and amount.

members-data-api: /user-attributes/me/mma
before:
image

After:
image

subscription is this one: https://apisandbox.zuora.com/platform/subscriptions/8ad081dd8f390950018f39c730e226d3

If we were to give out more than a year free we would have to extend the preview window, but so far it's only 2 months.

### DIFF
--- a/membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
+++ b/membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
@@ -90,9 +90,7 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
       case _ => None
     }
 
-  private def getNextBill(subId: Id, account: Account, numberOfBills: Int)(implicit
-      logPrefix: LogPrefix,
-  ): Future[Option[Bill]] =
+  private def getNextBill(subId: Id, account: Account, numberOfBills: Int)(implicit logPrefix: LogPrefix): Future[Option[Bill]] =
     zuoraService.previewInvoices(subId, numberOfBills).map { previewInvoiceItems =>
       val maybeBillingSchedule = BillingSchedule.fromPreviewInvoiceItems(planMap, previewInvoiceItems)
       maybeBillingSchedule.flatMap { billingSched =>
@@ -105,20 +103,14 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
       }
     }
 
-  def getPaymentMethod(defaultPaymentMethodId: Option[String], defaultMandateIdIfApplicable: Option[String] = None)(implicit
+  def getPaymentMethod(maybePaymentMethodId: Option[String], defaultMandateIdIfApplicable: Option[String] = None)(implicit
       logPrefix: LogPrefix,
   ): Future[Option[PaymentMethod]] =
-    for {
-      maybePaymentMethod <- getPaymentMethod(defaultPaymentMethodId).withLogging(s"get payment method for $defaultPaymentMethodId")
+    (for {
+      paymentMethodId <- maybePaymentMethodId
     } yield for {
-      soapPaymentMethod <- maybePaymentMethod
-      memsubPaymentMethod <- buildPaymentMethod(defaultMandateIdIfApplicable, soapPaymentMethod)
-    } yield memsubPaymentMethod
-
-  private def getPaymentMethod(maybePaymentMethodId: Option[String])(implicit logPrefix: LogPrefix): Future[Option[Queries.PaymentMethod]] =
-    maybePaymentMethodId match {
-      case Some(paymentMethodId) => zuoraService.getPaymentMethod(paymentMethodId).map(Some(_))
-      case None => Future.successful(None)
-    }
+      soapPaymentMethod <- zuoraService.getPaymentMethod(paymentMethodId).withLogging(s"get payment method for $maybePaymentMethodId")
+    } yield buildPaymentMethod(defaultMandateIdIfApplicable, soapPaymentMethod))
+      .getOrElse(Future.successful(None))
 
 }
--- a/membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
+++ b/membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
@@ -109,19 +109,16 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
       logPrefix: LogPrefix,
   ): Future[Option[PaymentMethod]] =
     for {
-      maybePaymentMethod <- getPaymentMethodByAccount(defaultPaymentMethodId).withLogging(s"get payment method for $defaultPaymentMethodId")
+      maybePaymentMethod <- getPaymentMethod(defaultPaymentMethodId).withLogging(s"get payment method for $defaultPaymentMethodId")
     } yield for {
       soapPaymentMethod <- maybePaymentMethod
       memsubPaymentMethod <- buildPaymentMethod(defaultMandateIdIfApplicable, soapPaymentMethod)
     } yield memsubPaymentMethod
 
-  private def getPaymentMethodByAccount(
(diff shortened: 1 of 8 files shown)


membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
```diff
@@ -35,47 +36,22 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
...
-      }
-    }
-    (futureMaybePrice |@| getPaymentMethod(sub.accountId, defaultMandateIdIfApplicable) |@| lastPaymentDate) {
-      case (maybePrice, paymentMethod, lpd) => {
-        val nextPayment = maybePrice.map { price => Payment(price, nextPaymentDate) }
-        PaymentDetails(sub, paymentMethod, nextPayment, lpd)
-      }
+    for {
+      account <- zuoraService.getAccount(sub.accountId)
+      eventualMaybeBill = getNextBill(sub.id, account, 15).withLogging(s"next bill for $sub")
```
johnduffell, line 47: we used to avoid calling `billingSchedule` (now called `getNextBill`) where not necessary, e.g. membership and recurring contribution, this is because they behave(d) predictably, so we could optimise away the call to zuora.
However now contribution can be discounted (suspended) and membership is not such a mainstream product, it makes sense to simplify the code at the cost of the extra zuora call.

membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
```diff
@@ -35,47 +36,22 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
...
+      eventualMaybePaymentMethod = getPaymentMethod(account.defaultPaymentMethodId, defaultMandateIdIfApplicable) // kick off async
```
johnduffell, line 48: I got rid of the whole `|@|` applicative thing and just used the for comprehension trick where I create the futures first and flatmap them later.

membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
```diff
@@ -113,59 +90,34 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
...
-  ): Future[Option[BillingSchedule]] = {
-    val finder: ProductRatePlanChargeId => Option[Benefit] = planMap.get
-    val adapter: Seq[Queries.PreviewInvoiceItem] => Option[BillingSchedule] = BillingSchedule.fromPreviewInvoiceItems(finder)
-    val scheduleFuture = zuoraService.previewInvoices(subId, numberOfBills).map(adapter)
-    for {
-      account <- accountFuture
-      scheduleOpt <- scheduleFuture
-    } yield {
-      scheduleOpt.map(_.withCreditBalanceApplied(account.creditBalance))
+  ): Future[Option[Bill]] =
```
johnduffell, line 95: before this returned the whole list of bills (invoices) and now it only gets the one actually needed

membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
```diff
@@ -113,59 +90,34 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
...
+    zuoraService.previewInvoices(subId, numberOfBills).map { previewInvoiceItems =>
+      val maybeBillingSchedule = BillingSchedule.fromPreviewInvoiceItems(planMap, previewInvoiceItems)
+      maybeBillingSchedule.flatMap { billingSched =>
+        billingSched
+          .withCreditBalanceApplied(account.creditBalance)
+          .invoices
+          .list
+          .find(_.amount > 0)
```
johnduffell, line 103: this is the important bit - filter out zero invoices

membership-common/src/main/scala/com/gu/memsub/BillingSchedule.scala
```diff
@@ -36,14 +36,13 @@ case class BillingSchedule(invoices: NonEmptyList[Bill]) {
...
 
-  type ProductFinder = ProductRatePlanChargeId => Option[Benefit]
   case class BillItem(name: String, product: Option[Benefit], amount: Float, unitPrice: Float) {
     override lazy val toString = s"| $amount\t| ($unitPrice)\t| $name | ${product.mkString}"
   }
 
   object BillItem {
-    def fromItem(finder: ProductFinder)(in: PreviewInvoiceItem) =
-      BillItem(in.chargeName, finder(ProductRatePlanChargeId(in.productRatePlanChargeId)), in.price, in.unitPrice)
+    def fromItem(planMap: Map[ProductRatePlanChargeId, Benefit])(in: PreviewInvoiceItem) =
```
johnduffell, line 44: although technically more complex, it seemed less confusing to use a built in data type rather than aliasing a function

rupertbates (approved):
👍

membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
```diff
@@ -113,59 +90,36 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
...
+    } yield for {
+      soapPaymentMethod <- maybePaymentMethod
+      memsubPaymentMethod <- buildPaymentMethod(defaultMandateIdIfApplicable, soapPaymentMethod)
+    } yield memsubPaymentMethod
 
-  private def getPaymentMethodByAccount(account: Account)(implicit logPrefix: LogPrefix): Future[Option[Queries.PaymentMethod]] = {
-    val maybeEventualPaymentMethod = account.defaultPaymentMethodId.map(zuoraService.getPaymentMethod)
-    maybeEventualPaymentMethod match {
-      case Some(eventualPaymentMethod) => eventualPaymentMethod.map(Some.apply)
+  private def getPaymentMethodByAccount(
```
tjmw, line 118: Is this the right name now? It doesn't look like it does anything related to an account now? Before it took an `Account` as a parameter.

membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
johnduffell: correct

membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
johnduffell: good point, it's by the default payment method id (there was no point passing in the whole account to use just one field)

membership-attribute-service/app/services/zuora/payment/ZuoraPaymentService.scala
```diff
@@ -113,61 +90,27 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
...
-      case Some(eventualPaymentMethod) => eventualPaymentMethod.map(Some.apply)
-      case None => Future.successful(None)
-    }
-  }
+    (for {
+      paymentMethodId <- maybePaymentMethodId
+    } yield for {
+      soapPaymentMethod <- zuoraService.getPaymentMethod(paymentMethodId).withLogging(s"get payment method for $maybePaymentMethodId")
+    } yield buildPaymentMethod(defaultMandateIdIfApplicable, soapPaymentMethod))
+      .getOrElse(Future.successful(None))
```
johnduffell, line 114: following on from @tjmw 's comment https://github.com/guardian/members-data-api/pull/1066#discussion_r1601598518
I realised there were 2 methods with similar name that were mostly just juggling the types badly, so I've turned it into this chained for comprehension that seems to avoid most of the fuss.

PR #1067, opened 2024-05-15
use basic "reads" for update contribution amount

johnduffell:
This PR is a simpification preparation.

I was having a look through the whole SubPlanReads etc stuff and I noticed some opportunities to simplify.

We pass around a **lot** of `[P <: AnyPlan : SubPlanReads]` type parameters throughout the code.  The intention behind this was that we could ask for a "Partner" or whatever effectively based on its shape, and it would fail to read if it was the wrong thing, and the type checker would help us out.  Kind of like json parser implicits replicated into very deeply nested and general structures.
However I think that cleverness was mainly used by membership and subscribe frontends - I noticed that in this repo we almost always use AnyPlan, apart from in update contribution amount.

1. (**This PR**) update update contribution amount to also use AnyPlan and remove the non subscription specific version (it hasn't been called in the last fortnight according to the ELB logs)
2. (next PR) will remove the type parameter from everywhere, as well as from the `Subscription[P <: AnyPlan]` object, so that all subscriptions are of type AnyPlan.
3. (next PR) will remove Friend from the "Catalog readers" as noone has it any more -the last Friend plan ended in 2020 according to bigquery
4. (next PR) will remove "Free" plans and ChargeLists completely and flatten it out so that everything is treated as Paid.   it looks like Staff can just deserialise as a one time charge of zero.  I will need to test to make sure things aren't broken for staff, but we should probably cancel them all in zuora anyway.

Possibly other stuff will pop up along the way.

Tested in DEV that it rejects non contributions still with an error
```
Name(A-S00879989) plan is not a contribution: PaidSubscriptionPlan(RatePlanId(8ad081dd8f390950018f39c7311a26da),ProductRatePlanId(8ad08cbd8586721c01858804e3275376),Supporter Plus V2 - Monthly,,Supporter Plus,None,Supporter Plus,SupporterPlus,List(),SupporterPlusCharges(Month,List(PricingSummary(Map(GBP -> Price(0.0,GBP))), PricingSummary(Map(GBP -> Price(10.0,GBP))))),Some(2024-05-18),2024-04-18,2025-04-18)
```
, and there's also a (happy case) test in the AcccountControllerAcceptanceTest

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -4,6 +4,7 @@ import actions._
 import com.gu.i18n.Currency
 import com.gu.memsub
 import com.gu.memsub.BillingPeriod.RecurringPeriod
+import com.gu.memsub.Product.Contribution
 import com.gu.memsub.subsv2.reads.ChargeListReads._
 import com.gu.memsub.subsv2.reads.SubPlanReads
 import com.gu.memsub.subsv2.reads.SubPlanReads._
@@ -40,12 +41,11 @@ object AccountHelpers {
   case object NoFilter extends OptionalSubscriptionsFilter
 
   def subscriptionSelector[P <: SubscriptionPlan.AnyPlan](
-      subscriptionNameOption: Option[memsub.Subscription.Name],
+      subscriptionName: memsub.Subscription.Name,
       messageSuffix: String,
-  )(subscriptions: List[Subscription[P]]): Either[String, Subscription[P]] = subscriptionNameOption match {
-    case Some(subName) => subscriptions.find(_.name == subName).toRight(s"$subName was not a subscription for $messageSuffix")
-    case None => subscriptions.headOption.toRight(s"No current subscriptions for $messageSuffix")
-  }
+      subscriptions: List[Subscription[P]],
+  ): Either[String, Subscription[P]] =
+    subscriptions.find(_.name == subscriptionName).toRight(s"$subscriptionName was not a subscription for $messageSuffix")
 
   def annotateFailableFuture[SuccessValue](failableFuture: Future[SuccessValue], action: String)(implicit
       executionContext: ExecutionContext,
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -40,12 +41,11 @@ object AccountHelpers {
   case object NoFilter extends OptionalSubscriptionsFilter
 
   def subscriptionSelector[P <: SubscriptionPlan.AnyPlan](
-      subscriptionNameOption: Option[memsub.Subscription.Name],
+      subscriptionName: memsub.Subscription.Name,
```
johnduffell, line 44: this doesn't need to be optional any more as the "old" update contribution amount endpoint that doesn't specify which subscription is not called in the last fortnight

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -260,21 +257,25 @@ class AccountController(
...
-              .current[SubscriptionPlan.Contributor](contact)
-              .map(subs => subscriptionSelector(subscriptionNameOption, s"the sfUser $contact")(subs)),
+              .current[SubscriptionPlan.AnyPlan](contact)
+              .map(subs => subscriptionSelector(subscriptionName, s"the sfUser $contact", subs)),
           )
-          billingPeriod = subscription.plan.charges.billingPeriod.asInstanceOf[RecurringPeriod]
-          applyFromDate = subscription.plan.chargedThrough.getOrElse(subscription.plan.start)
-          currency = subscription.plan.charges.price.prices.head.currency
+          contributionPlan <- SimpleEitherT.fromEither(subscription.plan match {
+            case p: SubscriptionPlan.Contributor @unchecked /* extra guard needed due to type erasure --> */ if p.product == Contribution => Right(p)
```
johnduffell, line 264: this is the key update - now we are asking for it as AnyPlan, we need to downcast it back to Contributor

rupertbates (approved):
This is brilliant, so happy to see this code being simplified!

PR #1068, opened 2024-05-16
make supporter plus a subscription in MMA rather than recurring support

johnduffell:
At the moment supporter plus is classed as recurring support in MMA.
```
    {
      "mmaCategory": "recurringSupport",
      "tier": "Supporter Plus",
...
```

It's needed (TBC) to change it so we describe it as a subscription to align with the main checkout language.

This PR moves it over accordingly.

TODO
- confirm whether this is definitely ok cc @rBangay 
- fix test at https://github.com/guardian/members-data-api/blob/jd-splus-subscription/membership-attribute-service/test/acceptance/AccountControllerAcceptanceTest.scala#L260


johnduffell:
mmaCategory has been removed completely now

PR #925, opened 2023-02-14
Bringing membership-common in as a submodule

pawel-krupinski-guardian:
Bringing membership-common in as a submodule.
Rationale: most changes made to membership-common are for members-data-api.

PR #1069, opened 2024-05-16
switch to using membership-common code rather than untested duplicates

johnduffell:
So I did a load of simplification in previous PRs last month.
However I went in the wrong direction a bit based on what I just noticed:

--

Last year, we switched some code from membership-common to local copies (but didn't copy the relevant tests) e.g. https://github.com/guardian/members-data-api/pull/903
Later, membership-common as a whole was copied across, including its tests. https://github.com/guardian/members-data-api/pull/925
This meant that the code in use wasn't always the code that was functionally tested.

As part of my cleanups, I was deleting the unused (i.e. tested) code, and keeping the used/untested code.  Although it seems like I did keep some of the tested code.

This wasn't a great situation because it's hard to manually test things, so we rely ont he automated tests.

--

This PR removes some of the copied code, changes the controllers to use the "original" membership-common code, and restores and removed methods and tests (to the best of my ability)

--

Unfortunately it does involve bringing back some of the frowned upon `[M[_]: Monad]` stuff that has been removed at some point, but at this point I'd rather have that than missing tests.  We can always change the tests to use Futures as well if we feel like it.


membership-attribute-service/app/components/TouchpointBackends.scala
```diff
@@ -19,9 +19,9 @@ class TouchpointBackends(
     createMetrics: CreateMetrics,
     supporterProductDataServiceOverride: Option[SupporterProductDataService] = None,
     contactRepositoryOverride: Option[ContactRepository] = None,
-    subscriptionServiceOverride: Option[SubscriptionService] = None,
+    subscriptionServiceOverride: Option[SubscriptionService[Future]] = None,
     zuoraRestServiceOverride: Option[ZuoraRestService] = None,
-    catalogServiceOverride: Option[CatalogService] = None,
+    catalogServiceOverride: Option[CatalogService[Future]] = None,
```
johnduffell, line 24: we had at some point got rid of the Future type parameter as it was only needed for the tests.  I have brought back the tests with minimal changes (for now) so we need the type parameters back again.

membership-attribute-service/app/controllers/ExistingPaymentOptionsController.scala
```diff
@@ -1,13 +1,17 @@
 package controllers
 
+import _root_.services.salesforce.ContactRepository
+import _root_.services.zuora.rest.ZuoraRestService.ObjectAccount
```
johnduffell, line 4: we have `services` package as well as `com.gu.memsub.services`, so intellij adds `_root_` to differentiate them

membership-attribute-service/conf/logback.xml
```diff
@@ -19,7 +19,7 @@
         </rollingPolicy>
 
         <encoder>
-            <pattern>%date %file:%L [%level]: %msg%n%xException{3}</pattern>
+            <pattern>%date %file:%L [%level]: %msg%n%xException{full}</pattern>
```
johnduffell, line 22: we were only getting the first 3 lines of any exception in the logs, now we will get them all

membership-common/src/main/scala/com/gu/memsub/subsv2/services/CatalogService.scala
```diff
@@ -55,7 +60,7 @@ class CatalogService[M[_]: Monad](productIds: ProductIds, fetchCatalog: M[String
   ): Validation[NonEmptyList[String], P] = {
     many(plans.filter(_.frontendId.contains(frontendId)), name).flatMap {
       case a if a.size == 1 => Validation.s(a.head)
-      case e => Validation.failureNel(s"Too many plans! $e")
+      case e => Validation.failureNel(s"Too many plans! for name $name with frontendId $frontendId $e")
```
johnduffell, line 63: this was a change in the MDAPI version but not in the -common version, so I've preserved it (same with the below toDisjunction line)

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
```diff
@@ -88,6 +104,16 @@ class ZuoraSoapService(soapClient: soap.ClientWithFeatureSupplier)(implicit ec:
...
+  // Creates a payment method in zuora and sets it as default in the specified account.To satisfy zuora validations this has to be done in three steps
+  def createPaymentMethod(command: CreatePaymentMethod)(implicit logPrefix: LogPrefix): Future[UpdateResult] = for {
+    _ <- setGatewayAndClearDefaultMethod(
+      command.accountId,
+      command.paymentGateway,
+    ) // We need to set gateway correctly because it must match with the payment method we'll create below
+    createMethodResult <- soapClient.extendedAuthenticatedRequest[CreateResult](new XmlWriterAction(command)(createPaymentMethodWrites))
+    result <- setDefaultPaymentMethod(command.accountId, createMethodResult.id, command.paymentGateway, command.invoiceTemplateOverride)
+  } yield result
+
```
johnduffell, line 116: had to add back in to membership-common methods that were actually used in the copied version

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
```diff
@@ -31,13 +34,26 @@ object ZuoraSoapService {
   }
 }
 
-class ZuoraSoapService(soapClient: soap.ClientWithFeatureSupplier)(implicit ec: ExecutionContext) extends SafeLogging {
+trait SoapClient[M[_]] {
+
+  def getAccountIds(contactId: ContactId)(implicit logPrefix: LogPrefix): M[List[AccountId]]
+}
```
johnduffell, line 40: had to add the trait (with the type parameter) because just the one method was mocked in a test

membership-common/src/main/scala/com/gu/zuora/soap/Client.scala
```diff
@@ -60,7 +60,8 @@ class Client(
       logger.info(
         s"Zuora SOAP call in environment ${apiConfig.envName}. Request info:\n${action.prettyLogInfo}. Is authentication defined: ${authentication.isDefined}",
       )
-    client.execute(request)
+    client
+      .execute(request)
```
johnduffell, line 64: I don't think scalafmt is actually enforced on the membership-common code, but I did run it locally hence the random changes.

membership-common/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionService.scala
```diff
@@ -54,74 +64,210 @@ object Trace {
 
 }
 
-import com.gu.memsub.subsv2.services.Trace.Traceable
+class SubscriptionService[M[_]: Monad](pids: ProductIds, futureCatalog: => M[CatalogMap], rest: SimpleClient[M], soap: SoapClient[M])
```
johnduffell, line 67: all this was deleted as unused last month, but it turned out it was diplicated in the main codebase, so this puts it back again

membership-common/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionTransform.scala
```diff
@@ -17,7 +17,6 @@ import scalaz.syntax.all._
 import scalaz.syntax.std.either._
 import scalaz.syntax.std.option._
 
-import scala.language.higherKinds
```
johnduffell, line 20: language flag is on by default in scala 2.13

membership-attribute-service/app/components/TouchpointBackends.scala
rupertbates: 😢

johnduffell: of course we can get rid of them by making the tests use Future as well just like the main code (rather than just using unwrapped values) but I wasn't prepared to do it in this PR

rupertbates (approved):
pullmebackin

johnduffell:
@rupertbates 
image

PR #1071, opened 2024-05-16
remove friend

johnduffell:
TODO test in CODE before shipping due to config changes

This PR removes friend from all the data structures, meaning that if someone was a friend, they woudln't get their benefits.

according to this query, the last friend ended in 2020
```
select c.effective_end_date
from `intermediate.stg_zuora_subscription` s
join `intermediate.stg_zuora_rate_plan` rp on rp.subscription_id = s.subscription_id
join `intermediate.stg_zuora_rate_plan_charge` c on c.rate_plan_id = rp.rate_plan_id
where 1=1
and s.status = 'Cancelled'
and rp.rate_plan_name = 'Friends'
order by c.effective_end_date desc
```

next steps:
The next PR can get rid of this parameter, which is only needed to handle pending friend downgrades
https://github.com/guardian/members-data-api/blob/7d50dd2537e9c683d17166e1aec3dca4ed0dfebc/membership-common/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionTransform.scala#L123

PR #1072, opened 2024-05-20
remove staff membership

johnduffell:
This follows on from https://github.com/guardian/members-data-api/pull/1071
and removes staff as well as friend.

Although some people still have staff membership, it doesn't give any benefit and so we can ignore it in zuora and the catalog.

All staff should get free digital subscription which would prevent supporter asks and give ad free anyway.
Events discounts are done via other means now.

The next PR will clean up all the Paid/Free distinction, to keep things easier to review.
And there will also be a PR to remove the date based downgrade-to-friend logic. (Since friend/staff are modelled as one time charges, there was more complexity in working out if people were members.)


membership-attribute-service/app/models/FeastApp.scala
```diff
@@ -17,8 +17,7 @@ object FeastApp {
   private def isBeforeFeastLaunch(dt: LocalDate): Boolean = dt.isBefore(FeastIosLaunchDate)
 
   def shouldGetFeastAccess(attributes: Attributes): Boolean =
-    attributes.isStaffTier ||
```
johnduffell, line 20: this was misleading - staff who get digi access will get feast via `attributes.digitalSubscriberHasActivePlan` below (although in a convoluted way - this logic needs refactoring)

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
```diff
@@ -350,14 +336,12 @@ object SupporterRatePlanToAttributesMapper {
 sealed abstract class MembershipTier(val name: String, val value: Int)
 
 object MembershipTier {
-  case object Friend extends MembershipTier("Friend", 1)
-  case object Staff extends MembershipTier("Staff", 2)
   case object Supporter extends MembershipTier("Supporter", 3)
   case object Partner extends MembershipTier("Partner", 4)
   case object Patron extends MembershipTier("Patron", 5)
```
johnduffell, line 341: wasn't sure whether to renumber the ordering but I don't think It'll hurt to leave it

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
```diff
@@ -139,7 +138,6 @@ object Benefit {
       (id == Weekly.id).option(Weekly)
 
   sealed trait MemberTier extends Benefit
-  sealed trait FreeMemberTier extends MemberTier
```
johnduffell, line 142: I removed a few of the `Free...` things, but I tried to avoid doing to much in this PR, the next PR will be imminent.

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/ChargeListReads.scala
```diff
@@ -152,16 +139,9 @@ object ChargeListReads {
...
-    def read(cat: PlanChargeMap, charges: List[ZuoraCharge]): ValidationNel[String, FreeChargeList] = {
-      readFreeCharge[Benefit](readAnyProduct).read(cat, charges).map(identity[FreeChargeList])
-    }
-  }
-
   implicit def readChargeList: ChargeListReads[ChargeList] = new ChargeListReads[ChargeList] {
     override def read(cat: PlanChargeMap, charges: List[ZuoraCharge]) = {
-      readPaidChargeList.read(cat, charges) orElse2
-        readFreeChargeList.read(cat, charges)
+      readPaidChargeList.read(cat, charges).map(identity[ChargeList])
```
johnduffell, line 144: need to map with identity because functions are (co)variant but a ChargeListRead is not.  This means 
`ChargeListReads[PaidChargeList]` is not a subclass of `ChargeListReads[ChargeList]`

Whether it should be or not was not something I wanted to delve in to, since this function will disappear in the next PR.

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/SubPlanReads.scala
```diff
@@ -132,20 +131,15 @@ object SubPlanReads {
...
-            free =>
-              freePlanReads(implicitly[SubPlanReads[P]], ChargeListReads.pure(free))
-                .read(p, z, c)
-                .map(identity[SubscriptionPlan[P, ChargeList with SingleBenefit[B]]]),
-            paid =>
-              paidPlanReads(implicitly[SubPlanReads[P]], ChargeListReads.pure(paid))
-                .read(p, z, c)
-                .map(identity[SubscriptionPlan[P, ChargeList with SingleBenefit[B]]]),
-          ),
-        )
```
johnduffell, line 148: took me a minute to work this out - I think it might be using an Either in order to keep the correct types - so we don't need this.   I will revisit in the next PR when I remove Free/Paid distinction.

membership-common/src/test/scala/com/gu/memsub/subsv2/reads/ChargeListReadsTest.scala
```diff
@@ -135,17 +135,6 @@ class ChargeListReadsTest extends AnyFlatSpec {
 
   }
 
-  "product reads" should "not read any supporter as a Free member" in {
```
johnduffell, line 138: no such thing as a free member now.  We have tests for other kinds of incorrect charge lists below so I deleted this one.

membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
```diff
@@ -31,27 +31,26 @@ import utils.TestLogPrefix.testLogPrefix
   */
 class SubscriptionServiceTest extends Specification {
 
-  // this is the UAT prpId of friend, which we need for the catalog
-  val prpId = ProductRatePlanId("2c92c0f94cc6ea05014cdb4b1d1f037d")
+  val digipackAnnualPrpId = ProductRatePlanId("2c92c0f94bbffaaa014bc6a4212e205b")
   val partnerPrpId = ProductRatePlanId("2c92c0f84c510081014c569327003593")
-  val supporterPrpId = ProductRatePlanId("2c92c0f84bbfeca5014bc0c5a793241d")
+  val supporterPlusPrpId = ProductRatePlanId("8ad08cbd8586721c01858804e3275376")
```
johnduffell, line 36: I decided to update these tests to use S+ and digi sub, rather than friend and supporter.

membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
```diff
@@ -135,9 +134,9 @@ class SubscriptionServiceTest extends Specification {
   val subscriptions: HttpClient[scalaz.Id.Id] = new HttpClient[Id]() {
     override def execute(request: Request)(implicit logPrefix: SafeLogger.LogPrefix): scalaz.Id.Id[Response] =
       request.url().uri().getPath match {
-        case "/subscriptions/accounts/foo" => jsonResponse("rest/plans/accounts/Friend.json")(request)
-        case "/subscriptions/accounts/bar" => jsonResponse("rest/plans/accounts/Friend.json")(request)
```
johnduffell, line 139: not sure why it was reading the same subscription twice?

membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
```diff
@@ -341,13 +314,6 @@ class SubscriptionServiceTest extends Specification {
       result mustEqual \/-("idPartner")
     }
 
-    "tell you you aren't a staff membership if your subscription is cancelled regardless of the date" in {
```
johnduffell, line 344: no such thing as instant change of a one time charge subscription any more.  we only modelled friend and staff that way.

membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
```diff
@@ -382,27 +348,8 @@ class SubscriptionServiceTest extends Specification {
 
   "Subscription service" should {
 
-    "Be able to fetch a staff subscription" in {
```
johnduffell, line 385: need to put this back as a (e.g.) S+ test

membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
```diff
@@ -412,14 +359,14 @@ class SubscriptionServiceTest extends Specification {
 
     "Leverage the soap client to fetch subs by contact ID" in {
       val subs = service.current[SubscriptionPlan.AnyPlan](contact)
-      subs.headOption.map(_.name) mustEqual Some(memsub.Subscription.Name("subscriptionNumber")) // what is in the config file
+      subs.map(_.name.get).sorted mustEqual List("A-S00890520", "A-S00890521") // from the test resources jsons
```
johnduffell, line 378: not sure what the point of testing read the same sub twice, so I read both the annual digi and the monthly S+ and chec they both appear

membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
johnduffell: done, changed to S+

PR #1075, opened 2024-05-21
remove code to handle pending free plan i.e. friend

johnduffell:
This removes pending free plan facility, that was only really useful for downgrade to friend situation.  Since tier change is no longer possible in membership, and there are no friends, this code is redundant.


membership-common/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionService.scala
```diff
@@ -95,53 +90,47 @@ class SubscriptionService[M[_]: Monad](pids: ProductIds, futureCatalog: => M[Cat
...
+    for {
+      subJson <- rest.get[JsValue](url)(idReads, logPrefix)
+      catalog <- futureCatalog
+    } yield {
+      val allSubscriptionsForSubscriberName =
+        for {
+          jsValue <- subJson.withLogging(s"get subscription $name from zuora")
+          subscription <- SubscriptionTransform.getSubscription(catalog, pids)(jsValue).withLogging(s"getAllValidSubscriptionsFromJson for $name")
+        } yield subscription
+      allSubscriptionsForSubscriberName.toOption
```
johnduffell, line 110: most of the code in this file seemed to use nested flatmaps which didn't seem clear so I've rewritten as for comprehensions

membership-common/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionService.scala
```diff
@@ -95,53 +90,46 @@ class SubscriptionService[M[_]: Monad](pids: ProductIds, futureCatalog: => M[Cat
...
+    for {
+      subJson <- rest.get[JsValue](url)(idReads, logPrefix)
+      catalog <- futureCatalog
+    } yield for {
+      jsValue <- subJson.withLogging(s"get subscription $name from zuora").toOption
+      subscription <- SubscriptionTransform
+        .getSubscription(catalog, pids)(jsValue)
+        .withLogging(s"getAllValidSubscriptionsFromJson for $name")
+        .toOption
+    } yield subscription
```
johnduffell, line 110: most of the code in this file used nested flatmaps etc so I converted to for comprehensions which I think is clearer

rupertbates (approved):

PR #1078, opened 2024-05-23
simplify the catalog related code

johnduffell:
This PR is part of the refactoring series and this one simplifies the catalog related code.  See inline comments


membership-attribute-service/app/components/TouchpointBackends.scala
```diff
@@ -21,7 +22,7 @@ class TouchpointBackends(
     contactRepositoryOverride: Option[ContactRepository] = None,
     subscriptionServiceOverride: Option[SubscriptionService[Future]] = None,
     zuoraRestServiceOverride: Option[ZuoraRestService] = None,
-    catalogServiceOverride: Option[CatalogService[Future]] = None,
+    catalogServiceOverride: Option[Future[CatalogMap]] = None,
```
johnduffell, line 25: before we had a CatalogService and basically the only thing we needed was the CatalogMap which goes from ProductRatePlanId to CatalogZuoraPlan (which is the low level catalog model)
All the high level stuff was read in and validated into fully qualified types but never used in the actual code.

So now the whole thing is just the above Map and the CatalogService is just a way of parsing the Json into the Map.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -102,10 +96,9 @@ class TouchpointComponents(
 
   lazy val zuoraSoapService = {
     lazy val zuoraSoapClient =
-      new ClientWithFeatureSupplier(
+      new Client(
```
johnduffell, line 99: we used to have this ClientWithFeatureSupplier class as a subclass of Client, but all the code was deleted from it recently as it was not needed, so I've flattened it out of the hierarchy

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -102,10 +96,9 @@ class TouchpointComponents(
...
         apiConfig = backendConfig.zuoraSoap,
         httpClient = RequestRunners.configurableFutureRunner(timeout = Duration(30, SECONDS)),
-        extendedHttpClient = RequestRunners.futureRunner,
```
johnduffell, line 108: we had an http client and an extendedHttpClient but they were both the same in members-data-api.
The extended one has the default (shorter, iirc 10 seconds) timeout, which seems counterintuitive and I don't think it was intentional.
The separation was introcued in this pr to increase the timeout for createSubscription (which was used by subscribe and membership)
https://github.com/guardian/membership-common/pull/322
and pulled into this repo in bulk and with identcal parameter for each
https://github.com/guardian/members-data-api/pull/132/files#diff-246f8135447df9219beb681d466ff8f39bd2afae1699a967ec880ef97a2d39e9R59-R60

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -124,28 +117,21 @@ class TouchpointComponents(
...
+        throw error
+      },
   )
 
-  private lazy val futureCatalog: Future[CatalogMap] = catalogService.catalog
-    .map(_.fold[CatalogMap](error => { logger.errorNoPrefix(scrub"error: ${error.list.toList.mkString}"); Map() }, _.map))
-    .recover { case error =>
-      logger.errorNoPrefix(scrub"Failed to load the product catalog from Zuora due to: $error")
-      throw error
-    }
```
johnduffell, line 141: this code was a bit convoluted so now it's just a small snippet that gives us a `Future[CatalogMap]` without as much fanfare

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
```diff
@@ -84,17 +83,17 @@ class PaymentService(zuoraService: ZuoraSoapService, planMap: Map[ProductRatePla
...
-          .invoices
-          .list
-          .find(_.amount > 0)
-          .toOption
-      }
-    }
+    for {
+      previewInvoiceItems <- zuoraService.previewInvoices(subId, numberOfBills)
+    } yield for {
+      billingSched <- BillingSchedule.fromPreviewInvoiceItems(previewInvoiceItems)
```
johnduffell, line 89: only change is remove planMap from the parameters on this line. 
Also converted the nested code into for comprehensions

membership-attribute-service/test/acceptance/AccountControllerAcceptanceTest.scala
```diff
@@ -202,7 +201,6 @@ class AccountControllerAcceptanceTest extends AcceptanceTest {
       subscriptionServiceMock.current(contact)(any) was called
       zuoraRestServiceMock.getGiftSubscriptionRecordsFromIdentityId("200067388")(any) was called
       subscriptionServiceMock.get(Subscription.Name(giftSubscription.Name), isActiveToday = false)(any) was called
-      catalogServiceMock.unsafeCatalog was called
```
johnduffell, line 205: I would have thought it would only be important to check that side effecting calls were callled or not called?  unless we thought this would be expensive if called unnecessarily - but we only read the catalog once on startup.

membership-common/src/main/scala/com/gu/memsub/BillingSchedule.scala
```diff
@@ -36,13 +36,13 @@ case class BillingSchedule(invoices: NonEmptyList[Bill]) {
 
 object BillingSchedule {
 
-  case class BillItem(name: String, product: Option[Benefit], amount: Float, unitPrice: Float) {
-    override lazy val toString = s"| $amount\t| ($unitPrice)\t| $name | ${product.mkString}"
+  case class BillItem(name: String, amount: Float, unitPrice: Float) {
+    override lazy val toString = s"| $amount\t| ($unitPrice)\t| $name"
```
johnduffell, line 40: this is the toString that used to use the product from the catalog.  But although this might be useful for debugging, it didn't seem worth passing it all the. way through.

membership-common/src/main/scala/com/gu/memsub/Plan.scala
```diff
@@ -3,18 +3,15 @@ package com.gu.memsub
...
-  override val name: String = "current"
-}
-case class Upcoming() extends Status {
-  override val name: String = "upcoming"
-}
-
 object Status {
-  val legacy = Legacy()
-  val current = Current()
-  val upcoming = Upcoming()
```
johnduffell, line 19: I guess someone didn't know how to do Current.type to get the type of a case object, hence it was done this way.

membership-common/src/main/scala/com/gu/memsub/Plan.scala
```diff
@@ -3,18 +3,15 @@ package com.gu.memsub
 sealed trait Status {
   def name: String
 }
-case class Legacy() extends Status {
```
johnduffell, line 6: I guess someone didn't know how to do Current.type to get the type of a case object, hence it was done this way.

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -225,139 +224,15 @@ object FrontendId {
 
 }
 
-object ProductRatePlan {
```
johnduffell, line 228: 🔪  🔪  🔪 unused

membership-common/src/main/scala/com/gu/memsub/subsv2/services/CatalogService.scala
```diff
@@ -42,125 +34,23 @@ object FetchCatalog {
 
 }
 
-class CatalogService[M[_]: Monad](productIds: ProductIds, fetchCatalog: M[String \/ JsValue], unsafeGet: M[Catalog] => Catalog, stage: String) {
```
johnduffell, line 45: 🔪  🔪  🔪 don't need to use the CatPlanReads to construct the high level model as it's not used!

membership-common/src/main/scala/com/gu/memsub/subsv2/services/CatalogService.scala
```diff
@@ -42,125 +34,23 @@ object FetchCatalog {
...
-  ): ValidationNel[String, NonEmptyList[P]] = {
-    val parsed: List[(ProductRatePlanId, ValidationNel[String, P])] =
-      plans.map(plan => (plan.id, p.read(productIds, plan)))
 
-    val failures = parsed
-      .filter { case (_, validation) => validation.isFailure }
-      .map { case (id, err) => (id, err.swap.map(_.list.toList).getOrElse(List.empty).mkString(", ")) }
-      .toMap
+    catalog.map(_.leftMap[CatalogMap](error => {
+      logger.errorNoPrefix(scrub"error: $error"); Map.empty
```
johnduffell, line 52: not sure why it returns an empty catalog if there's an error (a Left)?? but that's what it did before.  Ideally it should fail the health check to avoid broken instances coming in service

membership-common/src/main/scala/com/gu/okhttp/RequestRunners.scala
```diff
@@ -45,13 +45,13 @@ object RequestRunners {
 
   /** Standard no frills run this request and return a response asynchronously A solid choice for the beginner SimpleClient user
     */
-  def futureRunner(implicit ec: ExecutionContext): FutureHttpClient =
+  def futureRunner: FutureHttpClient =
```
johnduffell, line 48: unused

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
```diff
@@ -110,7 +110,7 @@ class ZuoraSoapService(soapClient: soap.ClientWithFeatureSupplier)(implicit ec:
       command.accountId,
       command.paymentGateway,
     ) // We need to set gateway correctly because it must match with the payment method we'll create below
-    createMethodResult <- soapClient.extendedAuthenticatedRequest[CreateResult](new XmlWriterAction(command)(createPaymentMethodWrites))
+    createMethodResult <- soapClient.authenticatedRequest[CreateResult](new XmlWriterAction(command)(createPaymentMethodWrites))
```
johnduffell, line 113: this (createPaymentMethod) is the only call that used the "extended" (actually shorter) timeout

membership-common/src/test/scala/com/gu/memsub/BillingScheduleTest.scala
```diff
@@ -179,30 +176,26 @@ class BillingScheduleTest extends Specification {
     "Include any products found with the productFinder in the bill items" in {
```
johnduffell, line 176: not sure whether to delete this test as there's no product finder (i.e. catalog) any more

membership-common/src/test/scala/com/gu/memsub/subsv2/services/CatalogServiceTest.scala
```diff
@@ -23,12 +22,13 @@ class CatalogServiceTest extends Specification {
     "Read a catalog in CODE with the config product IDs" in {
```
johnduffell, line 22: I had to rewrite this test as it was using the high level catalog model

membership-attribute-service/app/components/TouchpointComponents.scala
rupertbates: Is it loading the catalog from S3? The name makes it sound like it's going to Zuora

johnduffell: yes it does look like it's getting it from zuora (with a 60s timeout) 
I did wonder if the S3 catalog was originally for the benefit of lambdas, but I didn't look into it at this stage.  There's code there to read from S3 but it's not used, but I've kept it just in case.

membership-common/src/main/scala/com/gu/memsub/subsv2/services/CatalogService.scala
```diff
@@ -42,125 +34,23 @@ object FetchCatalog {
...
-  in time, I also expect they will all be tagged in zuora, so we don't need to then try untagged ones
-   */
-  def one[P <: ProductRatePlan[Product, RatePlanChargeList, Status]](plans: List[CatalogZuoraPlan], name: String, frontendId: FrontendId)(implicit
-      p: CatPlanReads[P],
-  ): Validation[NonEmptyList[String], P] = {
-    many(plans.filter(_.frontendId.contains(frontendId)), name).flatMap {
-      case a if a.size == 1 => Validation.s(a.head)
-      case e => Validation.failureNel(s"Too many plans! for name $name with frontendId $frontendId $e")
+    val catalog: M[String \/ CatalogMap] = {
+      fetchCatalog.map(_ => ())
```
johnduffell, line 42: no idea why I added this line, I don't think it does anything?

PR #1079, opened 2024-05-29
remove "high level" subscription model

johnduffell:
This PR removes the "high level" subscription model from members-data-api.

The purpose of the high level model was to combine the subscription, catalog, and product mappings into one thing.  This would be guaranteed valid as a real guardian subscription and have all the info we need in a data structure.

After this PR everything is just blindly read in from zuora as a low level model, and it's up to individual bits of code to pick out what they need in conjunction with the catalog and product ids.

This new way does mean that errors that would be handled quietly in the high level reads (by discarding the plan) would then be surfaced during the business logic.  If the code is looking at the wrong plan for whatever reason, this error may not be recoverable at that point.  I have tried to deal with those situations on a case by case basis, but there will be a lot of holiday credits etc floating around in the business logic that weren't before.

There is probably more work to be done, especially to rename the individual fields to be closer to how they come back from zuora.  Also e.g. we seem to have two fields `ProductType__c` in zuora catalog, one on the product and one on the charge, and used separately.  But all this will be easier once it's all one set of objects.

trello https://trello.com/c/Y7MSKlAj/243-add-to-mdapi-useful-values-for-customers-in-a-free-period

- This is tested and running locally in terms of /me and /me/mma
- tested in CODE
  - for /me/mma with 6day delivery, S+, Contrubution, TierThree(ie GW plus S+)
  - , update contribution amount, cancel contrubtion, change payment method to DD and CC

Tests all compiling and running locally fine with extensive changes made.

(Possible) Future work 
  - `PricingSummary` is only used to read from the subscription not the catalog now, so we could make it only handle a single currency which would simplify things a bit.
  - make SubReadsTest check that it does *not* discard discounts - we need to be able to read discount PricingSummary


membership-attribute-service/app/components/TouchpointBackends.scala
```diff
@@ -22,7 +22,7 @@ class TouchpointBackends(
     contactRepositoryOverride: Option[ContactRepository] = None,
     subscriptionServiceOverride: Option[SubscriptionService[Future]] = None,
     zuoraRestServiceOverride: Option[ZuoraRestService] = None,
-    catalogServiceOverride: Option[Future[CatalogMap]] = None,
+    catalogServiceOverride: Option[Future[Catalog]] = None,
```
johnduffell, line 25: Catalog is now an object containing the product ids as well (to make it easier to pass around)

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -60,7 +60,7 @@ class TouchpointComponents(
 
   lazy val supporterProductDataTable = environmentConfig.getString("supporter-product-data.table")
 
-  lazy val productIds = config.SubsV2ProductIds(environmentConfig.getConfig("zuora.productIds"))
+  lazy val productIds = config.SubsV2ProductIds.load(environmentConfig.getConfig("zuora.productIds"))
```
johnduffell, line 63: reduce the use of `apply` methods

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -129,10 +126,11 @@ class AccountController(
             subscription.termEndDate,
           )
           result = cancellationEffectiveDate.getOrElse("now").toString
+          catalog <- EitherT.rightT(services.futureCatalog)
           _ <- sendSubscriptionCancelledEmail(
             request.user.primaryEmailAddress,
             contact,
-            subscription.plan,
+            subscription.plan(catalog).productType(catalog),
```
johnduffell, line 133: `subscription` was a high level but is now a low level one.
Where fields no longer exist, I have created a similarly named method, which takes the catalog and generates the old object on the fly.

regarding adding `.productType(catalog)` in this case we were passing in the high level rate plan object, but only using the productType, so I've narrowed the method parameter

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -198,16 +196,22 @@ class AccountController(
...
-              logger.warn(s"Unable to retrieve payment details result for identity user $userId due to $message")
-              InternalServerError("Failed to retrieve payment details due to an internal error")
-          }
+        for {
+          catalog <- request.touchpoint.futureCatalog
+          result <- paymentDetails(userId, filter, request.touchpoint).toEither
+            .map {
+              case Right(subscriptionList) =>
+                logger.info(s"Successfully retrieved payment details result for identity user: $userId")
+                val productsResponseWrites = new ProductsResponseWrites(catalog)
```
johnduffell, line 205: toJson's implicit writes (used just below) now need the catalog, so I've put the code in a class `ProductsResponseWrites` and passed the catalog in.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -129,10 +127,11 @@ class AccountController(
             subscription.termEndDate,
           )
           result = cancellationEffectiveDate.getOrElse("now").toString
+          catalog <- EitherT.rightT(services.futureCatalog)
```
johnduffell, line 130: now there's a lot of places we have to flatmap the catalog, I did wonder whether to carry on passing the catalog in to the deserialiser, but instead of using it, just store it in a private field in the Subscription object.  Then when we are reading the old fields (which are now defined on the fly) it could return the values without requiring the catalog as a parameter.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -222,27 +227,30 @@ class AccountController(
     } yield fromZuora ++ fromStripe
   }
 
-  def fetchCancelledSubscriptions(): Action[AnyContent] =
+  def cancelledSubscriptions(): Action[AnyContent] =
```
johnduffell, line 230: I renamed this (and a couple of other methods) because there was a method further down the file that just called this method. (I didn't move the code to the other method as the diff is already big)

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -258,30 +266,36 @@ class AccountController(
...
             case nc => Left(s"$subscriptionName plan is not a contribution: " + nc)
           })
-          billingPeriod = contributionPlan.charges.billingPeriod.asInstanceOf[RecurringPeriod]
-          applyFromDate = contributionPlan.chargedThrough.getOrElse(contributionPlan.start)
-          currency = contributionPlan.charges.price.prices.head.currency
+          billingPeriod <- SimpleEitherT.fromEither(contributionPlan.billingPeriod.toEither)
+          recurringPeriod <- SimpleEitherT.fromEither(billingPeriod match {
+            case period: RecurringPeriod => Right(period)
+            case period: BillingPeriod.OneOffPeriod => Left(s"period $period was not recurring for contribution update")
+          })
```
johnduffell, line 281: `asInstanceOf` should not have been used as it always compiles even if the types are incompatible
I have updated it to use a pattern match which is fully type checked

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -258,30 +266,36 @@ class AccountController(
...
+          applyFromDate = contributionPlan.chargedThroughDate.getOrElse(contributionPlan.start)
+          currency = contributionPlan.charges.head.pricing.prices.head.currency
           currencyGlyph = currency.glyph
-          oldPrice = contributionPlan.charges.price.prices.head.amount
+          oldPrice = contributionPlan.totalChargesMinorUnit.toDouble / 100
```
johnduffell, line 285: contribution only has one charge so the `head` price is the same as the total charges

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -90,11 +87,12 @@ class AccountController(
         Left(badRequest("Malformed request. Expected a valid reason for cancellation."))
       }
 
-  def cancelSubscription(subscriptionName: memsub.Subscription.Name): Action[AnyContent] =
+  def cancelSubscription(subscriptionNameString: String): Action[AnyContent] =
```
johnduffell, line 90: I changed this (and a couple of other methods) because there was a method further down the file that just called this method with a wrapper on the subscription name.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -230,19 +235,22 @@ class AccountController(
...
-            }).getOrElse(emptyResponse).leftMap(_ => emptyResponse).merge // we discard errors as this is not critical endpoint
-
+            for {
+              catalog <- tp.futureCatalog
+              result <-
+                (for {
+                  contact <- OptionT(EitherT(tp.contactRepository.get(identityId)))
+                  subs <- OptionT(EitherT(tp.subscriptionService.recentlyCancelled(contact)).map(Option(_)))
+                } yield {
+                  Ok(Json.toJson(subs.map(CancelledSubscription(_, catalog))))
```
johnduffell, line 245: the only change here is passing in the catalog

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -96,41 +97,40 @@ object AccountDetails {
...
+          charge <- plan.ratePlanCharges.list.toList
+            .filterNot(_.pricing.isFree) // note 'Echo Legacy' rate plan has all days of week but some are zero price, this filters those out
+          catalogZuoraPlan <- catalog.catalogMap.get(plan.productRatePlanId)
+          dayName <- catalogZuoraPlan.productRatePlanCharges
+            .get(charge.productRatePlanChargeId)
+            .collect { case benefit: PaperDay => DayOfWeek.of(benefit.dayOfTheWeekIndex).getDisplayName(TextStyle.FULL, Locale.ENGLISH) }
+        } yield dayName
+
+        if (dayNames.nonEmpty) Json.obj("daysOfWeek" -> dayNames) else Json.obj()
+      }
```
johnduffell, line 117: this code is extracted and rewritten from the code from old line 117 below, that relied on pattern matching the high level PaperCharges.

membership-attribute-service/app/services/GuardianPatronService.scala
```diff
@@ -58,9 +55,9 @@ class GuardianPatronService(
       } yield accountDetailsFromStripeSubscription(subscription, paymentDetails, stripePatronsPublicKey)
     }
 
-  private def billingPeriodFromInterval(interval: String): RecurringPeriod = interval match {
-    case "year" => Year
-    case _ => Month
+  private def billingPeriodFromInterval(interval: String): ZBillingPeriod = interval match {
+    case "year" => ZYear
+    case _ => ZMonth
```
johnduffell, line 60: we used to convert the low level stripe API data into our high level model, however since the high level model is removed, we need to produce the low level model instead.
(This also means adding a fake Guardian Patron to the catalog, so that it can find everything on the fly.)

membership-attribute-service/app/services/GuardianPatronService.scala
```diff
@@ -85,29 +82,29 @@ class GuardianPatronService(
         casActivationDate = None,
         promoCode = None,
         isCancelled = subscription.isCancelled,
-        plans = CovariantNonEmptyList(
+        ratePlans = List(
           RatePlan(
```
johnduffell, line 86: note that `RatePlan` (and `RatePlanCharge` below) are totally different from before.
The old ones were the high level ones and have been deleted.
The new ones are the old low level ones and have been renamed.

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -321,9 +335,6 @@ class AccountController(
     } yield validAmount
   }
 
-  def cancelSpecificSub(subscriptionName: String): Action[AnyContent] =
```
johnduffell, line 324: these last few wrapper methods aren't needed any more as the called method now takes a string directly

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -258,30 +266,36 @@ class AccountController(
...
           })
-          billingPeriod = contributionPlan.charges.billingPeriod.asInstanceOf[RecurringPeriod]
-          applyFromDate = contributionPlan.chargedThrough.getOrElse(contributionPlan.start)
-          currency = contributionPlan.charges.price.prices.head.currency
+          billingPeriod <- SimpleEitherT.fromEither(contributionPlan.billingPeriod.toEither)
+          recurringPeriod <- SimpleEitherT.fromEither(billingPeriod match {
+            case period: RecurringPeriod => Right(period)
+            case period: BillingPeriod.OneOffPeriod => Left(s"period $period was not recurring for contribution update")
+          })
+          applyFromDate = contributionPlan.chargedThroughDate.getOrElse(contributionPlan.start)
```
johnduffell, line 282: this was called `chargedThroughDate` in the low level model (and `chargedThrough` in the high level model), 

The low level model matches what it's called in zuora, so I've gone with that.

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -96,41 +97,40 @@ object AccountDetails {
...
+        val dayNames = for {
+          charge <- plan.ratePlanCharges.list.toList
+            .filterNot(_.pricing.isFree) // note 'Echo Legacy' rate plan has all days of week but some are zero price, this filters those out
+          catalogZuoraPlan <- catalog.catalogMap.get(plan.productRatePlanId)
+          dayName <- catalogZuoraPlan.productRatePlanCharges
+            .get(charge.productRatePlanChargeId)
+            .collect { case benefit: PaperDay => DayOfWeek.of(benefit.dayOfTheWeekIndex).getDisplayName(TextStyle.FULL, Locale.ENGLISH) }
+        } yield dayName
+
+        if (dayNames.nonEmpty) Json.obj("daysOfWeek" -> dayNames) else Json.obj()
```
johnduffell, line 116: instead of pattern matching the result of the logic that was in the ChargeListReads, we have to do it all in this function
Old Reads code was here:
  https://github.com/guardian/members-data-api/blob/1433e4f1800f3b286bab67916b82fb6a7ae81e0b/membership-common/src/main/scala/com/gu/memsub/subsv2/reads/ChargeListReads.scala#L172

membership-attribute-service/app/services/mail/Emails.scala
```diff
@@ -21,15 +21,15 @@ object Emails {
         "first_name" -> contact.firstName.getOrElse(""),
         "last_name" -> contact.lastName,
         "payment_method" -> paymentMethod.valueForEmail,
-        "product_type" -> plan.productType,
+        "product_type" -> productType.productTypeString,
```
johnduffell, line 24: productType is now a value class rather than a string so it needs unpacking when it's serialised

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,121 @@ case object Years extends UpToPeriodsType {
...
+    ratePlanCharges.map(c => (c.pricing.prices.head.amount * 100).toInt).list.toList.sum
+
+  def chargesPrice: PricingSummary =
+    ratePlanCharges
+      .map(_.pricing)
+      .list
+      .toList
+      .reduce((f1, f2) =>
+        PricingSummary(
+          f1.underlying.keySet.intersect(f2.underlying.keySet).map(c => c -> Price(f1.underlying(c).amount + f2.underlying(c).amount, c)).toMap,
```
johnduffell, line 187: this logic is from PricingSemigroup which is now deleted above - membership-common/src/main/scala/com/gu/memsub/PricingSummary.scala

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
```diff
@@ -125,15 +97,15 @@ object Product {
...
+// Benefit is the catalog charge level ProductType__c
+sealed trait ProductRatePlanChargeProductType {
   val id: String
-  val isPhysical: Boolean // i.e. needs delivery address
-  override def toString = s"$id (isPhysical? = $isPhysical)"
+  override def toString = s"benefit $id"
 }
 
-object Benefit {
+object ProductRatePlanChargeProductType {
```
johnduffell, line 106: there are two product type fields in zuora, one on the product and one on the product rate plan charge.  The latter used to be treated as the "benefit" but now it's got a more accurate name (if a little verbose)

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
```diff
@@ -125,15 +97,15 @@ object Product {
   type Contribution = Contribution.type
 }
 
-sealed trait Benefit {
+// Benefit is the catalog charge level ProductType__c
+sealed trait ProductRatePlanChargeProductType {
```
johnduffell, line 101: there are two product type fields in zuora, one on the product and one on the product rate plan charge. The latter used to be treated as the "benefit" but now it's got a more accurate name (if a little verbose)

membership-common/src/main/scala/com/gu/memsub/subsv2/Catalog.scala
```diff
@@ -0,0 +1,17 @@
...
+import com.gu.memsub.Subscription.{ProductRatePlanChargeId, ProductRatePlanId}
+import com.gu.memsub.subsv2.Catalog.CatalogMap
+
+object Catalog {
+
+  type CatalogMap = Map[ProductRatePlanId, ProductRatePlan]
+
+  // dummy ids for stripe (non zuora) products
+  val guardianPatronProductRatePlanId: ProductRatePlanId = ProductRatePlanId("guardian_patron")
+  val guardianPatronProductRatePlanChargeId: ProductRatePlanChargeId = ProductRatePlanChargeId("guardian_patron")
```
johnduffell, line 13: now that we pass the catalog around everywhere to interpret the low level zuora ids, we need to add a fake entry for patron stripe subscriptions, so that they can be looked up by the controllers.

membership-common/src/main/scala/com/gu/memsub/subsv2/services/CatalogService.scala
```diff
@@ -36,21 +40,35 @@ object FetchCatalog {
...
+    Catalog.guardianPatronProductRatePlanId -> ProductRatePlan(
+      Catalog.guardianPatronProductRatePlanId,
+      "Guardian Patron", // was subscription.plan.id but isn't used for patron
+      SubsV2ProductIds.guardianPatronProductId,
+      Map[ProductRatePlanChargeId, ProductRatePlanChargeProductType](
+        Catalog.guardianPatronProductRatePlanChargeId -> ProductRatePlanChargeProductType.GuardianPatron,
+      ),
+      Some(ProductType("Membership")), // not used for patron - only used for payment related emails
+    ),
+  )
```
johnduffell, line 72: this is where we add the fake patron plans to the in-memory catalog.  This means they can be looked up as normal by the controllers etc (rather than trying to shoehorn eithers or coproduct types everywhere)

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,121 @@ case object Years extends UpToPeriodsType {
...
 
 /** Low level model of a Zuora rate plan charge
   */
-case class ZuoraCharge(
+case class RatePlanCharge(
     id: SubscriptionRatePlanChargeId,
     productRatePlanChargeId: ProductRatePlanChargeId,
     pricing: PricingSummary,
-    billingPeriod: Option[ZBillingPeriod],
+    zBillingPeriod: Option[ZBillingPeriod],
```
johnduffell, line 121: this field is renamed and the old low->high level deserialiser is moved to the billingPeriod method below
used to be 
implicit def anyBpReads[B <: BillingPeriod: ClassTag]: ChargeReads[B]
membership-common/src/main/scala/com/gu/memsub/subsv2/reads/ChargeListReads.scala

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,121 @@ case object Years extends UpToPeriodsType {
...
-case class SubscriptionZuoraPlan(
+case class RatePlan(
     id: RatePlanId,
     productRatePlanId: ProductRatePlanId,
     productName: String,
     lastChangeType: Option[String],
     features: List[Feature],
     chargedThroughDate: Option[LocalDate],
-    charges: NonEmptyList[ZuoraCharge],
+    ratePlanCharges: NonEmptyList[RatePlanCharge],
```
johnduffell, line 172: renamed to make it match the zuora response

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,121 @@ case object Years extends UpToPeriodsType {
...
+      .map(_.pricing)
+      .list
+      .toList
+      .reduce((f1, f2) =>
+        PricingSummary(
+          f1.underlying.keySet.intersect(f2.underlying.keySet).map(c => c -> Price(f1.underlying(c).amount + f2.underlying(c).amount, c)).toMap,
+        ),
+      )
+
+  def billingPeriod: Validation[String, BillingPeriod] = {
```
johnduffell, line 191: this came from here
https://github.com/guardian/members-data-api/blob/0909d820448fd6e62abc18106f9124b8bfe443fe/membership-common/src/main/scala/com/gu/memsub/subsv2/reads/ChargeListReads.scala#L190-L195

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,121 @@ case object Years extends UpToPeriodsType {
...
-/** Paper plans will have lots of rate plan charges, but the general structure of them is that they'll give you the paper on a bunch of days, and if
-  * you're on a plus plan you'll have a digipack
-  */
-case class PaperCharges(dayPrices: Map[PaperDay, PricingSummary], digipack: Option[PricingSummary]) extends RatePlanChargeList {
-  def benefits = NonEmptyList.fromSeq[Benefit](dayPrices.keys.head, dayPrices.keys.tail.toSeq ++ digipack.map(_ => Digipack))
-  def price: PricingSummary = (dayPrices.values.toSeq ++ digipack.toSeq).reduce(_ |+| _)
-  override def billingPeriod: BillingPeriod = BillingPeriod.Month
-  def chargedDays = dayPrices.filterNot(_._2.isFree).keySet // Non-subscribed-to papers are priced as Zero on multi-day subs
-  val subRatePlanChargeId = SubscriptionRatePlanChargeId("")
 }
```
johnduffell, line 213: the "diff" is not very helpful here, what's useful to know is the helper methods on these low level classes are simulating fields that used to exist on the high level classes.
This means that we don't need to make many changes to the business logic, and the work of the low->high level deserialisers is moved to these helper methods.

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,121 @@ case object Years extends UpToPeriodsType {
...
 
-/** Supporter Plus V2 has two rate plan charges, one for the subscription element and one for the additional contribution.
+/** Low level model of a product rate plan, as it appears in the Zuora product catalog
```
johnduffell, line 215: this used to be high level, now it's low level

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,121 @@ case object Years extends UpToPeriodsType {
...
-)
+    productId: ProductId,
+    productRatePlanCharges: Map[ProductRatePlanChargeId, ProductRatePlanChargeProductType],
+    private val productTypeOption: Option[ProductType],
+) {
+  lazy val productType: ProductType = productTypeOption.getOrElse(throw new RuntimeException("Product type is undefined for plan: " + name))
+}
+
+// ProductType is the catalog product level ProductType__c
+case class ProductType(productTypeString: String) extends AnyVal
```
johnduffell, line 228: this is as distinct from the one on the charge.  This wrapper just gives us a little more compile time type checking.

membership-common/src/main/scala/com/gu/memsub/subsv2/Subscription.scala
```diff
@@ -84,17 +67,22 @@ object GetCurrentPlans {
...
       val digipackGiftEnded = alreadyStarted.ensure(DiscardedPlan(plan, "has a digipack gift plan which has ended").wrapNel)(_)
-      if (plan.product == Product.Contribution)
+      if (product == Product.Contribution)
         contributorPlanCancelled(_ => !sub.isCancelled && !plan.lastChangeType.contains("Remove"))
-      else if (plan.product == Product.Digipack && plan.charges.billingPeriod == OneTimeChargeBillingPeriod)
+      else if (
+        product == Product.Digipack && plan.billingPeriod
+          .leftMap(e => throw new RuntimeException("no billing period: " + e))
+          .toOption
+          .get == OneTimeChargeBillingPeriod
```
johnduffell, line 84: now that billingPeriod is generated on the fly we have to deal with the error at this point rather than it failing when it's read in originally.

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/CatJsonReads.scala
```diff
@@ -1,91 +1,64 @@
 package com.gu.memsub.subsv2.reads
```
johnduffell, line 1: changes in this file are mostly just tidying up fields that we no longer need and renaming to use more standard zuora terminology

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/CommonReads.scala
```diff
@@ -47,37 +28,8 @@ object CommonReads {
...
-      val parsedPrices = normalisedPricingList.flatMap { priceStrings =>
-        priceStrings
-          .map(PriceParser.parse)
-          .sequence[Option, Price]
-          .toJsSuccess(s"Failed to parse $normalisedPricingList")
-      }
-
-      parsedPrices.map(priceList => priceList.map(p => p.currency -> p).toMap).map(PricingSummary(_))
-    }
-  }
```
johnduffell, line 77: moved to SubJsonReads below as we only read the pricing summary from a subscription, not the catalog any more

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/SubJsonReads.scala
```diff
@@ -72,31 +88,23 @@ object SubJsonReads {
...
+      (json \ "subscriptions").validate(Reads.traversableReads[List, Subscription](implicitly, subscriptionReads))
   }
 
   private val lenientDateTimeReader: Reads[DateTime] =
     JodaReads.DefaultJodaDateTimeReads orElse Reads.IsoDateReads.map(new DateTime(_))
 
-  val subscriptionReads: Reads[NonEmptyList[RatePlan] => Subscription] = new Reads[NonEmptyList[RatePlan] => Subscription] {
-    override def reads(json: JsValue): JsResult[NonEmptyList[RatePlan] => Subscription] = {
+  val subscriptionReads: Reads[Subscription] = new Reads[Subscription] {
+    override def reads(json: JsValue): JsResult[Subscription] = {
```
johnduffell, line 107: this PR meant we managed to get rid of the whole "Reads of a function type" that enabled us to do two layers of deserialisation on the plans.

membership-common/src/main/scala/com/gu/memsub/subsv2/services/CatalogService.scala
```diff
@@ -36,21 +40,35 @@ object FetchCatalog {
...
```
johnduffell, line 72: the main thing is we need to add the (fake) stripe patrons entry into the catalog, so that the business logic that looks at the subscriptions will not fall over on them.  Before we used to populate stripe patrons straight into the high level model, thus bypassing the catalog.

membership-common/src/main/scala/com/gu/memsub/subsv2/services/SubscriptionService.scala
```diff
@@ -98,30 +84,17 @@ class SubscriptionService[M[_]: Monad](pids: ProductIds, futureCatalog: => M[Cat
...
-    for {
-      subJson <- rest.get[JsValue](url)(idReads, logPrefix)
-      catalog <- futureCatalog
-    } yield for {
-      jsValue <- subJson.withLogging(s"get subscription $name from zuora").toOption
-      subscription <- SubscriptionTransform
-        .getSubscription(catalog, pids)(jsValue)
-        .withLogging(s"getAllValidSubscriptionsFromJson for $name")
-        .toOption
-    } yield subscription
```
johnduffell, line 110: we used to have a bit of a rigmarole getting the zuora response as a JsValue, and then running it through the deserliasers (together with the catalog) to get it all the way to the high level model.
This is all gone and we just deserialise the response straight into the (low level) model

membership-common/src/main/scala/com/gu/services/model/PaymentDetails.scala
```diff
@@ -33,37 +32,35 @@ object PaymentDetails {
...
       lastPaymentDate = lastPaymentDate,
       nextPaymentDate = nextPayment.map(_.date),
       nextInvoiceDate = nextInvoiceDate,
       termEndDate = sub.termEndDate,
       paymentMethod = paymentMethod,
-      plan = PersonalPlan.paid(sub),
+      plan = PersonalPlan(
+        name = plan.productName,
+        price = plan.chargesPrice.prices.head,
+        interval = plan.billingPeriod.leftMap(e => throw new RuntimeException("no billing period: " + e)).toOption.get.noun,
```
johnduffell, line 57: this used to be checked by the deserialisation stage, now we find out at this point so we deal with it as an exception

build.sbt
```diff
@@ -92,6 +92,6 @@ val api = Project("membership-attribute-service", file("membership-attribute-ser
     addCommandAlias("batch-load", "runMain BatchLoader"),
     addCommandAlias("play-artifact", "riffRaffNotifyTeamcity"),
   )
-  .dependsOn(`membership-common`)
+  .dependsOn(`membership-common` % "test->test;compile->compile")
```
johnduffell, line 95: this is needed so we can access TestCatalog in membership common from the MDAPI tests.
https://www.scala-sbt.org/1.x/docs/Multi-Project.html#:~:text=A%20useful%20declaration,compile%2D%3Ecompile%22).

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -258,30 +266,36 @@ class AccountController(
...
+          currency = contributionPlan.chargesPrice.prices.head.currency
           currencyGlyph = currency.glyph
-          oldPrice = contributionPlan.charges.price.prices.head.amount
+          oldPrice = contributionPlan.chargesPrice.prices.head.amount
```
johnduffell, line 285: chargesPrice has the logic that was previously in the high level deserialiser (that returns the total pricing summary for all charges in the plan)
The reason it's not kept the old name is that charges is now called ratePlanCharges (so it would be `ratePlanCharges.prices`) and is now a list rather than a high level ChargeList so we couldn't add a dummy prices method without using extension methods.

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -96,41 +97,40 @@ object AccountDetails {
...
-          )
-        case _ => Json.obj()
-      })
-
-      val sortedPlans = subscription.plans.list.sortBy(_.start.toDate)
+        "chargedThrough" -> plan.chargedThroughDate,
+        "price" -> plan.chargesPrice.prices.head.amount * 100,
+        "currency" -> plan.chargesPrice.prices.head.currency.glyph,
+        "currencyISO" -> plan.chargesPrice.prices.head.currency.iso,
+        "billingPeriod" -> plan.billingPeriod.leftMap(e => throw new RuntimeException("no billing period: " + e)).toOption.get.noun,
```
johnduffell, line 129: this would have failed to deserialise in the first place if the billing period wasn't right, now we fail when we actually try to use it (we could just leave it blank in the response alternatively)

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -96,41 +97,40 @@ object AccountDetails {
...
-                .sorted
-                .map(DayOfWeek.of)
-                .map(_.getDisplayName(TextStyle.FULL, Locale.ENGLISH)),
-          )
-        case _ => Json.obj()
-      })
-
-      val sortedPlans = subscription.plans.list.sortBy(_.start.toDate)
+        "chargedThrough" -> plan.chargedThroughDate,
+        "price" -> plan.chargesPrice.prices.head.amount * 100,
```
johnduffell, line 126: FIXME for some reason this has a rounding error that it didn't have in the pre-pr version
image

johnduffell: fixed by adding .toInt in the two places that it converts to pence

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -96,41 +97,42 @@ object AccountDetails {
...
-                .sorted
-                .map(DayOfWeek.of)
-                .map(_.getDisplayName(TextStyle.FULL, Locale.ENGLISH)),
-          )
-        case _ => Json.obj()
-      })
-
-      val sortedPlans = subscription.plans.list.sortBy(_.start.toDate)
+        "chargedThrough" -> plan.chargedThroughDate,
+        "price" -> (plan.chargesPrice.prices.head.amount * 100).toInt,
```
johnduffell, line 128: for some reason we got a .0005p on the sixday paper which we didn't before even though the code was the same, I suspect it was adding them in a different order, and maybe addition on a float is not associative?

membership-attribute-service/app/services/AccountDetailsFromZuora.scala
```diff
@@ -130,7 +134,7 @@ class AccountDetailsFromZuora(
           .getPaymentDetails(contactAndSubscription)
           .map(Right(_))
           .recover { case x =>
-            Left(s"error retrieving payment details for subscription: freeOrPaidSub.name. Reason: $x")
+            Left(s"error retrieving payment details for subscription: ${contactAndSubscription.subscription.name}. Reason: $x")
```
johnduffell, line 137: maybe the $ sign was missed off originally

membership-attribute-service/test/acceptance/AccountControllerAcceptanceTest.scala
```diff
@@ -49,7 +49,7 @@ class AccountControllerAcceptanceTest extends AcceptanceTest {
     contactRepositoryMock = mock[ContactRepository]
     subscriptionServiceMock = mock[SubscriptionService[Future]]
     zuoraRestServiceMock = mock[ZuoraRestService]
-    catalogServiceMock = TestCatalog()
+    catalogServiceMock = catalog
```
johnduffell, line 52: now we use the TestCatalog from membership-common tests instead of the blank dummy one, as the code needs to access the catalog in the business logic in order to get some of the information

membership-attribute-service/test/acceptance/AccountControllerAcceptanceTest.scala
```diff
@@ -149,7 +164,7 @@ class AccountControllerAcceptanceTest extends AcceptanceTest {
       val giftSubscriptionFromSubscriptionService = TestSubscription(
         id = Subscription.Id(giftSubscription.Id),
         name = Subscription.Name(giftSubscription.Name),
-        plans = CovariantNonEmptyList(TestPaidSubscriptionPlan(product = Product.SupporterPlus), Nil),
+        plans = List(TestPaidSubscriptionPlan(productRatePlanId = TestCatalog.digipackPrpId)),
```
johnduffell, line 167: this test was a bit confusing, it was testing
- a gift supporter plus, a contribution, and a patron.
Since you can only have a gift digi sub (or GW) I changed it to test a more realistic situation
- gift digisub, a supporter plus and a patron.

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,123 @@ case object Years extends UpToPeriodsType {
...
+          s"period =${zBillingPeriod} specificBillingPeriod=${specificBillingPeriod} uptoPeriodsType=${upToPeriodsType}, uptoPeriods=${upToPeriods}",
+        )
+    }
+
 }
 
 /** Low level model of a rate plan, as it appears on a subscription in Zuora
   */
-case class SubscriptionZuoraPlan(
+case class RatePlan(
```
johnduffell, line 165: although this is still the low level model, now it has a load of helper methods on it to mimic the fields that were on the old high level model.  The new methods take the catalog as a parameter in most cases.

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/CatJsonReads.scala
```diff
@@ -1,91 +1,64 @@
...
     }
   }
 
-  implicit val statusReads: Reads[Status] = new Reads[Status] {
-    override def reads(json: JsValue): JsResult[Status] = json match {
-      case JsString("Expired") => JsSuccess(Status.Legacy)
-      case JsString("Active") => JsSuccess(Status.Current)
-      case JsString("NotStarted") => JsSuccess(Status.Upcoming)
-      case a => JsError(s"Unknown status $a")
-    }
```
johnduffell, line 59: we no longer do this checking - before, the implicit deserialisers would discard any plans that were not active in the catalog.  However since MDAPI is only working on existing subscriptions, we would only be looking things up rather than searching for one to create.

membership-common/src/test/scala/com/gu/memsub/subsv2/reads/SubReadsTest.scala
```diff
@@ -2,47 +2,40 @@ package com.gu.memsub.subsv2.reads
...
 import org.specs2.mutable.Specification
-import utils.Resource
 import scalaz.NonEmptyList
+import utils.Resource
 
 class SubReadsTest extends Specification {
 
   "Subscription JSON reads" should {
 
     "Discard discount rate plans when reading JSON" in {
```
johnduffell, line 18: this only discards them because the Pricing Summary can't parse discounts so the whole rateplan is discarded.
If it's a discount amount instead, it wil be retained.  This test will have to change, probably in a later PR.
https://github.com/guardian/members-data-api/blob/729767723fe998ad888fc8858f83e9619b0e2bad/membership-common/src/test/resources/rest/plans/Promo.json#L64

membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -258,30 +266,36 @@ class AccountController(
...
           reasonForChange =
             s"User updated contribution via self-service MMA. Amount changed from $currencyGlyph$oldPrice to $currencyGlyph$newPrice effective from $applyFromDate"
           result <- SimpleEitherT(
             services.zuoraRestService.updateChargeAmount(
               subscription.name,
-              contributionPlan.charges.subRatePlanChargeId,
+              contributionPlan.ratePlanCharges.head.id,
```
rupertbates, line 291: Do we know that ratePlanCharges has a head value?

rupertbates (approved):
Wow, that is a lot of changes. It looks reasonable to me as long as the discount issue is now fixed. Well done for sticking with this!

membership-attribute-service/app/controllers/AccountController.scala
johnduffell: yes good point, in reality there is always (exactly) one charge in the rateplan which would be the contribution.
If a broken contribution subscription came in then they would not even be able to see it in manage, let alone update the charge amount.
In the old system it would fail to deserialise the broken contribution, now it would only notice the issue when it comes to use it at this point.
It does raise a good point (along with the fact that your paper subscription found an issue).  Before if there was an issue with the data it was discovered by the deserialiser and the subscription was usually silently discarded.  Now if there is an issue, it will trip up when it tries to use it.  I wonder how many broken subscriptions (or incomplete traits etc) we have that slip under the radar.

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,126 @@ case object Years extends UpToPeriodsType {
...
+  def productType(catalog: Catalog): ProductType =
+    productRatePlan(catalog).productType
 
-/** Paper plans will have lots of rate plan charges, but the general structure of them is that they'll give you the paper on a bunch of days, and if
-  * you're on a plus plan you'll have a digipack
-  */
-case class PaperCharges(dayPrices: Map[PaperDay, PricingSummary], digipack: Option[PricingSummary]) extends RatePlanChargeList {
-  def benefits = NonEmptyList.fromSeq[Benefit](dayPrices.keys.head, dayPrices.keys.tail.toSeq ++ digipack.map(_ => Digipack))
-  def price: PricingSummary = (dayPrices.values.toSeq ++ digipack.toSeq).reduce(_ |+| _)
-  override def billingPeriod: BillingPeriod = BillingPeriod.Month
```
johnduffell, line 259: actually not all papers are monthly, I found some migrated in Echo Legacy that are still running and are actually quarterly

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -122,178 +108,126 @@ case object Years extends UpToPeriodsType {
...
+            .toMap,
+        ),
+      )
+
+  def billingPeriod: Validation[String, BillingPeriod] = {
+    val billingPeriods = ratePlanCharges.list.toList.map(c => c.billingPeriod).distinct
+    billingPeriods match {
+      case Nil => Validation.f[BillingPeriod]("No billing period found")
+      case b :: Nil => b
+      case _ => Validation.f[BillingPeriod]("Too many billing periods found")
```
johnduffell, line 201: I checked bigquery and there were only 5 active subscriptions with rateplans containing mixed billing periods, all were quarterly migrated in echo legacy subscriptions (the monthly charges were all zero)
before it would have been incorrectly hard coded as monthly, now it will end up as unknown_billing_period.
I did consider filtering the free charges out, but I found there are quite a lot of free charges in the system, mainly recurring contributions that have been set to zero, but a few digi sub and others.

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -135,17 +135,12 @@ object AccountDetails {
...
 
-      val sortedPlans = subscription.ratePlans.sortBy(_.start.toDate)
-      val currentPlans = sortedPlans.filter(plan => !plan.start.isAfter(now) && plan.end.isAfter(now))
-      val futurePlans = sortedPlans.filter(plan => plan.start.isAfter(now))
-
-      val startDate: LocalDate = sortedPlans.headOption.map(_.start).getOrElse(paymentDetails.customerAcceptanceDate)
-      val endDate: LocalDate = sortedPlans.headOption.map(_.end).getOrElse(paymentDetails.termEndDate)
-
-      if (currentPlans.length > 1) logger.warn(s"More than one 'current plan' on sub with id: ${subscription.id}")
+      val subscriptionData = new FilterPlans(subscription, catalog)
```
johnduffell, line 138: extracted for testability

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -229,6 +224,28 @@ object AccountDetails {
...
+class FilterPlans(subscription: Subscription, catalog: Catalog)(implicit val logPrefix: LogPrefix) extends SafeLogging {
+
+  private val sortedPlans = subscription.ratePlans
+    .filter(_.product(catalog) match {
+      case _: Product.ContentSubscription => true
+      case Product.UnknownProduct => false
+      case Product.Membership => true
+      case Product.GuardianPatron => true
+      case Product.Contribution => true
+      case Product.Discounts => false
```
johnduffell, line 236: discounts used to be filtered out in the high level reads, now they must be removed before we write the plans out to the client

johnduffell:
this has been tested in PROD for a couple of 1 hour periods 24/25th July, plus since yesterday morning 29th July, with no new issues seen via fastly logs, searching the cloudwatch logs, and checking sentry.

PR #1081, opened 2024-06-05
Add feast android offer tags

rupertbates:
### Why do we need this? 
This PR adds 'offer tags' to the /user-attributes/me response to enable the Android Feast app to work out whether to offer users an extended free trial.

### Testing
To test this you can deploy this build to CODE and then sign in to 
https://manage.code.dev-theguardian.com/
with the username rupert.bates+supporter-plus-only@observer.co.uk and password rupert.bates+supporter-plus-only
and then check the members data api response at 
https://members-data-api.code.dev-theguardian.com/user-attributes/me
The `feast` boolean in the `contentAccess` object should be `true`

Next sign in with the username rupert.bates+paper-only@observer.co.uk and password rupert.bates+paper-only
and check the mdapi response
The `feast` boolean should be `false` and the response should have a `feastAndroidOfferTags` attribute at the top level with the value `initial_supporter_launch_offer`

### DIFF
--- a/membership-attribute-service/app/models/FeastApp.scala
+++ b/membership-attribute-service/app/models/FeastApp.scala
@@ -16,7 +16,7 @@ object FeastApp {
 
   object AndroidOfferTags {
     // Offer tags are the Android equivalent of iOS subscription groups - used by the app to work out which offer to show to the user
-    val InitialSupporterLaunchOffer = "initial_supporter_launch_offer"
+    val ExtendedTrial = "initial_supporter_launch_offer"
   }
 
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/models/FeastApp.scala
```diff
@@ -14,6 +14,11 @@ object FeastApp {
     val RegularSubscription = "21396030"
   }
 
+  object AndroidOfferTags {
+    // Offer tags are the Android equivalent of iOS subscription groups - used by the app to work out which offer to show to the user
+    val ExtendedTrial = "initial_supporter_launch_offer"
```
jamesgorrie, line 19: Where does this get set? Is it in the app?

jamesgorrie (approved):

PR #1082, opened 2024-06-14
Add new Tier Three product to the mdapi MMA response

rupertbates:
### Why do we need this? 
This PR adds the new Tier Three (T3) product into the `/user-attributes/me/mma` endpoint with the tier value of 'Tier Three'.
There is an example response [in this json document](https://github.com/user-attachments/files/15838172/mma-t3.json)

To view an account with one of these subs in CODE MMA you can use the login:
username: rupert.bates+t3@observer.co.uk
password: rupert.bates+t3


membership-attribute-service/test/integration/AccountDetailsFromZuoraIntegrationTest.scala
```diff
@@ -23,14 +24,15 @@ class AccountDetailsFromZuoraIntegrationTest extends Specification {
...
-            .fetch("200152344", AccountHelpers.NoFilter)
+            .fetch("200264404", AccountHelpers.NoFilter)
             .run
             .map { list =>
               println(s"Fetched this list: $list")
               list
             },
           Duration.Inf,
         )
+      result.map(_.foreach(accountDetails => println(Json.prettyPrint(accountDetails.toJson))))
```
jamesgorrie, line 35: Assuming we don't want this debugging?

rupertbates: This is an integration test and it's set up to be ignored, so the only time I run it is when I'm debugging stuff and then I usually do want the logging

membership-common/conf/touchpoint.PROD.conf
```diff
@@ -66,6 +66,7 @@ touchpoint.backend.environments {
                     nationalDelivery="8a12999f8a268c57018a27ebddab1460"
                     digipack="2c92a0fb4edd70c8014edeaa4ddb21e7"
                     supporterPlus="8a12865b8219d9b4018221061563643f"
+                    tierThree="8a1295998ff2ec180190024b287b64c7"
```
johnduffell, line 69: not for this PR but next time we're refactoring we can move these into string literals like we do in other codebases rather than trying to pick them out of the config file by key

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -216,6 +216,7 @@ object AccountDetails {
     case _: Product.Paper => "subscriptions" // Paper includes GW 🤦‍
     case _: Product.ZDigipack => "subscriptions"
     case _: Product.SupporterPlus => "recurringSupport"
+    case _: Product.TierThree => "recurringSupport"
```
johnduffell, line 219: there was a question about this (for S+) and how that affects things in manage, @rBangay  woud know more about the implications.

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
```diff
@@ -206,6 +212,11 @@ object Benefit {
     override val isPhysical: Boolean = true
   }
 
+  object TierThree extends PaidMemberTier {
```
johnduffell, line 215: it's not really a member tier (although this is all gone in  this PR)
https://github.com/guardian/members-data-api/pull/1079/files#diff-8434c9ffdcbb8fc26673d1f84012690fee7a77abf6ab033c73335a7b03a32f1aL141

johnduffell (approved):
looks good, hope once we get the other (draft) pr merged it will be even simpler

PR #1083, opened 2024-06-14
Change tier three to be a separate product in user-attributes/me

rupertbates:
### Why do we need this? 
This PR makes some changes to support the new Tier Three product, following on from [this PR](https://github.com/guardian/members-data-api/pull/1065), now that Tier Three is a separate product in Zuora and not just a new rate plan in Supporter Plus.


jamesgorrie (approved):
Tests make this really nice!

PR #1084, opened 2024-06-21
Switch which GoCardless Zuora gateway is used when user updates Direct Debit details in MMA

paulbrown1982:
### Why do we need this? 
Currently the GoCardless reconciliation batch job (https://knowledgecenter.zuora.com/Zuora_Payments/Payment_gateway_integrations/Batch_Gateway_Reconciliation) happens twice because we have 2 GoCardless payment gateways configured in Production Zuora. I'd like to reduce down to having 1, this change is a prerequisite to ensure no new payment methods are created referencing this gateway. Salesforce HPP may also need to be updated.

### The changes 
Setting the gateway to `GoCardless` for Direct Debits rather than `GoCardless - Zuora Gateway` so the latter one (with far fewer accounts using it) can be deprovisioned.

https://trello.com/c/uaCj9MdG


johnduffell (approved):
good and the description is helpful too 👍

PR #1086, opened 2024-06-24
add potential cancellation date field to the /mma response

johnduffell:
Manage frontend uses the next payment date to be the cancellation date.

This is not the case if we are in a multi period free period, as there are several (zero) invoice dates before the next payment.

This PR adds another field to say what the next invoice date is, as opposed to the next actual payment.

This can then be used on the client side to show the user the correct date they will be cancellled.


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -164,12 +164,12 @@ object AccountDetails {
             "end" -> endDate,
             "nextPaymentPrice" -> paymentDetails.nextPaymentPrice,
             "nextPaymentDate" -> paymentDetails.nextPaymentDate,
+            "potentialCancellationDate" -> paymentDetails.nextInvoiceDate,
             "lastPaymentDate" -> paymentDetails.lastPaymentDate,
             "chargedThroughDate" -> paymentDetails.chargedThroughDate,
             "renewalDate" -> paymentDetails.termEndDate,
             "anniversaryDate" -> anniversary(startDate),
             "cancelledAt" -> paymentDetails.pendingCancellation,
-            "subscriberId" -> paymentDetails.subscriberId, // TODO remove once nothing is using this key (same time as removing old deprecated endpoints)
```
johnduffell, line 172: I removed this while. I was in, as it's not used by manage and was confusing things
https://github.com/guardian/manage-frontend/blob/98923b8d2735de2bd2b4343b5b7701852118c1c6/shared/productResponse.ts#L168
cc @rBangay

membership-attribute-service/app/services/AccountDetailsFromZuora.scala
```diff
@@ -67,7 +67,7 @@ class AccountDetailsFromZuora(
         isAutoRenew = isAutoRenew,
         alertText = alertText,
         accountId = accountSummary.id.get,
-        effectiveCancellationDate,
+        cancellationEffectiveDate = effectiveCancellationDate,
```
johnduffell, line 70: existing field - make it explicit like all the rest

membership-attribute-service/app/services/GuardianPatronService.scala
```diff
@@ -122,6 +122,7 @@ class GuardianPatronService(
         nextPaymentPrice = Some(subscription.plan.amount),
         lastPaymentDate = Some(subscription.currentPeriodStart),
         nextPaymentDate = subscription.nextPaymentDate,
+        nextInvoiceDate = subscription.nextPaymentDate,
```
johnduffell, line 125: don't have a suspension/free months for patron so it matches the payment date

membership-attribute-service/app/services/PaymentDetailsForSubscription.scala
```diff
@@ -31,6 +31,7 @@ class PaymentDetailsForSubscription(paymentService: PaymentService) extends Safe
     nextPaymentPrice = None,
     lastPaymentDate = None,
     nextPaymentDate = None,
+    nextInvoiceDate = None,
```
johnduffell, line 34: gifts are not invoiced as they are paid up front

membership-common/src/main/scala/com/gu/services/model/PaymentDetails.scala
```diff
@@ -26,10 +27,11 @@ object PaymentDetails {
   case class Payment(price: Price, date: LocalDate)
   implicit def dateToDateTime(date: LocalDate): DateTime = date.toDateTimeAtStartOfDay()
 
-  def apply(
+  def fromSubAndPaymentData(
```
johnduffell, line 30: renamed this, the confusion between the default apply method and this one could be easily resolved that way

shtukas (approved):

rupertbates:
> Manage frontend uses the next payment date to be the cancellation date.
> 
> This is not the case if we are in a multi period free period, as there are several (zero) invoice dates before the next payment.

What is an example of a multi period free period?

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -164,12 +164,12 @@ object AccountDetails {
             "end" -> endDate,
             "nextPaymentPrice" -> paymentDetails.nextPaymentPrice,
             "nextPaymentDate" -> paymentDetails.nextPaymentDate,
+            "potentialCancellationDate" -> paymentDetails.nextInvoiceDate,
```
rupertbates, line 167: Why did you choose `potentialCancellationDate` over `nextInvoiceDate`?

johnduffell:
> > Manage frontend uses the next payment date to be the cancellation date.
> > This is not the case if we are in a multi period free period, as there are several (zero) invoice dates before the next payment.
> 
> What is an example of a multi period free period?

basically "2 months free" cancellation offer.
So it specifically wouldn't include the 14 days free trial because that's not more than one billing period.

membership-attribute-service/app/models/AccountDetails.scala
rBangay: nice one, I'll remove the old references to this in manage as well 👍

membership-attribute-service/app/models/AccountDetails.scala
johnduffell: I wanted it to be clear why it was useful for the user rather than where it came from, (the reason this field is added today is because we reused the nextPaymentDate on the client side as the cancellation date).

As a later PR, we should probably flip this to use the actual cancellation calculation (which is in the "decideCancellationDate" function) as that is what's actually used when you ask MDAPI to cancel a sub.

Happy to take suggestions though.

johnduffell:
I've merged this but that doesn't make it a done deal so I'm happy to do followup PRs

rupertbates:
> I've merged this but that doesn't make it a done deal so I'm happy to do followup PRs

All good with me

membership-attribute-service/app/models/AccountDetails.scala
rBangay: https://github.com/guardian/manage-frontend/pull/1346

PR #1087, opened 2024-06-24
add inDiscountPeriod flag to /mma response

johnduffell:
The manage page needs to flag up to users whether their subscription is discounted or not.

This PR adds a flag so that it can be guessed whether it's in the "cancellation save" free period.

--

As future work, once the MDAPI refactor in another PR https://github.com/guardian/members-data-api/pull/1079 is finished, it will be easier to identify discounts in the code and return the name or other details in the API response.


johnduffell:
this was done on the client side in the end

PR #1088, opened 2024-06-26
fix test users to use email rather than first name

johnduffell:
Apparently test users has been broken for a few years in manage.

Traditiainlly we have expected test users to put the special key in the first name field.  This should come through to the display name (along with the last name field separated by a space)

However since 2020 the display name has been populated by "user" by default, meaning that test users stopped working https://github.com/guardian/identity/pull/1701

This PR switches it to look in the email address instead.
The recommended format is `something+KEY@blah` or `KEY@blah` and it should handle both

### DIFF
--- a/membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
+++ b/membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
@@ -2,7 +2,7 @@ package actions
 
 import com.gu.identity.auth.AccessScope
 import components.TouchpointBackends
-import filters.IsTestUser
+import filters.TestUserChecker
 import play.api.mvc.{ActionRefiner, Request, Result, Results}
 import services.AuthenticationFailure
 
@@ -11,7 +11,7 @@ import scala.concurrent.{ExecutionContext, Future}
 class AuthAndBackendViaAuthLibAction(
     touchpointBackends: TouchpointBackends,
     requiredScopes: List[AccessScope],
-    isTestUser: IsTestUser,
+    testUserChecker: TestUserChecker,
 )(implicit
     ex: ExecutionContext,
 ) extends ActionRefiner[Request, AuthenticatedUserAndBackendRequest] {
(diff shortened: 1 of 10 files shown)


membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
```diff
@@ -2,7 +2,7 @@ package actions
 
 import com.gu.identity.auth.AccessScope
 import components.TouchpointBackends
-import filters.IsTestUser
+import filters.TestUserChecker
```
johnduffell, line 5: I renamed this to make it less surprising, rather than having `IsTestUser` class with an `apply` method (as it sounds like a boolean)

membership-attribute-service/app/actions/AuthAndBackendViaAuthLibAction.scala
```diff
@@ -22,7 +22,7 @@ class AuthAndBackendViaAuthLibAction(
       case Left(AuthenticationFailure.Unauthorised) => Left(Results.Unauthorized)
       case Left(AuthenticationFailure.Forbidden) => Left(Results.Forbidden)
       case Right(authenticatedUser) =>
-        val backendConf = if (isTestUser(authenticatedUser.username)) {
+        val backendConf = if (testUserChecker.isTestUser(authenticatedUser.primaryEmailAddress)(authenticatedUser.logPrefix)) {
```
johnduffell, line 25: now when a test user is detected, it is written to the logs together with its identity id

membership-attribute-service/app/filters/TestUserChecker.scala
```diff
@@ -0,0 +1,23 @@
+package filters
```
johnduffell, line 1: this file is the old IsTestUser.scala above, and it does the main work

membership-attribute-service/test/filters/AddGuIdentityHeadersTest.scala
```diff
@@ -27,14 +28,17 @@ class AddGuIdentityHeadersTest extends Specification with IdiomaticMockito {
...
-  val isTestUser = new IsTestUser(testUsernames)
+  val testUserChecker = new TestUserChecker(testUsernames)
 
-  val addGuIdentityHeaders = new AddGuIdentityHeaders(identityService, isTestUser)
+  def setup = {
+    val identityService = mock[IdentityAuthService]
+    when(identityService.user(requiredScopes = Nil)(request)).thenReturn(Future.successful(Right(user)))
+    val addGuIdentityHeaders = new AddGuIdentityHeaders(identityService, testUserChecker)
+    (identityService, addGuIdentityHeaders)
+  }
```
johnduffell, line 41: interestingly this test was failing randomly because specs2 only creates one test instance per execution and runs the tests in parallel.  This means the mock identity service was sometimes called by the second test before it was asserted by the first.
This technique was suggested here: https://stackoverflow.com/a/56251734/2076522

rupertbates (approved):
👍

johnduffell:
tested in CODE and seems ok for /me and /me/mma for a non test user

PR #1089, opened 2024-06-26
fix test users to use email rather than first name

johnduffell:
following on from #1088 which didn't work
This PR pulls in the updated test user library that will only use lower case tokens - https://github.com/guardian/identity-test-users/pull/10

I have tested locally using the browser direct to MDAPI and the X-Gu-Membership-Test-User header is correctly set to true or false depending on whether it's a test user.

If I try to get manage running locally, the /me/mma response is now correct (however I seem to have a setup issue causing an Oops error)

Testing:

If I hit MDAPI I can see the test users line only appears in the logs when I'm a test user 
`2024-07-09 12:55:12,291 TestUserChecker.scala:11 [INFO]: 123456: redacted.user+oudfhj53j323@guardian.co.uk is a test user`
(in DEV, the normal and test environments are the same so it's harder to test)

### DIFF
--- a/membership-attribute-service/app/filters/TestUserChecker.scala
+++ b/membership-attribute-service/app/filters/TestUserChecker.scala
@@ -6,18 +6,11 @@ import com.gu.monitoring.SafeLogging
 
 class TestUserChecker(testUsernames: TestUsernames) extends SafeLogging {
   def isTestUser(primaryEmailAddress: String)(implicit logPrefix: LogPrefix): Boolean = {
-    val maybeValidTestUser = for {
-      localPart <- primaryEmailAddress.split('@').headOption
-      possibleTestUsername <- localPart.split('+').toList match {
-        case _ :: subAddress :: _ => Some(subAddress)
-        case noPlus :: Nil => Some(noPlus)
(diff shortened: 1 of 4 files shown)


membership-attribute-service/app/filters/TestUserChecker.scala
```diff
@@ -6,18 +6,11 @@ import com.gu.monitoring.SafeLogging
...
   def isTestUser(primaryEmailAddress: String)(implicit logPrefix: LogPrefix): Boolean = {
-    val maybeValidTestUser = for {
-      localPart <- primaryEmailAddress.split('@').headOption
-      possibleTestUsername <- localPart.split('+').toList match {
-        case _ :: subAddress :: _ => Some(subAddress)
-        case noPlus :: Nil => Some(noPlus)
-        case _ => None // invalid email address - no @ sign
-      }
-    } yield testUsernames.isValid(possibleTestUsername)
-    if (maybeValidTestUser.contains(true)) {
```
johnduffell, line 17: this logic is in the library now so it can be shared between here and membership-workflow

tjmw (approved):
Great! 👍

PR #1092, opened 2024-07-30
add logging for dd update failures

johnduffell:
sometimes we throw 500 errors due to failed DD updates, but they look ok to me.

Users think that it has failed and retry, but it's actually ok by the look.s

this PR updates the logging so we have a better idea what went wrong.

I also removed some redundant identity ids, as all logging includes an identity id as a prefix now.


membership-attribute-service/app/controllers/PaymentUpdateController.scala
```diff
@@ -116,11 +115,13 @@ class PaymentUpdateController(
         ),
       )
     case Some(_) =>
-      logger.error(scrub"New payment method for user $userId, does not match the posted Direct Debit details")
+      logger.error(
+        scrub"New payment method $freshDefaultPaymentMethodOption, does not match the posted Direct Debit details $bankSortCode $bankAccountNumber $bankAccountName",
```
johnduffell, line 119: this is the important added info

rupertbates (approved):

PR #1093, opened 2024-08-05
filter out subscriptions with unknown products e.g. staff memberships

johnduffell:
This PR follows on from the recent refactor PRs.

Staff membership is no longer defined or understood by MDAPI.

Previously this meant it was ignored.

However since the high level model was removed, all subscriptions are deserialised and retutned to the business logic, and the business logic is supposed to ignore what it doesn't like.  However in this case, the defunct staff membership was being returned to manage, with an `"mmaCategory": "UNKNOWN_PRODUCT"`. This broke the whole page!

This PR updates it to filter out all subscriptions that only have discounts and/or unknown products added.

Tested in CODE and seems to be working ok and displays the manage frontend.

### DIFF
--- a/membership-attribute-service/app/controllers/AccountController.scala
+++ b/membership-attribute-service/app/controllers/AccountController.scala
@@ -200,18 +200,17 @@ class AccountController(
         for {
           catalog <- request.touchpoint.futureCatalog
           result <- paymentDetails(userId, filter, request.touchpoint).toEither
-            .map {
-              case Right(subscriptionList) =>
-                logger.info(s"Successfully retrieved payment details result for identity user: $userId")
-                val productsResponseWrites = new ProductsResponseWrites(catalog)
-                val response = productsResponseWrites.from(user, subscriptionList)
-                import productsResponseWrites.writes
-                Ok(Json.toJson(response))
-              case Left(message) =>
-                logger.warn(s"Unable to retrieve payment details result for identity user $userId due to $message")
-                InternalServerError("Failed to retrieve payment details due to an internal error")
-            }
-        } yield result
+        } yield result match {
+          case Right(subscriptionList) =>
+            logger.info(s"Successfully retrieved payment details result for identity user: $userId")
+            val productsResponseWrites = new ProductsResponseWrites(catalog)
+            val response = productsResponseWrites.from(user, subscriptionList)
+            import productsResponseWrites.writes
+            Ok(Json.toJson(response))
+          case Left(message) =>
+            logger.warn(s"Unable to retrieve payment details result for identity user $userId due to $message")
+            InternalServerError("Failed to retrieve payment details due to an internal error")
+        }
 
       }
     }
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -214,13 +214,13 @@ object AccountDetails {
   }
   def mmaCategoryFrom(product: Product): String = product match {
     case _: Product.Paper => "subscriptions" // Paper includes GW 🤦‍
-    case _: Product.ZDigipack => "subscriptions"
-    case _: Product.SupporterPlus => "recurringSupport"
-    case _: Product.TierThree => "recurringSupport"
-    case _: Product.GuardianPatron => "subscriptions"
-    case _: Product.Contribution => "recurringSupport"
-    case _: Product.Membership => "membership"
-    case _ => product.name // fallback
+    case Product.Digipack => "subscriptions"
+    case Product.SupporterPlus => "recurringSupport"
+    case Product.TierThree => "recurringSupport"
+    case Product.GuardianPatron => "subscriptions"
+    case Product.Contribution => "recurringSupport"
+    case Product.Membership => "membership"
+    case _ => "subscriptions" // fallback - passing undefined value breaks manage
   }
 }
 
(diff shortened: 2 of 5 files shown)


membership-attribute-service/app/controllers/AccountController.scala
```diff
@@ -200,18 +200,17 @@ class AccountController(
...
-                logger.info(s"Successfully retrieved payment details result for identity user: $userId")
-                val productsResponseWrites = new ProductsResponseWrites(catalog)
-                val response = productsResponseWrites.from(user, subscriptionList)
-                import productsResponseWrites.writes
-                Ok(Json.toJson(response))
-              case Left(message) =>
-                logger.warn(s"Unable to retrieve payment details result for identity user $userId due to $message")
-                InternalServerError("Failed to retrieve payment details due to an internal error")
-            }
-        } yield result
```
johnduffell, line 214: just a quick tidy up to use yield rather than map directly

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -214,13 +214,13 @@ object AccountDetails {
...
-    case _: Product.Contribution => "recurringSupport"
-    case _: Product.Membership => "membership"
-    case _ => product.name // fallback
+    case Product.Digipack => "subscriptions"
+    case Product.SupporterPlus => "recurringSupport"
+    case Product.TierThree => "recurringSupport"
+    case Product.GuardianPatron => "subscriptions"
+    case Product.Contribution => "recurringSupport"
+    case Product.Membership => "membership"
+    case _ => "subscriptions" // fallback - passing undefined value breaks manage
```
johnduffell, line 223: this line means that all unknown products will be treated as subscriptions (although they will have already been filtered out earlier in the logic)

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -214,13 +214,13 @@ object AccountDetails {
...
-    case _: Product.GuardianPatron => "subscriptions"
-    case _: Product.Contribution => "recurringSupport"
-    case _: Product.Membership => "membership"
-    case _ => product.name // fallback
+    case Product.Digipack => "subscriptions"
+    case Product.SupporterPlus => "recurringSupport"
+    case Product.TierThree => "recurringSupport"
+    case Product.GuardianPatron => "subscriptions"
+    case Product.Contribution => "recurringSupport"
+    case Product.Membership => "membership"
```
johnduffell, line 222: we were using type aliases before which just complicated things.

membership-attribute-service/app/services/AccountDetailsFromZuora.scala
```diff
@@ -110,7 +110,15 @@ class AccountDetailsFromZuora(
...
       catalog <- SimpleEitherT.rightT(futureCatalog)
-      filtered = applyFilter(filter, contactAndSubscriptions, catalog)
+      subsWithRecognisedProducts = contactAndSubscriptions.filter(_.subscription.plan(catalog).product(catalog) match {
+        case _: Product.ContentSubscription => true
+        case Product.UnknownProduct => false
+        case Product.Membership => true
+        case Product.GuardianPatron => true
+        case Product.Contribution => true
+        case Product.Discounts => false
+      })
```
johnduffell, line 120: this is the important bit, it filters out any that are unknown product or discounts as the primary product of the subscription

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
```diff
@@ -87,21 +87,6 @@ object Product {
...
-  type Delivery = Delivery.type
-  type NationalDelivery = NationalDelivery.type
-  type Voucher = Voucher.type
-  type DigitalVoucher = DigitalVoucher.type
-  type WeeklyZoneA = WeeklyZoneA.type
-  type WeeklyZoneB = WeeklyZoneB.type
-  type WeeklyZoneC = WeeklyZoneC.type
-  type WeeklyDomestic = WeeklyDomestic.type
-  type WeeklyRestOfWorld = WeeklyRestOfWorld.type
-  type Contribution = Contribution.type
```
johnduffell, line 104: these were unnecessary and actually made the logic less clear.

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -214,13 +214,13 @@ object AccountDetails {
   }
   def mmaCategoryFrom(product: Product): String = product match {
     case _: Product.Paper => "subscriptions" // Paper includes GW 🤦‍
```
tjmw, line 216: Should this line have been converted too?

tjmw (approved):

membership-attribute-service/app/models/AccountDetails.scala
johnduffell: no, good spot but paper is a trait

johnduffell: see
https://github.com/guardian/members-data-api/blob/d80b77bac1af42560a40299dbdd511cadcb418f4/membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala#L17

membership-attribute-service/app/models/AccountDetails.scala
rBangay: I wonder whether this case ought to be handled in the client. Presumably the scenario where a product ends up with it's 'mmaCategory' property set to "subscriptions" by default is an error. The client (manage-frontend) uses the mmaCategory as a visual filter on the account overview, but also to mark selected products for specific functionality, like in the snappy named `augmentProductDetailWithDeliveryAddressChangeEffectiveDateForToday` function.

I guess at a higher level, what do we expect manage-frontend to render if its get's passed one of these products?

johnduffell: good question - in actual fact the logic should filter them out before this line, but the compiler doesn't know that (any more since I removed the high level model)
I would have expected manage to just put all the unknown ones at the bottom, but I'd rather it put them in the wrong place than crashed completely.
I would prefer manage to either hide them completely, or more ideally print the subscription id and mark it as unknown in some way.

rBangay: `or more ideally print the subscription id and mark it as unknown in some way`

I like this idea

PR #1096, opened 2024-09-02
feat: remove hard-coded invoice template IDs

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/h7MX4D5w/260-remove-all-references-to-hard-coded-invoice-template-ids-and-move-all-customers-to-new-invoice-template)

This removes all references to hard-coded invoice template IDs.

## Why are you doing this?

We have recently created a [generic HTML template](https://github.com/guardian/zuora-config/blob/main/finance/invoice-templates/system-default.html) that will be used by default for all products / accounts.


rupertbates (approved):
Looks good 👍

PR #1097, opened 2024-09-24
Add ids for new T3 newspaper archive rate plans

rupertbates:
members-data-api needs to know the productRatePlanIds of any plan which we sell, so that it can work out what benefits a user is entitled to. This PR adds the ids for the four new newspaper archive rate plans in the Tier Three product.


paul-daniel-dempsey (approved):
LG

PR #1098, opened 2024-09-25
Fix filter plans spec

rupertbates:
The FilterPlansSpec test has started failing on main because it is testing the result of a filtering function against a hard coded list. 

The filtering function checks that the ratePlans it is filtering are still current (not expired) so when the end date in the test data is in the past (happened on 23rd Sept 2024) the test starts failing.

This PR puts the ratePlan end dates in the far future so the test will pass.


johnduffell (approved):
ideally we should be passing in `today` as a hard coded value in the tests, but this should keep us going for another few years!

PR #1100, opened 2024-11-05
update Subscription structure to match zuora more closely

johnduffell:
This PR just updates the fields and nesting in the Subscription case class model to match zuora terminology a bit better


membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -123,6 +125,9 @@ case class RatePlanCharge(
     endDateCondition: EndDateCondition,
     upToPeriods: Option[Int],
     upToPeriodsType: Option[UpToPeriodsType],
+    chargedThroughDate: Option[LocalDate],
+    effectiveStartDate: LocalDate,
+    effectiveEndDate: LocalDate,
```
johnduffell, line 130: theses dates were on the RatePlan (see old lines 171-174) but in zuora they are actually on the charges.  We were just taking the max chargedThroughDate and the min of the other two across all charges

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -168,12 +173,16 @@ case class RatePlan(
...
-    start: LocalDate,
-    end: LocalDate,
 ) {
 
+  def chargedThroughDate: Option[LocalDate] =
+    ratePlanCharges.map(_.chargedThroughDate).list.toList.max // None is less than Some
+  def effectiveStartDate: LocalDate =
+    ratePlanCharges.map(_.effectiveStartDate).list.toList.min
+  def effectiveEndDate: LocalDate =
+    ratePlanCharges.map(_.effectiveEndDate).list.toList.min
```
johnduffell, line 184: these helper methods replicate the old logic that used to be in the deserialiser, (not to mention it was pretty sensible, it just didn't match the zuora data)

membership-common/src/main/scala/com/gu/memsub/subsv2/Subscription.scala
```diff
@@ -13,22 +13,18 @@ import scalaz.{NonEmptyList, Validation, \/}
...
+    contractEffectiveDate: LocalDate,
+    customerAcceptanceDate: LocalDate,
     termEndDate: LocalDate,
-    casActivationDate: Option[DateTime],
-    promoCode: Option[PromoCode],
     isCancelled: Boolean,
     ratePlans: List[RatePlan],
     readerType: ReaderType,
-    gifteeIdentityId: Option[String],
     autoRenew: Boolean,
```
johnduffell, line 24: these are renamed to pretty much match the zuora model.  I found that the removed fields were not actually used so I removed them for clarity.  There's no reason not to add them back later if we want them.

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/SubJsonReads.scala
```diff
@@ -61,38 +60,21 @@ object SubJsonReads {
       (__ \ "specificBillingPeriod").readNullable[Int] and
       (__ \ "endDateCondition").read[EndDateCondition] and
       (__ \ "upToPeriods").readNullable[Int] and
-      (__ \ "upToPeriodsType").readNullable[UpToPeriodsType]
-  )(RatePlanCharge.apply _)
+      (__ \ "upToPeriodsType").readNullable[UpToPeriodsType] and
+      (__ \ "chargedThroughDate").readNullable[LocalDate] and
+      (__ \ "effectiveStartDate").read[LocalDate] and
+      (__ \ "effectiveEndDate").read[LocalDate]
```
johnduffell, line 66: these fields we read from each charge as we go rather than separately grabbing them on the old line 71-82

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/SubJsonReads.scala
```diff
@@ -114,14 +96,10 @@ object SubJsonReads {
...
               (__ \ "contractEffectiveDate").read[LocalDate] and
               (__ \ "customerAcceptanceDate").read[LocalDate] and
-              (__ \ "termStartDate").read[LocalDate] and
               (__ \ "termEndDate").read[LocalDate] and
-              (__ \ "ActivationDate__c").readNullable[DateTime](lenientDateTimeReader) and
-              (__ \ "PromotionCode__c").readNullable[String].map(_.map(PromoCode)) and
               (__ \ "status").read[String].map(_ == "Cancelled") and
               (__ \ "ratePlans").read[List[RatePlan]](niceListReads(ratePlanReads)) and
               (__ \ "ReaderType__c").readNullable[String].map(ReaderType.apply) and
-              (__ \ "GifteeIdentityId__c").readNullable[String] and
```
johnduffell, line 124: a few unused fields removed

rupertbates (approved):
Makes sense to me 👍

membership-attribute-service/app/services/GuardianPatronService.scala
```diff
@@ -73,14 +73,11 @@ class GuardianPatronService(
       deliveryAddress = None,
       subscription = Subscription(
         id = Id(subscription.id),
-        name = Name(subscription.id),
+        subscriptionNumber = Name(subscription.id),
```
johnduffell, line 76: should rename the value class to match as well
```suggestion
        subscriptionNumber = SubscriptionNumber(subscription.id),
```

johnduffell: done this in the second commit on this PR

PR #1104, opened 2024-12-06
Remove extended Feast free trial period for existing subscribers

rupertbates:
### Why do we need this? 
When we initially launched the Feast app, we gave an extended free trial period to users who have a subscription to some of our other products. 

This is now being reduced to the 14 day free trial offered to all new users, because we could only offer an extended trial as a one-off 'gift' to this cohort (from a legal and tax perspective), rather than an ongoing benefit.

This PR removes the extended free trial from the /user-attributes/me response and tidies up.


paulbrown1982 (approved):
👍 LGTM

PR #1109, opened 2025-01-16
log catalog fetch failures together with identity id

johnduffell:
Zuora sandbox is throwing 500 errors, however strangely if the catalog fetch fails, it seemed to populate the catalog with an empty map.  This meant that when trying to read a subscription it can't look up the product rate plan id, which is confusing, as the real error is buried somewhere way up the log file (when the first request happened)

This PR makes it so that the error is thrown from the catalog fetch, and further changes to make it log it against the identity id of the current request, so if you're filtering the logs you will still see it.

After, you can see that searching for this test user's identity id (200275381) would pick up the error clearly 
```
2025-01-16 15:06:05,494 AccountController.scala:198 [INFO]: 200275381: Attempting to retrieve payment details for identity user: 200275381
2025-01-16 15:06:05,494 TouchpointComponents.scala:143 [ERROR]: 200275381: Failed to load the product catalog from Zuora due to: java.lang.RuntimeException: List((,List(JsonValidationError(List(No product array found, got JsUndefined('products' is undefined on object. Available keys are 'success', 'processId', 'reasons', 'requestId')),List()))))
2025-01-16 15:06:05,495 HttpErrorHandler.scala:305 [ERROR]: 

! @8870e91k4 - Internal server error, for (GET) [/user-attributes/me/mma] ->
 
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[RuntimeException: List((,List(JsonValidationError(List(No product array found, got JsUndefined('products' is undefined on object. Available keys are 'success', 'processId', 'reasons', 'requestId')),List()))))]]
	at play.api.http.HttpErrorHandlerExceptions$.$anonfun$convertToPlayException$2(HttpErrorHandler.scala:400)
	at scala.Option.map(Option.scala:242)
	at play.api.http.HttpErrorHandlerExceptions$.convertToPlayException(HttpErrorHandler.scala:398)

....
```
Note that the whole stack trace (reported by play framework ErrorHandler) would not be visible without viewing the original log.  I could get around that by wrapping the original exception in another one that contains the ID, but I think what I've done is less invasive and will be good enough.

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -19,6 +19,7 @@ import configuration.Stage
 import monitoring.CreateMetrics
 import org.apache.pekko.actor.ActorSystem
 import org.http4s.Uri
+import scalaz.{-\/, \/-}
 import scalaz.std.scalaFuture._
 import services._
 import services.salesforce.{ContactRepository, CreateScalaforce, SimpleContactRepository}
@@ -121,18 +122,27 @@ class TouchpointComponents(
     zuoraRestServiceOverride.getOrElse(simpleClientZuoraRestService)
   }
 
-  lazy val catalogRestClient = rest.SimpleClient[Future](backendConfig.zuoraRest, RequestRunners.configurableFutureRunner(60.seconds))
-  lazy val futureCatalog: Future[Catalog] = catalogServiceOverride.getOrElse(
-    CatalogService
-      .read(FetchCatalog.fromZuoraApi(catalogRestClient)(implicitly, LogPrefix.noLogPrefix), productIds)
+  private lazy val futureCatalogNoPrefix: Future[Catalog] = {
+    logger.infoNoPrefix(s"Loading catalog in $stage")
+    val catalogRestClient = rest.SimpleClient[Future](backendConfig.zuoraRest, RequestRunners.configurableFutureRunner(60.seconds))
+    catalogServiceOverride.getOrElse(
+      CatalogService
+        .read(FetchCatalog.fromZuoraApi(catalogRestClient)(implicitly, LogPrefix.noLogPrefix), productIds)
+        .flatMap {
+          case -\/(error) => Future.failed(new RuntimeException(error))
+          case \/-(result) => Future.successful(result)
+        },
+    )
+  }
+  def futureCatalog(implicit logPrefix: LogPrefix): Future[Catalog] =
+    futureCatalogNoPrefix
       .recover { case error =>
-        logger.errorNoPrefix(scrub"Failed to load the product catalog from Zuora due to: $error")
+        logger.error(scrub"Failed to load the product catalog from Zuora due to: $error")
         throw error
-      },
-  )
+      }
 
   lazy val subscriptionService: SubscriptionService[Future] = {
-    lazy val zuoraSubscriptionService = new SubscriptionService(futureCatalog, zuoraRestClient, zuoraSoapService)
+    lazy val zuoraSubscriptionService = new SubscriptionService(futureCatalog(_), zuoraRestClient, zuoraSoapService)
 
     subscriptionServiceOverride.getOrElse(zuoraSubscriptionService)
   }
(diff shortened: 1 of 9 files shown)


membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -121,18 +122,27 @@ class TouchpointComponents(
...
+  private lazy val futureCatalogNoPrefix: Future[Catalog] = {
+    logger.infoNoPrefix(s"Loading catalog in $stage")
+    val catalogRestClient = rest.SimpleClient[Future](backendConfig.zuoraRest, RequestRunners.configurableFutureRunner(60.seconds))
+    catalogServiceOverride.getOrElse(
+      CatalogService
+        .read(FetchCatalog.fromZuoraApi(catalogRestClient)(implicitly, LogPrefix.noLogPrefix), productIds)
+        .flatMap {
+          case -\/(error) => Future.failed(new RuntimeException(error))
+          case \/-(result) => Future.successful(result)
+        },
```
johnduffell, line 134: this pushes "left"s into a failed Future, as we don't really need to keep it as a `Future[Either....]]` at this point

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -121,18 +122,27 @@ class TouchpointComponents(
...
+    )
+  }
+  def futureCatalog(implicit logPrefix: LogPrefix): Future[Catalog] =
+    futureCatalogNoPrefix
       .recover { case error =>
-        logger.errorNoPrefix(scrub"Failed to load the product catalog from Zuora due to: $error")
+        logger.error(scrub"Failed to load the product catalog from Zuora due to: $error")
         throw error
```
johnduffell, line 141: I wasn't sure if this line was supposed to prevent the lazy val from initialising if it fails (and thus retry) but that doesn't seem to be possible.  The catalog fetch is attempted once, and after that if it fails, the server is basically useless.
If it's in PROD there is a healthcheck but it doesn't look like it would fail so it wouldn't be replaced,(the health check intentionally doesn't include test users)

Maybe one for another PR to add catalog fetch to the health check (and/or make it retry)

membership-common/src/main/scala/com/gu/memsub/subsv2/services/CatalogService.scala
```diff
@@ -50,13 +50,9 @@ object CatalogService extends SafeLogging {
         plans <- Json.fromJson[List[ProductRatePlan]](catalog)(productsReads).asEither.toDisjunction.leftMap(_.toString)
       } yield plans.map(catalogZuoraPlan => catalogZuoraPlan.id -> catalogZuoraPlan).toMap
 
-      val catalogMap = failableCatalog
-        .leftMap[ProductRatePlanMap](error => {
-          logger.errorNoPrefix(scrub"error: $error"); Map.empty // not sure why we empty-on-failure
```
johnduffell, line 55: I wasn't sure when I added this comment and I'm even less sure now!

rupertbates (approved):
Looks good 👍

membership-common/src/main/scala/com/gu/memsub/subsv2/services/CatalogService.scala
```diff
@@ -50,13 +50,9 @@ object CatalogService extends SafeLogging {
...
-        })
-        .merge
+      failableCatalog.leftMap(error => logger.errorNoPrefix(scrub"Failed to load catalog: $error"))
```
tomrf1, line 53: since we're no longer using the results of this `leftMap`, is it worth changing it to `.swap.foreach`?
...though that's not particularly clear either!

johnduffell: yeah agree, I was looking for a "leftForEach" which doesn't exist.  Another option was a pattern match but it felt a bit verbose.

I am sure I could have made a more invasive refactoring given more time, as we sholdn't need `Future[Either[...]]` as future incorporates failure.

johnduffell:
checked the logs in PROD and someone managed to get their MMA attributes successfully on a new instance

image

PR #1110, opened 2025-01-28
Give Feast access to supporter members

rupertbates:
### Why do we need this? 
This PR gives access to Feast for supporter members

### DIFF
--- a/membership-attribute-service/app/models/FeastApp.scala
+++ b/membership-attribute-service/app/models/FeastApp.scala
@@ -5,8 +5,6 @@ import org.joda.time.LocalDate
 
 object FeastApp {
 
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/models/FeastApp.scala
```diff
@@ -5,8 +5,6 @@ import org.joda.time.LocalDate
 
 object FeastApp {
 
-  val FeastFullLaunchDate = LocalDate.parse("2024-07-10")
```
rupertbates, line 8: This was unused

andrewHEguardian (approved):
Nice 👍

tjmw (approved):
👏  Great that this is such a simple change now!

PR #1111, opened 2025-01-30
Update variables and references from Guardian Light to Guardian Ad-Lite

david-pepper:
Update code in response to a product name change


membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
```diff
@@ -118,7 +118,7 @@ object SupporterRatePlanToAttributesMapper {
     ) -> tierThreeTransformer,
     List(
       "8a12831492c341730192dd1c39207038",
```
tjmw, line 120: This rate plan ID has now changed (a new one was created when we setup the new product). That's expected right @rupertbates?

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
```diff
@@ -250,7 +250,7 @@ object SupporterRatePlanToAttributesMapper {
     ) -> tierThreeTransformer,
     List(
       "71a1c43a1e192b28f702b3b47113000a",
```
tjmw, line 252: There's a new rate plan ID in code too.

membership-attribute-service/test/controllers/AttributeControllerTest.scala
```diff
@@ -96,7 +96,7 @@ class AttributeControllerTest extends Specification with AfterAll with Idiomatic
   private val newspaperPlusCookie = Cookie("newspaperPlus", "true")
   private val guardianWeeklyCookie = Cookie("guardianWeekly", "true")
   private val tierThreeCookie = Cookie("tierThree", "true")
-  private val guardianLightCookie = Cookie("guardianLight", "true")
+  private val guardianAdLiteCookie = Cookie("guardianAdLite", "true")
```
tjmw, line 99: Hmm I'm not sure what this is used for here, but we don't drop a cookie specifically related to Ad-Lite any more, rather a cookie is dropped representing the benefit (`gu_allow_reject_all`).

membership-common/conf/touchpoint.PROD.conf
```diff
@@ -67,7 +67,7 @@ touchpoint.backend.environments {
                     digipack="2c92a0fb4edd70c8014edeaa4ddb21e7"
                     supporterPlus="8a12865b8219d9b4018221061563643f"
                     tierThree="8a1295998ff2ec180190024b287b64c7"
-                    guardianLight="8a128dfb92c341720192dd19626429dc"
+                    guardianAdLite="8a128dfb92c341720192dd19626429dc"
```
tjmw, line 70: I guess this is a product ID here? It probably needs updating since Ad-Lite is a new product in Zuora (not a rename of a the existing Light product).

graham228221: @david-pepper the new Product Id is `8a1285e294443da501944b04cb382c9a`

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
graham228221: @david-pepper the new Rate Plan ID: `8a1285e294443da501944b04cb692c9e`

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
graham228221: @david-pepper new Rate Plan ID for CODE: `71a1bebf6be9444afad446c5ebaf0019`

membership-common/conf/touchpoint.CODE.conf
```diff
@@ -82,7 +82,7 @@ touchpoint.backend.environments {
                     digipack="2c92c0f84b786da2014b91d3629b4298"
                     supporterPlus="8ad09fc281de1ce70181de3b23b2363d"
                     tierThree="8ad097b48ff26452019001c67ad32035"
-                    guardianLight="8ad09658929f4cc80192afc6ec08389e"
```
graham228221, line 85: @david-pepper new Product ID for CODE: 8ad0869c9444afc7019446c5eaf33503

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
tjmw: Thanks @graham228221, sorry I should have looked up and added these myself! (Using the [product catalog](https://product-catalog.guardianapis.com/product-catalog.json) is a good way to check).

graham228221: Ah yes I didn't think of that, good tip!

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
david-pepper: Done

membership-common/conf/touchpoint.PROD.conf
david-pepper: Done

membership-common/conf/touchpoint.CODE.conf
david-pepper: Done

membership-attribute-service/test/controllers/AttributeControllerTest.scala
david-pepper: Ok, I've updated the tests file

membership-common/src/test/scala/com/gu/memsub/subsv2/services/TestCatalog.scala
```diff
@@ -27,7 +27,7 @@ object TestCatalog {
   val contributorPrpId = ProductRatePlanId("asdfasdf")
   val supporterPlusPrpId = ProductRatePlanId("8ad08cbd8586721c01858804e3275376")
   val tierThreePrpId = ProductRatePlanId("8ad097b48ff26452019001cebac92376")
-  val guardianLightPrpId = ProductRatePlanId("8a12831492c341730192dd1c39207038")
+  val guardianAdLitePrpId = ProductRatePlanId("8a12831492c341730192dd1c39207038")
```
rupertbates, line 30: I think this ID needs to change as well doesn't it?

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
```diff
@@ -177,8 +177,8 @@ object ProductRatePlanChargeProductType {
     override val id = "Tier Three"
   }
 
-  object GuardianLight extends ProductRatePlanChargeProductType {
-    override val id = "Guardian Light"
+  object GuardianAdLite extends ProductRatePlanChargeProductType {
```
johnduffell, line 180: To follow precedent I think `AdLite` would be enough, similar to `Weekly`.  AFAIK the only reason we had to add `GuardianPatron` is because we already have `Patron` for the membership version

membership-attribute-service/app/services/AccountDetailsFromZuora.scala
```diff
@@ -224,7 +224,7 @@ class AccountDetailsFromZuora(
       case Product.Digipack => requestedProductType == "Digipack" || requestedProductTypeIsContentSubscription
       case Product.SupporterPlus => requestedProductType == "SupporterPlus" || requestedProductTypeIsContentSubscription
       case Product.TierThree => requestedProductType == "TierThree" || requestedProductTypeIsContentSubscription
-      case Product.GuardianLight => requestedProductType == "GuardianLight" || requestedProductTypeIsContentSubscription
+      case Product.GuardianAdLite => requestedProductType == "GuardianAdLite" || requestedProductTypeIsContentSubscription
```
johnduffell, line 227: according to ProductFamily.scala ad lite is not a content subscription
https://github.com/guardian/members-data-api/pull/1111/files#diff-8434c9ffdcbb8fc26673d1f84012690fee7a77abf6ab033c73335a7b03a32f1aR36
I'm not 100% sure the significance of this check, but we should be consistent.

johnduffell: seems like it's used in MMA for things like whether to display a delivery address change
https://github.com/guardian/manage-frontend/blob/a8c65fe9d719b5bef22a6931f3e9477ebeb54a73/client/components/mma/delivery/records/DeliveryAddressStep.tsx#L612
So I would say this shouldn't be returned for ContentSubscription
```suggestion
      case Product.GuardianAdLite => requestedProductType == "GuardianAdLite"
```

membership-attribute-service/app/services/SupporterRatePlanToAttributesMapper.scala
AndreaDiotallevi: Done

membership-common/src/test/scala/com/gu/memsub/subsv2/services/TestCatalog.scala
AndreaDiotallevi: [Done](https://github.com/guardian/members-data-api/pull/1111/commits/18a04b865d2e29bf91991ac2913bf8e39533ad69)

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
AndreaDiotallevi: The Portfolio team have decided to not make these change as it would require lots of downstream systems to be updated in the process.

johnduffell: fair enough if that's decided, but it's just a scala case class in one project isn't it?  Unless that would require renaming the API name and in zuora etc as well which I agree might be a bit much by this time.

AndreaDiotallevi: It's a valid comment. But based on the fact that we will not change it everywhere, I don't see the point in updating just the Scala case class when the string still refers the old name?

johnduffell: to be honest it was really just to declutter and make it consistent with Weekly above but I can also see the value of using the full name everywhere so if you're happy then I'm happy!
https://github.com/guardian/members-data-api/blob/38b056bd7df882630bbc4140386e7701b686ff9f/membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala#L172-L174

rupertbates (approved):
Looks good to me

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
AndreaDiotallevi: I think it's a good compromise to update just the case class.

It is updated here: https://github.com/guardian/members-data-api/pull/1111/commits/21093cd1d6ac456c97d9bc8700902a3d6c5bf3d5

PR #1112, opened 2025-02-13
don't apply grace period to cancelled subscriptions

johnduffell:
There is a but that will affect digi sub and ad lite trials, plus possibly other deferred start subs that are cancelled.

Basically if you're in a free trial/before paid period, the customer acceptance date indicates the first paid day, but the contract effective date represents the date you agreed to the subscription.

In MDAPI you only get benefits based on the active/paid plans.  So to handle free trials we fast forward to the first day of the paid period to decide what benefits you will get.

However, for cancelled subscriptions, the paid trial is reduced to start and end at midnight at the point of first payment.
As a final issue, we actually give people an extra 1 day grace period after a rate plan ends.  This is to allow zuora’s auto renewal process to kick in which happens soon after the previous one goes out of term.

This means that when checking the subscription, in the trial period of a cancelled subscription, it fast forwards to the paid period (of zero length) adds one day grace period, and then returns the subscription as an active one.

This PR changes the code to notice a subscription in cancelled state, and doesn’t apply the 1 day grace period.

### DIFF
--- a/membership-common/src/main/scala/com/gu/memsub/subsv2/Subscription.scala
+++ b/membership-common/src/main/scala/com/gu/memsub/subsv2/Subscription.scala
@@ -76,7 +76,10 @@ object GetCurrentPlans {
       else if (product == Product.Digipack && plan.billingPeriod.toOption.contains(OneTimeChargeBillingPeriod))
         digipackGiftEnded(_ => sub.termEndDate >= dateToCheck)
       else
-        paidPlanEnded(_ => plan.effectiveEndDate >= dateToCheck)
+        paidPlanEnded(_ => {
+          val inGracePeriodAndNotCancelled = plan.effectiveEndDate == dateToCheck && !sub.isCancelled
+          plan.effectiveEndDate > dateToCheck || inGracePeriodAndNotCancelled
+        })
     }
 
     Sequence(
--- a/membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
+++ b/membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
@@ -8,6 +8,8 @@ import com.gu.memsub
 import com.gu.memsub.Subscription.{Id => _, _}
 import com.gu.memsub.subsv2._
 import com.gu.memsub.subsv2.reads.PlanWithCreditsTestData
+import com.gu.memsub.subsv2.reads.SubJsonReads.subscriptionReads
+import com.gu.memsub.subsv2.services.SubscriptionServiceTest.adLiteCancelledInTrial
 import com.gu.memsub.{Subscription => _, _}
 import com.gu.monitoring.SafeLogger
 import com.gu.okhttp.RequestRunners.HttpClient
@@ -18,6 +20,7 @@ import io.lemonlabs.uri.typesafe.dsl._
 import okhttp3._
 import org.joda.time.LocalDate
 import org.specs2.mutable.Specification
+import play.api.libs.json.Json
 import scalaz.Id._
 import scalaz.{-\/, NonEmptyList, \/, \/-}
 import utils.Resource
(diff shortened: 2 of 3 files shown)


membership-common/src/test/scala/com/gu/memsub/subsv2/services/TestCatalog.scala
```diff
@@ -27,7 +27,7 @@ object TestCatalog {
   val contributorPrpId = ProductRatePlanId("asdfasdf")
   val supporterPlusPrpId = ProductRatePlanId("8ad08cbd8586721c01858804e3275376")
   val tierThreePrpId = ProductRatePlanId("8ad097b48ff26452019001cebac92376")
-  val guardianAdLitePrpId = ProductRatePlanId("8a1285e294443da501944b04cb692c9e")
+  val guardianAdLitePrpId = ProductRatePlanId("71a1bebf6be9444afad446c5ebaf0019")
```
johnduffell, line 30: this was input as the PROD id, but actually it's only used in tests

membership-common/src/main/scala/com/gu/memsub/subsv2/Subscription.scala
```diff
@@ -76,7 +76,10 @@ object GetCurrentPlans {
       else if (product == Product.Digipack && plan.billingPeriod.toOption.contains(OneTimeChargeBillingPeriod))
         digipackGiftEnded(_ => sub.termEndDate >= dateToCheck)
       else
-        paidPlanEnded(_ => plan.effectiveEndDate >= dateToCheck)
+        paidPlanEnded(_ => {
+          val inGracePeriodAndNotCancelled = plan.effectiveEndDate == dateToCheck && !sub.isCancelled
+          plan.effectiveEndDate > dateToCheck || inGracePeriodAndNotCancelled
+        })
```
johnduffell, line 82: this was the minimal change, but I have previously had trouble understanding this function, and I'm sure it could be written as a for comprehension or similar.  If I'd had more time I'd have tried to make the whole thing easier to understand.

membership-common/src/test/scala/com/gu/memsub/subsv2/services/SubscriptionServiceTest.scala
```diff
@@ -262,6 +265,13 @@ class SubscriptionServiceTest extends Specification {
       GetCurrentPlans.currentPlans(digipackSub, referenceDate, catalog).map(_.head.id.get) mustEqual \/-("idDigipack")
     }
 
+    "if you cancel an Ad-Lite in the trial period, tell you you're not a subscriber" in {
+      val now = new LocalDate(2025, 2, 13) // still in trial
+      val adLite = subscriptionReads.reads(Json.parse(adLiteCancelledInTrial)).get
```
johnduffell, line 270: the rest of the tests are using minimal data structures defined in scala, but the fashion now is to use raw data pasted from zuora responses, so I've done that instead.

membership-common/src/test/scala/com/gu/memsub/subsv2/services/TestCatalog.scala
```diff
@@ -131,6 +131,13 @@ object TestCatalog {
       Map(contributorChargeId -> Contributor),
       Some(ProductType("type")),
     ),
+    guardianAdLitePrpId -> ProductRatePlan(
+      guardianAdLitePrpId,
+      "Guardian Ad Lite",
+      idForProduct(Product.AdLite),
+      Map(guardianAdLiteChargeId -> GuardianAdLite),
+      Some(ProductType("typeAdLite")),
```
johnduffell, line 139: if I remember right this is mainly needed for the specific things e.g home delivery days

membership-common/src/main/scala/com/gu/memsub/subsv2/Subscription.scala
```diff
@@ -76,7 +76,10 @@ object GetCurrentPlans {
       else if (product == Product.Digipack && plan.billingPeriod.toOption.contains(OneTimeChargeBillingPeriod))
         digipackGiftEnded(_ => sub.termEndDate >= dateToCheck)
       else
-        paidPlanEnded(_ => plan.effectiveEndDate >= dateToCheck)
+        paidPlanEnded(_ => {
+          val inGracePeriodAndNotCancelled = plan.effectiveEndDate == dateToCheck && !sub.isCancelled
```
johnduffell, line 80: if it's only just run out at midnight, and the subscription is not cancelled, we assume it will auto renew later in the day (1 day grace period)

ecguardian (approved):
LGTM 👍

PR #1115, opened 2025-02-19
Delete snyk workflow as not in use anymore

MaelGNM:
### What does this change

We no longer use Snyk, so delete the GH workflow


rBangay (approved):
👍

PR #1116, opened 2025-02-24
class subs as recently cancelled from midnight on the end date

rBangay:
Follows on from https://github.com/guardian/members-data-api/pull/1112 which makes a similar fix for "active" subscriptions.

We noticed that when cancelling a product like a `Guardian ad-lite` with a free initial period, the product was not appearing in the `/cancelled` api endpoint response straight away, but waiting one day.  This is because we fast forward to the first payment date when you're in the free period, and we allow a one day grace period to allow for auto renewal jobs to run.

This issue would also apply to other products, and also (less obviously) where the cancellation was outside the free period.

This PR changes it so that a subscription appears as "recently cancelled" from the start of the term end date.  Therefore it will appear in manage in the correct section.


membership-common/src/main/scala/com/gu/memsub/subsv2/Subscription.scala
```diff
@@ -41,10 +41,14 @@ with the newest plan for upgrade and cancel scenarios, so in this case the most
  */
 object GetCurrentPlans {
 
-  def bestCancelledPlan(sub: Subscription): Option[RatePlan] =
-    if (sub.isCancelled && sub.termEndDate.isBefore(LocalDate.now()))
+  def bestCancelledPlan(sub: Subscription): Option[RatePlan] = {
+    val today = LocalDate.now()
+    val isAlreadyFinished = sub.termEndDate.isEqual(today) || sub.termEndDate.isBefore(today)
```
johnduffell, line 46: the `isEqual` is the addition - (if you're on the termEndDate, your subscription is finished.)

michaelbjacobson (approved):
Nice!

PR #1117, opened 2025-02-25
improve contributions store logging

johnduffell:
At the moment we don't really log the responses from contributions store.

This PR adds the logging, so we know what's happened.

Also, it improves the security of failure cases by not logging internal errors.

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -230,13 +230,18 @@ class AttributeController(
   def oneOffContributions =
     AuthorizeForScopes(requiredScopes = List(readSelf)).async { implicit request =>
       metrics.measureDuration("GET /user-attributes/me/one-off-contributions") {
+        import request.logPrefix
         val userHasValidatedEmail = request.user.userEmailValidated.getOrElse(false)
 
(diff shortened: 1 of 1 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -230,13 +230,18 @@ class AttributeController(
   def oneOffContributions =
     AuthorizeForScopes(requiredScopes = List(readSelf)).async { implicit request =>
       metrics.measureDuration("GET /user-attributes/me/one-off-contributions") {
+        import request.logPrefix
         val userHasValidatedEmail = request.user.userEmailValidated.getOrElse(false)
 
         val futureResult: Future[Result] =
           if (userHasValidatedEmail) {
             contributionsStoreDatabaseService.getAllContributions(request.user.identityId).map {
-              case Left(err) => Ok(err)
```
johnduffell, line 238: I don't think we should have been returning the postgres error to the client, there doesn't seem to be any discussion on the original pr to say it's needed https://github.com/guardian/members-data-api/pull/388/files#r1969905410

tomrf1: good spot!

tomrf1 (approved):

PR #1118, opened 2025-03-11
build: bump io.netty version to remove high vulnerability CVE-2025-24970

AndreaDiotallevi:
## What does this change?

Bump io.netty package to remove high vulnerability CVE-2025-24970

Link: https://github.com/advisories/GHSA-4g8c-wm8x-jfhw

## Additional changes

This test was failing `read an echo legacy monthly subscription`, because the effective end date of charges in the test file  `EchoLegacy.json` was set on a real date that is now in the past (2025-03-07). So I have updated those dates to far in the future (this might need to be revisited, but I am keen to get this PR merged).


JoeMitchellGuardian (approved):

project/Dependencies.scala
```diff
@@ -20,8 +20,8 @@ object Dependencies {
   val awsCloudWatch = "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsClientVersion
   val scalaz = "org.scalaz" %% "scalaz-core" % "7.3.8"
   val anorm = "org.playframework.anorm" %% "anorm" % "2.7.0"
-  val netty = "io.netty" % "netty-codec" % "4.1.87.Final"
-  val nettyHttp = "io.netty" % "netty-codec-http" % "4.1.87.Final"
+  val netty = "io.netty" % "netty-codec" % "4.1.118.Final"
+  val nettyHttp = "io.netty" % "netty-codec-http" % "4.1.118.Final"
```
johnduffell, line 24: not worth revisiting now, but for future, if there are two packages that have to be kept in step, it can be extracted e.g.
```scala
val nettyVersion = "4.1.118.Final"
val netty = "io.netty" % "netty-codec" % nettyVersion
val nettyHttp = "io.netty" % "netty-codec-http" % nettyVersion
```

PR #1119, opened 2025-03-20
add new endpoint to quickly check if someone is a test user

johnduffell:
At the moment it is quite difficult to tell if someone is a test user in manage, without calling the main (slow) /user-attributes/me endpoint on MDAPI.  Although this is usually called, for deep links we don't always make that call (e.g. billing page)

This PR adds a new super lightweight call just to do the test user check without calling zuora etc.

Tested locally.

### DIFF
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -275,9 +275,11 @@ class AttributeController(
       maybeAttributes
   }
 
-  def isTestUser: Action[AnyContent] =
-    AuthorizeForScopes(List(readSelf)) {
+  def isTestUser: Action[AnyContent] = {
+    val scope = List(readSelf) // this doesn't end in .secure so we won't call through to okta
+    AuthorizeForScopes(scope) {
       NoContent
     }
+  }
 
 }
--- a/membership-attribute-service/app/controllers/AttributeController.scala
+++ b/membership-attribute-service/app/controllers/AttributeController.scala
@@ -274,4 +274,10 @@ class AttributeController(
     else
       maybeAttributes
   }
+
+  def isTestUser: Action[AnyContent] =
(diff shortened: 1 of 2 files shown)


membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -274,4 +274,10 @@ class AttributeController(
     else
       maybeAttributes
   }
+
+  def isTestUser: Action[AnyContent] =
+    AuthorizeForScopes(List(readSelf)) {
```
johnduffell, line 279: we weren't 100% sure whether we need readSelf or nothing

johnduffell: TODO check if it calls okta/identity and if so can we save doing that

membership-attribute-service/conf/routes
```diff
@@ -23,6 +23,7 @@ GET         /user-attributes/me/one-off-contributions
 
 PUT         /user-attributes/me/delivery-address/:contactId                     controllers.ContactController.updateDeliveryAddress(contactId: String)
 
+GET         /user-attributes/me/is-test-user                                    controllers.AttributeController.isTestUser
```
johnduffell, line 26: we decided to go for a narrow scope of name rather than making it generic, as the function is unlikely to change in the short term.  But since manage will be the only client for now, it wouldn't be a problem to replace it later.

membership-attribute-service/app/controllers/AttributeController.scala
```diff
@@ -274,4 +274,10 @@ class AttributeController(
...
+      NoContent
+    }
```
johnduffell, line 281: this will return something like this (due to the AddGuIdentityHeadersFilter that applies to all endpoints)
x-gu-identity-id: NNNNNN
x-gu-membership-test-user: false

rBangay (approved):
👍

PR #1122, opened 2025-03-25
Tortoise Media Stripe Account

andresilva-guardian:
### Why do we need this? 
Currently, the manage-frontend project uses Stripe Elements for updating payment methods. This involves users entering their payment details directly on the site, which are then processed using Stripe's API. The existing implementation includes various flows for updating payment methods, such as credit cards and direct debits, and handles different scenarios like payment failures and reCAPTCHA verification.

Our goal here is to remove GNM from the Observer-only payment flow (“Home delivery” and “Subscription Card”). By taking Observer-only funds on behalf of TM, and later settling them to TM, GNM is acting as a “Payment Facilitator”. To avoid this, we are implementing a specific flow for this product, which consists of using the “Stripe” payment solution in its self-hosted format (Stripe Checkout) with the usage of a service account belonging to Tortoise Media.

Because of this, we must add the Tortoise Media Stripe Public/Private API Keys mapping to make sure that the system works and accepts credit cards on this new account.

### The changes 

• Added support for a new Stripe configuration for Tortoise Media Membership:
  - Modified `TouchpointComponents.scala` to include a third Stripe configuration parameter in the `chooseStripe` method
  - Updated `ChooseStripe.scala` to handle the new Tortoise Media Stripe service configuration
  - Added a new method parameter `tortoiseMediaStripeServiceConfig` to `createFor` method

• Updated configuration files:
  - Added Tortoise Media Membership Stripe keys in both `touchpoint.CODE.conf` and `touchpoint.PROD.conf`
    - Added test public key in CODE environment
    - Added live public key in PROD environment

• Modified `TouchpointBackendConfig.scala`:
  - Added a new field `stripeTortoiseMediaMembership` to the case class
  - Updated the configuration loading method to include the new Tortoise Media Membership Stripe configuration
  - Used the UK country configuration for the new Tortoise Media Membership variant

• Implemented additional mapping for the new Stripe service:
  - Added a new Stripe service to `stripeServicesByPublicKey` map in `ChooseStripe` object

These changes were implemented to prepare the system to support a new Stripe membership type for Tortoise Media.

#### Recorded Demo
https://github.com/user-attachments/assets/c5a2e9f1-08f5-433b-ab0e-8991d948b72a

#### Sequence Diagram
Update Payment Method - Credit Card for Sunday Only The Observer Subscription

#### Linked Trello Cards
https://trello.com/c/7pqmbEKX/743-tortoise-stripe-api-key-on-members-data-api
https://trello.com/c/uAwOrl8b/725-estimate-migrating-mma-to-use-stripe-checkout
https://trello.com/c/GDkXwLUz/731-poc-stripe-checkout-implementation-on-mma
https://trello.com/c/H9gGqX0x/737-stripe-checkout-implementation-on-mma

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -174,7 +174,7 @@ class TouchpointComponents(
     )
 
   lazy val chooseStripe: ChooseStripe = chooseStripeOverride.getOrElse(
-    ChooseStripe.createFor(backendConfig.stripeUKMembership, backendConfig.stripeAUMembership),
+    ChooseStripe.createFor(backendConfig.stripeUKMembership, backendConfig.stripeAUMembership, backendConfig.stripeTortoiseMediaMembership),
   )
 
   lazy val paymentDetailsForSubscription: PaymentDetailsForSubscription = new PaymentDetailsForSubscription(paymentService, futureCatalog(_))
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -177,7 +177,7 @@ class TouchpointComponents(
     ChooseStripe.createFor(
       backendConfig.stripeUKMembership,
       backendConfig.stripeAUMembership,
-      backendConfig.stripeTortoiseMediaMembership,
+      backendConfig.stripeTortoiseMedia,
     ),
   )
 
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -174,7 +174,11 @@ class TouchpointComponents(
     )
 
   lazy val chooseStripe: ChooseStripe = chooseStripeOverride.getOrElse(
-    ChooseStripe.createFor(backendConfig.stripeUKMembership, backendConfig.stripeAUMembership, backendConfig.stripeTortoiseMediaMembership),
+    ChooseStripe.createFor(
+      backendConfig.stripeUKMembership,
+      backendConfig.stripeAUMembership,
+      backendConfig.stripeTortoiseMediaMembership,
+    ),
   )
 
   lazy val paymentDetailsForSubscription: PaymentDetailsForSubscription = new PaymentDetailsForSubscription(paymentService, futureCatalog(_))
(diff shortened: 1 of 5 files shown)


membership-common/conf/touchpoint.PROD.conf
```diff
@@ -27,6 +27,10 @@ touchpoint.backend.environments {
                 secret = ""
                 public = "pk_live_yc7a7N3P0Z8r63PkGow0tOOt"
             }
+            tortoise-media-membership.key {
+                secret = ""
```
johnduffell, line 31: Make sure to put the actual value in S3 in both the CODE and PROD directories before deploying.
Although in this case I think it would be ok (only this exact function would be broken), in some cases it is possible that the server would start but be broken if the config is not right.

johnduffell (approved):
fantastic description, I like the video/gif
The changes look fine, I did make a minor formatting suggestion but good work getting it all set up and prepared into a PR!

membership-common/conf/touchpoint.PROD.conf
andresilva-guardian: It's already done for DEV + CODE @johnduffell , I will do it for PROD right after deploying these changes to CODE. 🙏

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -174,7 +174,7 @@ class TouchpointComponents(
     )
 
   lazy val chooseStripe: ChooseStripe = chooseStripeOverride.getOrElse(
-    ChooseStripe.createFor(backendConfig.stripeUKMembership, backendConfig.stripeAUMembership),
+    ChooseStripe.createFor(backendConfig.stripeUKMembership, backendConfig.stripeAUMembership, backendConfig.stripeTortoiseMediaMembership),
```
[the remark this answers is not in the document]
andresilva-guardian: Changed @johnduffell , thanks 💪

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -174,7 +174,11 @@ class TouchpointComponents(
     )
 
   lazy val chooseStripe: ChooseStripe = chooseStripeOverride.getOrElse(
-    ChooseStripe.createFor(backendConfig.stripeUKMembership, backendConfig.stripeAUMembership),
+    ChooseStripe.createFor(
+      backendConfig.stripeUKMembership,
+      backendConfig.stripeAUMembership,
+      backendConfig.stripeTortoiseMediaMembership,
```
rupertbates, line 180: I probably wouldn't have the word Membership on the end of this. The reason it is on the others is purely historical and has nothing to do with Tortoise

andresilva-guardian: Changed @rupertbates thanks for the heads-up. 💪 
When I'm not aware of the conventions I just clone them to be safe.

PR #1123, opened 2025-04-08
use alternative tier for observer only delivery

johnduffell:
This PR updates the "tier" property for home delivery sunday only to be "tier": "Newspaper Delivery - Observer" and likewise for other observer only subs

This means that manage can correctly display the branding and options for these subscribers.

Sunday+ is NOT classed as observer only (as they are giving a portion of their money for digital access)

Changes are also being made in Manage and will be shipped in advance of this PR (to avoid breaking MMA for observer people)

Tested OK in DEV with various subs:
```
"tier": "Newspaper Delivery",
"tier": "Newspaper Delivery - Observer",
"tier": "Guardian Ad-Lite",
"tier": "Guardian Weekly - Domestic",
"tier": "Supporter Plus",
```

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -211,13 +211,11 @@ object AccountDetails {
   }
 
   def getTier(catalog: Catalog, plan: RatePlan): Json.JsValueWrapper =
-    if (
-      plan.product(catalog) == Product.Delivery &&
-      plan.name(catalog) == "Sunday"
-    )
-      "Newspaper Delivery - Observer"
-    else
-      plan.productName
+    plan.product(catalog) match {
+      case Product.Delivery if plan.name(catalog) == "Sunday" => "Newspaper Delivery - Observer"
+      case Product.DigitalVoucher if plan.name(catalog) == "Sunday" => "Newspaper Digital Voucher - Observer"
+      case _ => plan.productName
+    }
 
 }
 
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -43,8 +43,6 @@ object AccountDetails {
 
       val product = accountDetails.subscription.plan(catalog).product(catalog)
 
-      val mmaCategory = mmaCategoryFrom(product)
```
johnduffell, line 46: this is unused and was removed in a previous PR from manage, so is also removed here now.

membership-attribute-service/app/services/GuardianPatronService.scala
```diff
@@ -83,7 +83,7 @@ class GuardianPatronService(
           RatePlan(
             id = RatePlanId("guardian_patron_unused"), // only used for contribution amount change
             productRatePlanId = Catalog.guardianPatronProductRatePlanId,
-            productName = "Guardian Patron",
+            productName = "guardianpatron",
```
johnduffell, line 86: this was misaligned with what we expect to appear in the tier property so I have updated to match.  (Although I would probably have gone for the former, it looks like manage relies on the latter in many places already.)

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -212,17 +209,16 @@ object AccountDetails {
...
-    case Product.AdLite => "subscriptions"
-    case Product.GuardianPatron => "subscriptions"
-    case Product.Contribution => "recurringSupport"
-    case Product.Membership => "membership"
-    case _ => "subscriptions" // fallback - passing undefined value breaks manage
-  }
+
+  def getTier(catalog: Catalog, plan: RatePlan): Json.JsValueWrapper =
+    if (
+      plan.product(catalog) == Product.Delivery &&
```
johnduffell, line 215: only covers Home delivery at the moment, as we don't think voucher or ND has that option (although we can check the data and update in another PR)

PR #1124, opened 2025-04-09
Add Support for Tortoise Media's Dedicated GoCardless Payment Gateway

andresilva-guardian:
### Why do we need this? 
This change enables directing direct debit payments from Tortoise Media customers through a dedicated GoCardless payment gateway. By adding support for a specific Tortoise Media gateway, we can properly route and track these transactions separately from our standard GoCardless payments, allowing for proper financial reconciliation and reporting.

### The changes

- Added a new payment gateway option `GoCardlessTortoiseMediaGateway` with the name "GoCardless - Observer - Tortoise Media"
- Modified the direct debit form to include an optional `gatewayOwner` field
- Updated the payment method creation logic to select the appropriate gateway based on the `gatewayOwner` value:
-- If `gatewayOwner` equals "tortoise-media", use the new Tortoise Media gateway
-- Otherwise, use the standard GoCardless gateway
- Registered the new gateway in the payment gateway map for proper serialization/deserialization

[Trello Card](https://trello.com/c/8DeIOlkm/760-direct-debit-gocardless-on-tortoise-media-for-observer-only-subscriptions)

### DIFF
--- a/membership-attribute-service/app/controllers/PaymentUpdateController.scala
+++ b/membership-attribute-service/app/controllers/PaymentUpdateController.scala
@@ -23,6 +23,7 @@ import services.mail.Emails.paymentMethodChangedEmail
 import services.mail.{Card, DirectDebit, PaymentType, SendEmail}
 import utils.SimpleEitherT
 import utils.SimpleEitherT.SimpleEitherT
+import models.GatewayOwner
 
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -138,7 +139,10 @@ class PaymentUpdateController(
             "accountName" -> nonEmptyText,
             "accountNumber" -> nonEmptyText,
             "sortCode" -> nonEmptyText,
-            "gatewayOwner" -> optional(text),
+            "gatewayOwner" -> optional(text).transform[Option[GatewayOwner]](
+              _.flatMap(GatewayOwner.fromString),
+              _.map(_.toString)
+            ),
           )
         }
 
@@ -171,12 +175,10 @@ class PaymentUpdateController(
             lastName = billToContact.lastName,
             countryCode = "GB",
           )
-          paymentGatewayToUse =
-            if (paymentGatewayOwner.equals("tortoise-media")) {
-              GoCardlessTortoiseMediaGateway
-            } else {
-              GoCardlessGateway
-            }
+          paymentGatewayToUse = paymentGatewayOwner match {
+            case Some(GatewayOwner.TortoiseMedia) => GoCardlessTortoiseMediaGateway
+            case _ => GoCardlessGateway
+          }
           createPaymentMethod = CreatePaymentMethod(
             accountId = subscription.accountId,
             paymentMethod = bankTransferPaymentMethod,
--- a/membership-attribute-service/app/controllers/PaymentUpdateController.scala
+++ b/membership-attribute-service/app/controllers/PaymentUpdateController.scala
@@ -141,7 +141,7 @@ class PaymentUpdateController(
             "sortCode" -> nonEmptyText,
             "gatewayOwner" -> optional(text).transform[Option[GatewayOwner]](
               _.flatMap(GatewayOwner.fromString),
-              _.map(_.toString)
+              _.map(_.toString),
             ),
           )
         }
--- a/membership-attribute-service/app/controllers/PaymentUpdateController.scala
+++ b/membership-attribute-service/app/controllers/PaymentUpdateController.scala
@@ -171,7 +171,8 @@ class PaymentUpdateController(
             lastName = billToContact.lastName,
             countryCode = "GB",
           )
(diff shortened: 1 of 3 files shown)


membership-attribute-service/app/controllers/PaymentUpdateController.scala
```diff
@@ -169,10 +171,16 @@ class PaymentUpdateController(
             lastName = billToContact.lastName,
             countryCode = "GB",
           )
+          paymentGatewayToUse =
+            if (paymentGatewayOwner.equals("tortoise-media")) {
```
johnduffell, line 175: what type is `paymentGatewayOwner`?  I would expect it to be an `Option[String]` in which case we would want to do 
```suggestion
            if (paymentGatewayOwner.contains("tortoise-media")) {
```
I'm not sure if `.equals` is the same as `==` otherwise, but it might only check the type at run time rather than compile time.

membership-attribute-service/app/controllers/PaymentUpdateController.scala
```diff
@@ -169,10 +171,16 @@ class PaymentUpdateController(
...
+              GoCardlessTortoiseMediaGateway
+            } else {
+              GoCardlessGateway
+            }
```
johnduffell, line 179: I'd recommend to miss out the braces if it's only a simple if expression like that

(or you could go for a pattern match if you think it looks neater for dealing with the Option:
```suggestion
            paymentGatewayOwner match {
              case Some("tortoise-media") => GoCardlessTortoiseMediaGateway
              case _ => GoCardlessGateway
            }
```
)

membership-common/src/main/scala/com/gu/zuora/api/PaymentGateway.scala
```diff
@@ -26,6 +26,9 @@ case object StripeAUPaymentIntentsMembershipGateway extends PaymentGateway {
 case object GoCardlessGateway extends PaymentGateway {
   val gatewayName = "GoCardless"
 }
+case object GoCardlessTortoiseMediaGateway extends PaymentGateway {
+  val gatewayName = "GoCardless - Observer - Tortoise Media"
```
johnduffell, line 30: I assume this is the accurate name that zuora sandbox and PROD are both using?

johnduffell (commented):
looks good although I was not sure about the .equals line.  Have you been able to test it locally/in CODE env?

membership-attribute-service/app/controllers/PaymentUpdateController.scala
andresilva-guardian: @johnduffell Using some AI:

> These two conditions are not the same in Scala.
> 
> `paymentGatewayOwner.equals("tortoise-media")`
> This checks for exact string equality
> Returns true only if paymentGatewayOwner is precisely "tortoise-media"
> 
> 
> `paymentGatewayOwner.contains("tortoise-media")`
> This checks if the string contains "tortoise-media" as a substring
> Returns true if paymentGatewayOwner is "tortoise-media" but also if it's something like "the-tortoise-media", "tortoise-media-company", etc.
> 
> The contains method is more permissive since it will match partial occurrences, while equals requires an exact match.

andresilva-guardian: Should we trust AI on this one? Because I'm afraid of using other values in the future that may include "tortoise-media". Also, should we create some sort of enum here?

membership-attribute-service/app/controllers/PaymentUpdateController.scala
andresilva-guardian: Thanks @johnduffell , since `paymentGatewayOwner` is an Option[String], I think your suggestion is better 💪

membership-common/src/main/scala/com/gu/zuora/api/PaymentGateway.scala
andresilva-guardian: @johnduffell I was given to me as it is right here by @paulbrown1982 yesterday, so I believe so 👍

johnduffell:
as discussed, some suggestions are on this PR https://github.com/guardian/members-data-api/pull/1125

andresilva-guardian:
After testing on CODE env, it results in the following error:

> 2025-04-09 14:01:01,687 PaymentUpdateController.scala:207 [ERROR]: 200388244: Failed to update direct debit due to failed to create direct debit payment method. Exception : com.gu.zuora.soap.models.errors.ZuoraPartialError: TRANSACTION_FAILED: Transaction declined.validation_failed - email can't be blank

This error is expected. @paulbrown1982 informed us that this is a known issue related to the Configuration of the GoCardless Gateway for Tortoise Media. We are waiting for help from the Zuora helpdesk to know how to deal with this issue.

PR #1126, opened 2025-04-15
add switch to make observer only appear in CODE

johnduffell:
This PR sits on top of https://github.com/guardian/members-data-api/pull/1123

It adds a switch so we can ship it to main and only see the behaviour in CODE.

It seemed like a nice idea to avoid conflicts, but now I'm wondering with the extent of the changes, whether it's actually worse overall.  Hence the separate branch/PR so we can decide separately.


johnduffell:
closing as it will be easier to redo the first PR if there are conflicts

PR #1128, opened 2025-05-09
add tier property to currentPlans/futurePlans

johnduffell:
At the moment you can switch from supporter membership to contribution.

Because you pay in advance, the switch happens at the end of your term which could be almost a year ahead.

This means that after the switch, you are displayed in manage as still a supporter but your next payment will be calculated for your contribution.

This is not ideal because there's no clue that you have switched, and there is not enough information in the MDAPI response to make it possible to display that on screen.

This PR changes the MDAPI response so it also contains a tier property in each current and future plan.  The one at the top level still remains the "best" plan's tier.

TODO
- test in CODE if needed
- add a unit test for a pending switch sub (this would be a good idea in general, not something specific to this PR)

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -117,6 +117,7 @@ object AccountDetails {
       }
 
       def jsonifyPlan(plan: RatePlan) = Json.obj(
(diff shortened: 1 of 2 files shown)


andresilva-guardian (approved):
Thank you @johnduffell , it is working as expected 💪

PR #1129, opened 2025-05-15
change the fetcher test to use the real catalog

johnduffell:
I tried to use this manual sub fetcher and it failed because it was using a hard coded test catalog rather than the actual one.

This PR updates the test to use the actual catalog.


rBangay (approved):
👍

PR #1131, opened 2025-06-02
Fixed Tortoise Media Stripe Intents Gateway selection 

andresilva-guardian:
### Why do we need this? 

| Expected | Current |
| -------- | ------- |
| image | image |

After @graham228221 sent this [message](https://chat.google.com/room/AAAAuotUxTg/0znlKYuRcdg/0znlKYuRcdg?cls=10) we have conducted an investigation to determine the causes of the fact that the value for Payments Gateway was not being well defined after a credit card update on MMA.
It seems that we were missing support a the Stripe Payment Gateway variant for "Tortoise Media" in our Zuora integration. This allows us to route payment intents for Tortoise Media subscriptions through a dedicated Stripe gateway, ensuring correct payment processing and reporting for this product line.

### The changes

- Added a new `StripeTortoiseMediaPaymentIntentsMembershipGateway` object to the `PaymentGateway` list.
- Updated the `StripeServiceConfig` and `StripeService` classes to accept an optional `variant` parameter.
- Changed the logic in both the attribute service and common Stripe service to select the new Tortoise Media gateway when the variant is set to `"tortoise-media"`.
- Registered the new gateway in the list of available payment gateways.
- Updated imports and configuration wiring to support the new variant.

Trello Card: [Wrong Payment Gateway for Observer after Credit Card Update](https://trello.com/c/ELfZjX2j/823-wrong-payment-gateway-for-observer-after-credit-card-update)


membership-common/src/test/resources/rest/plans/Digi.json
```diff
@@ -12,14 +12,14 @@
...
-  "termStartDate": "2024-05-20",
-  "termEndDate": "2025-05-20",
+  "contractEffectiveDate": "2025-05-20",
+  "serviceActivationDate": "2025-05-20",
+  "customerAcceptanceDate": "2025-06-05",
+  "subscriptionStartDate": "2025-05-20",
+  "subscriptionEndDate": "2026-05-20",
+  "lastBookingDate": "2025-05-20",
+  "termStartDate": "2025-05-20",
+  "termEndDate": "2026-05-20",
```
johnduffell, line 22: why were these changed? was it a test failure due to something using today's date in the tests?

If so I think a load of tests might have been added a year ago and they need to be parameterised to avoid them just constantly adding noise
See this PR where it already happened and I hard coded the date https://github.com/guardian/members-data-api/pull/1130

membership-attribute-service/app/services/stripe/StripeService.scala
```diff
@@ -6,9 +6,17 @@ import com.gu.stripe.StripeServiceConfig
...
-  val paymentIntentsGateway: PaymentGateway = RegionalStripeGateways.getPaymentIntentsGatewayForCountry(apiConfig.stripeAccountCountry)
+class StripeService(
+    apiConfig: StripeServiceConfig,
+    val basicStripeService: BasicStripeService,
+) {
+  val paymentIntentsGateway: PaymentGateway =
+    apiConfig.variant match {
+      case Some("tortoise-media") => StripeTortoiseMediaPaymentIntentsMembershipGateway
+      case _ => RegionalStripeGateways.getPaymentIntentsGatewayForCountry(apiConfig.stripeAccountCountry)
+    }
```
johnduffell, line 19: does seem a bit odd that we are first doing it off the "variant" (which includes country) and if that fails we use the country itself.
I think for consistency we should just change it to use variant for everything and get rid of the `getPaymentIntentsGatewayForCountry` function, then it's all clear and in one place:
```suggestion
    apiConfig.variant match {
      case "tortoise-media" => StripeTortoiseMediaPaymentIntentsMembershipGateway
      case "au-membership" => StripeAUPaymentIntentsMembershipGateway
      case _ => StripeUKPaymentIntentsMembershipGateway
    }
```
Since we're pattern matching it, it might also make sense to turn variant [into an ADT ](https://www.baeldung.com/scala/algebraic-data-types#sum-types) but I don't think there's a strong argument yet as it's mostly for config.

membership-common/src/test/resources/rest/plans/Digi.json
andresilva-guardian: Tests were not passing because end date was June 1 2025. Since I needed to test these changes asap on CODE, I just fixed the dates because a bigger change here would be out of scope/context.

membership-attribute-service/app/services/stripe/StripeService.scala
johnduffell: true but if you look in that function then it's just an if statement so we're just flattening it up a level.
Also I know that for recurring we have one AU account, one ROW account, and one tortioise account so it makes sense to represent that in the code.

membership-common/src/test/resources/rest/plans/Digi.json
johnduffell: got it OK. So once we merge this PR: https://github.com/guardian/members-data-api/pull/1133
then you can merge main in to yours and these changes can be reverted

membership-attribute-service/app/services/stripe/StripeService.scala
```diff
@@ -6,9 +6,20 @@ import com.gu.stripe.StripeServiceConfig
...
+class StripeService(
+    apiConfig: StripeServiceConfig,
+    val basicStripeService: BasicStripeService,
+) {
+  val paymentIntentsGateway: PaymentGateway =
+    apiConfig.variant match {
+      case "tortoise-media" => StripeTortoiseMediaPaymentIntentsMembershipGateway
+      case "au-membership" => StripeAUPaymentIntentsMembershipGateway
+      case _ => StripeUKPaymentIntentsMembershipGateway
+    }
```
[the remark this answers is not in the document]
andresilva-guardian: I don't believe so but it seemed a bit odd having 2 different logics around. Also I was trying to just fix a bug and not making a refactor, so I cloned the logic. If you think that is possible we can try so unify this call somewhere in between / shared.

membership-common/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -58,7 +63,12 @@ class StripeService(apiConfig: StripeServiceConfig, client: FutureHttpClient)(im
   val httpClient: FutureHttpClient = client
   val publicKey: String = apiConfig.credentials.publicKey
   val paymentGateway: PaymentGateway = RegionalStripeGateways.getGatewayForCountry(apiConfig.stripeAccountCountry)
-  val paymentIntentsGateway: PaymentGateway = RegionalStripeGateways.getPaymentIntentsGatewayForCountry(apiConfig.stripeAccountCountry)
+  val paymentIntentsGateway: PaymentGateway =
```
johnduffell, line 66: this whole StripeService class can be deleted as it's unused. I checked back ( https://github.com/guardian/members-data-api/pull/1069 ) to see if any tests were missed and there don't seem to have been any, so we can get rid of this one without losing any testing

johnduffell (approved):
great thanks for that, just a small one to delete the unused StripeService before merging 👍

membership-common/src/main/scala/com/gu/stripe/StripeService.scala
johnduffell: https://github.com/guardian/members-data-api/pull/1134

PR #1133, opened 2025-06-03
fix tests that depended on the date to hard code it

johnduffell:
supersedes https://github.com/guardian/members-data-api/pull/1132 but doesn't pass `today` around as much

similar to https://github.com/guardian/members-data-api/pull/1130

This fixes a bunch more tests that were failing due to depending on dates and were added roughly a year ago


andresilva-guardian (approved):
Thanks @johnduffell 💪

PR #1134, opened 2025-06-03
Removed StripeService unused class

andresilva-guardian:
This pull request removes the unused duplicated StripeService class.


johnduffell (approved):
nice, thanks for following up with that.

Build issue looks like a scalafmt thing.

Also unless I missed something please can you just put a one liner like "remove unused duplicate StripeService class" as the description as I don't think it needs such an explanation

PR #1136, opened 2025-06-17
Fixing JSON validation issue found by Copilot

paulbrown1982:
### Why do we need this? 
Fixing JSON validation for when `{ ... "workEmail": "" ... }` is returned as an empty string by Zuora, now it will properly result in `None()` rather than `Some("")`

This has no real downstream effect other than possibly wasting a line of whitespace in the UI.


membership-common/src/test/resources/rest/plans/Credits.json
```diff
@@ -16,10 +16,10 @@
   "serviceActivationDate": "2024-06-13",
   "customerAcceptanceDate": "2024-06-28",
   "subscriptionStartDate": "2024-06-13",
-  "subscriptionEndDate": "2025-06-13",
+  "subscriptionEndDate": "2025-07-13",
```
rupertbates, line 19: If this date needs to be in the future would it be worth setting the year to 2099?

johnduffell: the tests shouldn't really be relying on the current date at all, I think a load of tests were added about a year ago that did.  Some of them I have changed over to use a hard coded date passed in but I must have missed this one.

PR #1137, opened 2025-06-30
add new mmaProductKey with distinct value for paper+ subscriptions

johnduffell:
Some of our customer are on newspaper plans, and some are on newspaper plus digital.

At the moment they all get the same digital access.  We want to migrate everyone onto the actual digital plans over time.

In the mean time, people already on the non digital plan are being given the option to move across, or stay on the lower price but lose the digital access.

As part of this work, we want it to be clear in manage whether they are entitled to digital benefits.

At the moment, in the response, we return the same tier, and similar name if someone has digital access.

This PR adds a new mmaProductKey property to the subscription and current/futurePlans objects to indicate which product they have.

It is identical to the existing "tier" however if they have a plus paper then it will append " + Digital" on the end.

Tested in CODE, no obvious regressions and the mmaProductKey is coming through with reasonable looking values.

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -1,6 +1,6 @@
 package models
 import com.gu.i18n.Country
-import com.gu.memsub.ProductRatePlanChargeProductType.PaperDay
+import com.gu.memsub.ProductRatePlanChargeProductType.{Digipack, PaperDay}
 import com.gu.memsub._
 import com.gu.memsub.subsv2._
 import com.gu.monitoring.SafeLogger.LogPrefix
@@ -117,7 +117,8 @@ object AccountDetails {
       }
 
       def jsonifyPlan(plan: RatePlan) = Json.obj(
-        "tier" -> getTier(catalog, plan),
+        "tier" -> getTier(catalog, plan), // TODO deprecated - keep until MMA stops using it
+        "mmaProductKey" -> getMMAProductKey(catalog, plan),
         "name" -> externalisePlanName(plan),
         "start" -> plan.effectiveStartDate,
         "end" -> plan.effectiveEndDate,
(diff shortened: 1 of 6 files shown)


membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -141,7 +142,8 @@ object AccountDetails {
       val start = subscriptionData.startDate.getOrElse(paymentDetails.customerAcceptanceDate)
       val end = subscriptionData.endDate.getOrElse(paymentDetails.termEndDate)
       Json.obj(
-        "tier" -> getTier(catalog, mainPlan),
+        "tier" -> getTier(catalog, mainPlan), // TODO deprecated - keep until MMA stops using it
```
johnduffell, line 145: as above, we will delete this once manage has stopped using it

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -219,6 +221,14 @@ object AccountDetails {
       case _ => plan.productName
     }
 
+  def getMMAProductKey(catalog: Catalog, plan: RatePlan): String =
+    plan.product(catalog) match {
+      case _: Product.Newspaper if plan.name(catalog) == "Sunday" => plan.productName + " - Observer"
+      case _: Product.Newspaper if plan.hasChargeType(catalog, Digipack) => plan.productName + " + Digital"
+      // TODO if it's the new tier 3 (digital only) return a different key (later)
```
johnduffell, line 228: this will happen in a later PR once the catalog has it added and we know the chargeType or rate plan name depending on how we want to detect it

membership-common/src/main/scala/com/gu/memsub/ProductFamily.scala
```diff
@@ -16,6 +16,7 @@ object Product {
 
   sealed trait Paper extends ContentSubscription
   sealed trait Weekly extends Paper
+  sealed trait Newspaper extends Paper
```
johnduffell, line 19: I've slotted a new Newspaper trait in so that I can pattern match on any paper but not pick up GW by accident.

j-ruda-guardian (approved):
Thanks for the session!

PR #1138, opened 2025-07-01
keep dependabot happy

johnduffell:
Although the deployable app was actually using 2.15.0, because membership-common was technically still on 2.14, we werre getting a dependabot alert.  As the GHA plugin just scans all modules.

This PR
- bumps to 2.15 latest
- makes sure the override is present on membership-common
- stops sbt-dependency-submission from scanning membership-common itself https://github.com/scalacenter/sbt-dependency-submission?tab=readme-ov-file#--modules-ignore-optional

### DIFF
--- a/.github/workflows/sbt-dependency-graph.yaml
+++ b/.github/workflows/sbt-dependency-graph.yaml
@@ -23,6 +23,8 @@ jobs:
       - name: Submit dependencies
         id: submit
         uses: scalacenter/sbt-dependency-submission@64084844d2b0a9b6c3765f33acde2fbe3f5ae7d3 # v3.1.0
+        with:
+          modules-ignore: membership-common_2.13
       - name: Log snapshot for user validation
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -27,16 +27,15 @@ object Dependencies {
   val mockitoScala = "org.mockito" %% "mockito-scala" % "1.17.14" % Test
   val logback = "ch.qos.logback" % "logback-classic" % "1.4.14"
 
-  val jacksonVersion = "2.15.0"
-  val jacksonDatabindVersion = "2.15.0"
+  val jacksonVersion = "2.15.4"
```
johnduffell, line 30: https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core?p=2

ecguardian (approved):
LGTM 👍

PR #1197, opened 2026-01-09
Fix actions/setup-java version comment to match pinned SHA

tjsilver:
This PR updates the actions/setup-java comment to the tag that matches the pinned SHA. When this workflow was created by Dependency Graph integrator, the version tag comment was not the correct one for the SHA.

This meant that if/when Dependabot were to bump the action to a later version, it would update the commit SHA but not the version comment as it would if they matched.

This fix ensures that the version comment and SHA are matched so that future bumps by Dependabot update both the SHA and version comment.


johnduffell (approved):
Thanks for making the comment correct

PR #1233, opened 2026-04-06
add low redundancy alarm to CFN

johnduffell:
we noticed that the "lack of redundancy" alarm wasn't going via alarms handler.

This is because it was never defined in CFN, so didn't have a tag or a topic to send to.

Some alarms were defined in this PR, but not including this one: https://github.com/guardian/members-data-api/pull/425

This PR defines the alarm in CFN.

Once the PR is merged the alarm can be deleted, including the ones added in the above PR.

Tested by creating a changeset in PROD (I won't execute it until approval)

### DIFF
--- a/cloudformation/membership-attribute-service.yaml
+++ b/cloudformation/membership-attribute-service.yaml
@@ -488,6 +488,35 @@ Resources:
     DependsOn:
       - LoadBalancer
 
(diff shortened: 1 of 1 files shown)


tjmw (approved):
👍

PR #1251, opened 2026-06-09
add extraTaxApplies field to the /mma response

andrade0:
### Why do we need this?

Manage frontend needs to show a "price excludes tax" notice on the account overview and manage pages for Canadian tax-exclusive plans (guardian/manage-frontend#1665). Until now there was nothing in the /mma response to tell it whether a plan is tax exclusive, so the notice was being driven by a mock.

### The changes

We already load the Zuora product catalog, which carries a taxMode on each charge, but we were parsing it and throwing it away. This PR derives an extraTaxApplies flag - a plan applies extra tax when any of its charges has taxMode "TaxExclusive" - exposes it on ProductRatePlan, and adds it at the top level of the /mma response next to billingCountry.

We depend on the plan itself rather than the customer's billing country on purpose: the same plans will be rolled out to US customers later, so the flag should follow the plan from the start.

Tested with unit tests covering the catalog reads (a tax exclusive plan vs a tax inclusive one) and the /mma serialisation.

Tested in CODE and showing the correct value on the /mma response, next to billingCountry. The tax inclusive plans I could subscribe to all return false as expected; the tax exclusive case is covered by the catalog reads unit test, as the tax exclusive plans aren't on the normal acquisition flow in CODE.

Related frontend PR: guardian/manage-frontend#1665
Zuora config adding the tax-exclusive charges: guardian/zuora-config#198

### DIFF
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -42,7 +42,7 @@ object AccountDetails {
 
       val mainPlan = subscription.plan(catalog, today)
       val product = mainPlan.product(catalog)
-      val taxExclusive = catalog.productRatePlans.get(mainPlan.productRatePlanId).exists(_.taxExclusive)
+      val taxExclusive = mainPlan.taxExclusive(catalog)
 
       val paymentMethod = paymentDetails.paymentMethod match {
         case Some(payPal: PayPalMethod) =>
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -42,7 +42,7 @@ object AccountDetails {
 
       val mainPlan = subscription.plan(catalog, today)
       val product = mainPlan.product(catalog)
-      val taxExclusive = mainPlan.taxExclusive(catalog)
+      val extraTaxApplies = mainPlan.extraTaxApplies(catalog)
 
       val paymentMethod = paymentDetails.paymentMethod match {
         case Some(payPal: PayPalMethod) =>
@@ -154,7 +154,7 @@ object AccountDetails {
       ) ++
         regNumber.fold(Json.obj())({ reg => Json.obj("regNumber" -> reg) }) ++
         billingCountry.fold(Json.obj())({ bc => Json.obj("billingCountry" -> bc.name) }) ++
-        Json.obj("taxExclusive" -> taxExclusive) ++
+        Json.obj("extraTaxApplies" -> extraTaxApplies) ++
         Json.obj(
           "joinDate" -> paymentDetails.startDate,
           "optIn" -> !paymentDetails.pendingCancellation,
--- a/membership-attribute-service/app/models/AccountDetails.scala
+++ b/membership-attribute-service/app/models/AccountDetails.scala
@@ -42,6 +42,7 @@ object AccountDetails {
 
       val mainPlan = subscription.plan(catalog, today)
       val product = mainPlan.product(catalog)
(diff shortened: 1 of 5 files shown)


membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
```diff
@@ -242,6 +245,7 @@ case class ProductRatePlan(
     productId: ProductId,
     productRatePlanCharges: Map[ProductRatePlanChargeId, ProductRatePlanChargeProductType],
     private val productTypeOption: Option[ProductType],
+    taxExclusive: Boolean = false,
```
johnduffell: why default to false, shouldn't it always be read?

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/CatJsonReads.scala
```diff
@@ -37,13 +37,22 @@ object CatJsonReads {
     }
   }
 
+  // A plan is tax exclusive when any of its charges carry Zuora's taxMode "TaxExclusive" (the rest are "TaxInclusive" or absent).
+  private val taxExclusiveReads: Reads[Boolean] = new Reads[Boolean] {
+    override def reads(json: JsValue): JsResult[Boolean] = json match {
+      case JsArray(charges) => JsSuccess(charges.exists(charge => (charge \ "taxMode").asOpt[String].contains("TaxExclusive")))
+      case _ => JsSuccess(false)
```
johnduffell: if it's not an array then something is wrong we should not just assume it's not tax exclusive

membership-attribute-service/app/models/AccountDetails.scala
```diff
@@ -42,6 +42,7 @@ object AccountDetails {
 
       val mainPlan = subscription.plan(catalog, today)
       val product = mainPlan.product(catalog)
+      val taxExclusive = mainPlan.taxExclusive(catalog)
```
johnduffell: I don't like the negative/double negative naming, can we change it to something positive naming, either `taxIncluded` or `extraTaxApplies` or perhaps keep a similar enum to what zuora emits i.e. `TaxExclusive | TaxInclusive`  Otherwise you can end up tied up in knots of double negatives.

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
andrade0: Updated

membership-attribute-service/app/models/AccountDetails.scala
andrade0: Done

membership-common/src/main/scala/com/gu/memsub/subsv2/reads/CatJsonReads.scala
andrade0: Done ! thanks the usefull feedbacks @johnduffell  !

shtukas (approved):

membership-common/src/main/scala/com/gu/memsub/subsv2/Plan.scala
johnduffell: did this get reverted as I can still see it in the diff, I don't think it was pushed?

membership-attribute-service/app/models/AccountDetails.scala
johnduffell: thanks the new name sounds much better, easier to get your head around 👍

johnduffell (approved):
thanks, looks good apart from a couple of niggles
1. need to push the change I flagged in the comment
2. just check the description as it still refers to the old name
other than that thanks for doing this work 👏

PR #1259, opened 2026-07-03
replace SOAP amend-preview with REST billing-preview for next payment

andrade0:
### Why do we need this?

Zuora no longer supports the SOAP API for new development, only bug fixes, and members-data-api is still one of the heaviest SOAP users. This migrates the `soap:amend` calls (around 37k in the last 7 days, with a high error rate of ~2,257) off SOAP.

The `soap:amend` traffic here is not actually amending anything. It is a hack that fires a dummy preview amend to project a customer's next payment on the manage-account page (the `/user-attributes/me/mma` endpoints). The hack forces every subscription to an evergreen term to get the projection, which Zuora rejects for many subscription shapes (cancelled, expired, mid-amendment), which is where the errors come from. Those errors were silently swallowed, so the user just saw no next payment.

### The changes

- New `getBillingPreview` on the Zuora REST service, calling `operations/billing-preview` with `assumeRenewal`.
- `PaymentService.getNextBill` now uses billing-preview (filtered to the target subscription) instead of the SOAP amend, mapping the invoice items to the existing shape. Kept like for like: the projected amount includes tax, we look ~30 months ahead to still catch the first paid bill of long promotional periods (e.g. the 24 months free student offer).
- On any failure we show no next payment and never fail the whole `/mma` call, exactly as before, but it is now logged so the failure is visible in the MDAPI logs instead of being silent.
- Removed the now-unused SOAP preview code (`previewInvoices`, `getSubscription`, `PreviewInvoicesViaAmend`), so there is no more `soap:amend` traffic from this path.
- Added a unit test on the invoice-item mapping (that the price includes tax).

I validated billing-preview against real CODE data, including an already-billed annual subscription where `assumeRenewal` correctly projects the renewal charge a year out, so the evergreen trick is not needed. Ran the membership-common and membership-attribute-service test suites locally, all green.

Asana card: Zuora: Migrate remaining SOAP calls to REST API.

Follow-up for a later PR: the larger `soap:query` share (~225k: getAccountIds, getAccount, getContact, getPaymentSummary, getPaymentMethod) still needs migrating to the REST `action/query` / object endpoints.

### DIFF
--- a/membership-attribute-service/app/services/zuora/payment/PaymentService.scala
+++ b/membership-attribute-service/app/services/zuora/payment/PaymentService.scala
@@ -102,8 +102,7 @@ class PaymentService(zuoraService: ZuoraSoapService, restService: ZuoraRestServi
     } yield bill
 
   /** billing-preview is account-scoped, so we request the whole account and keep only the target subscription's items. A preview failure only means
-    * we show no next payment; it must never fail the whole /mma call, so we log it and carry on with an empty list (as before, but now with a log
-    * line so the failure is visible).
+    * we show no next payment; it must never fail the whole /mma call, so we log it and carry on.
     */
   private def getPreviewInvoiceItems(subId: Id, accountId: AccountId, targetDate: LocalDate)(implicit
       logPrefix: LogPrefix,
--- a/membership-attribute-service/app/services/zuora/payment/PaymentService.scala
+++ b/membership-attribute-service/app/services/zuora/payment/PaymentService.scala
@@ -101,31 +101,20 @@ class PaymentService(zuoraService: ZuoraSoapService, restService: ZuoraRestServi
         .toList
     } yield bill
 
-  // Zuora's billing-preview is account-scoped, so we ask for the whole account and keep only the target subscription's
-  // items, mapping them to the shape BillingSchedule expects. price includes tax to stay like-for-like with the old
-  // SOAP amend-with-preview. On failure we return no items (no preview shown), matching the previous behaviour.
+  // billing-preview is account-scoped, so we request the whole account and keep only the target subscription's items.
+  // A Zuora error means we show no next payment (as before); we let unexpected failures propagate rather than hide them.
+  // TODO if an account has several subscriptions this runs once per subscription; it could be fetched once per account
+  // and shared to save Zuora calls.
   private def getPreviewInvoiceItems(subId: Id, accountId: AccountId, targetDate: LocalDate)(implicit
       logPrefix: LogPrefix,
   ): Future[Seq[Queries.PreviewInvoiceItem]] =
-    restService
-      .getBillingPreview(accountId, targetDate)
-      .map {
-        case \/-(items) =>
-          items.collect {
-            case item if item.subscriptionId == subId.get =>
-              Queries.PreviewInvoiceItem(
-                price = (item.chargeAmount + item.taxAmount).toFloat,
-                serviceStartDate = new LocalDate(item.serviceStartDate),
-                serviceEndDate = new LocalDate(item.serviceEndDate),
-                productId = "",
-                productRatePlanChargeId = "",
-                chargeName = item.chargeName,
-                unitPrice = (item.chargeAmount + item.taxAmount).toFloat,
-              )
-          }
-        case -\/(_) => Nil
-      }
-      .recover { case _ => Nil }
+    restService.getBillingPreview(accountId, targetDate).map {
+      case \/-(items) =>
+        items.collect { case item if item.subscriptionId == subId.get => PaymentService.toPreviewInvoiceItem(item) }
+      case -\/(error) =>
+        logger.warn(s"could not get billing preview for account ${accountId.get}, showing no next payment: $error")
+        Nil
+    }
 
   def getPaymentMethod(maybePaymentMethodId: Option[String], defaultMandateIdIfApplicable: Option[String] = None)(implicit
       logPrefix: LogPrefix,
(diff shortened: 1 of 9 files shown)


shtukas (approved):

johnduffell:
> On a Zuora error we show no next payment as before, but it is now logged rather than silently swallowed, and unexpected failures are allowed to surface instead of being hidden.

what does "surface" mean? Does this mean the whole call will fail?  In the past it did log it in the MDAPI logs but it didn't fail the whole call, which is a good thing as we don't want to throw 500 errors.

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
```diff
@@ -98,6 +101,21 @@ class PaymentService(zuoraService: ZuoraSoapService)(implicit ec: ExecutionConte
         .toList
     } yield bill
 
+  // billing-preview is account-scoped, so we request the whole account and keep only the target subscription's items.
+  // A Zuora error means we show no next payment (as before); we let unexpected failures propagate rather than hide them.
+  // TODO if an account has several subscriptions this runs once per subscription; it could be fetched once per account
```
johnduffell: it's relatively rare to have multiple active subs in one account, so I'd say don't worry about that optimisation

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
```diff
@@ -98,6 +101,21 @@ class PaymentService(zuoraService: ZuoraSoapService)(implicit ec: ExecutionConte
         .toList
     } yield bill
 
+  // billing-preview is account-scoped, so we request the whole account and keep only the target subscription's items.
+  // A Zuora error means we show no next payment (as before); we let unexpected failures propagate rather than hide them.
```
johnduffell: I think this comment is wrong, on line 116 it seems like we just assume no future payments if there was an error.
I think the behaviour is fine (although you might consider adding a metric and alarm, not sure if that's worth it though)

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
```diff
@@ -98,6 +101,21 @@ class PaymentService(zuoraService: ZuoraSoapService)(implicit ec: ExecutionConte
...
+  // TODO if an account has several subscriptions this runs once per subscription; it could be fetched once per account
+  // and shared to save Zuora calls.
+  private def getPreviewInvoiceItems(subId: Id, accountId: AccountId, targetDate: LocalDate)(implicit
+      logPrefix: LogPrefix,
+  ): Future[Seq[Queries.PreviewInvoiceItem]] =
+    restService.getBillingPreview(accountId, targetDate).map {
+      case \/-(items) =>
+        items.collect { case item if item.subscriptionId == subId.get => PaymentService.toPreviewInvoiceItem(item) }
```
johnduffell: optional: might be easier to digest if we filter and then map
```suggestion
        items
          .filter(_.subscriptionId == subId.get)
          .map(PaymentService.toPreviewInvoiceItem)
```

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
```diff
@@ -109,3 +127,22 @@ class PaymentService(zuoraService: ZuoraSoapService)(implicit ec: ExecutionConte
       .getOrElse(Future.successful(None))
 
 }
+
+object PaymentService {
+
+  // Map a Zuora billing-preview invoice item to the internal preview shape BillingSchedule consumes.
```
johnduffell: is the idea in a later PR to push the new rest type through all the code, or to keep this mapping layer permanently?
(It would be nice to get rid of the mapping layer, but if it's a useful simplification of the type or too much code to touch then it's worth keeping.)

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
```diff
@@ -199,6 +200,24 @@ object ZuoraRestService {
   case class RestQuery(queryString: String)
   implicit val restQueryWrites = Json.writes[RestQuery]
 
+  // Billing Preview (REST replacement for the SOAP amend-with-preview hack). assumeRenewal projects the next
+  // charge for subs already billed for their current term, so the old "evergreen" trick is no longer needed.
```
johnduffell: would be useful to link to the zuora docs here or next to getBillingPreview if you have it handy

Also I notice you're generally using `//` style comments, but I think hover-help works better with `/*...*/` or `/** ... */` style comments.  these style of comments have more chance to get separated from their code too.  So I'd prefer the block style if possible.

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
```diff
@@ -98,6 +101,26 @@ class PaymentService(zuoraService: ZuoraSoapService)(implicit ec: ExecutionConte
         .toList
     } yield bill
 
+  // billing-preview is account-scoped, so we request the whole account and keep only the target subscription's items.
+  // A preview failure only means we show no next payment; it must never fail the whole /mma call, so we log it and
+  // carry on with an empty list (as before, but now with a log line so the failure is visible).
```
johnduffell: this is the existing failure in the log

Image

I think you can just remove the comment about the log line, people can just look at the code

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
andrade0: Agreed, it's rare enough that it isn't worth it. Dropped the TODO.

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
andrade0: Good catch, the comment contradicted the code. Fixed it to match. I've left a metric and alarm out for now but happy to add one as a follow-up if you think it's worth it.

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
andrade0: Applied, thanks, it does read better.

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
andrade0: For now it's a deliberate mapping layer rather than pushing the new REST type through all the code, which keeps this PR small and lets BillingSchedule stay unchanged. Longer term, once the rest of the SOAP to REST work lands, we could swap the internal PreviewInvoiceItem for a REST-native type and drop the adapter. It's a small contained mapping so I'd lean towards keeping it for now and revisiting later.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
andrade0: Done, added a link to the Zuora billing preview docs and switched these over to block comments.

andrade0:
On the 500 point, you're right, "surface" was a bad word choice from me. It's back to graceful degradation now: on any failure (a Zuora error or an unexpected one) it logs and returns no next payment, so the /mma call never fails, exactly like before but with a log line so it isn't silent anymore. Pushed the fixes.

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
andrade0: Good point, dropped that part of the comment.

PR #1260, opened 2026-07-06
match billing preview items by subscription number, not zuora id

andrade0:
### Why do we need this?

Follow-up to #1259 (SOAP amend to REST billing-preview). Testing that change in CODE showed the next payment coming back as £0.00 for every subscription. The cause: `getNextBill` filtered the account-scoped billing-preview items by the Zuora subscription id, but with `assumeRenewal` the projected items carry a simulated renewal subscription id, so the filter dropped every item and we ended up with no bills (nextPaymentPrice 0). Because a preview failure degrades gracefully (no 500), it was invisible until we looked at the actual value in CODE.

### The changes

- Filter the billing-preview items by subscription number, which is stable across renewals, instead of the Zuora id (the DTO now carries `subscriptionNumber`).

Verified against real CODE data: the next payment for a real subscription goes from £0.00 back to the correct £64.99, and the full test suites are green.

Follow-up to #1259. Asana: Zuora Migrate remaining SOAP calls to REST API.

### DIFF
--- a/membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
+++ b/membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
@@ -208,7 +208,7 @@ object ZuoraRestService {
   implicit val billingPreviewRequestWrites = Json.writes[BillingPreviewRequest]
 
   case class BillingPreviewInvoiceItem(
-      subscriptionId: String,
+      subscriptionNumber: String,
       chargeAmount: Double,
       taxAmount: Double,
       serviceStartDate: String,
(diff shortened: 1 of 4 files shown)


enesihsa (approved):

shtukas (approved):

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
```diff
@@ -208,7 +208,7 @@ object ZuoraRestService {
   implicit val billingPreviewRequestWrites = Json.writes[BillingPreviewRequest]
 
   case class BillingPreviewInvoiceItem(
-      subscriptionId: String,
+      subscriptionNumber: String,
```
johnduffell: let's keep the type safety, then we don't have to unwrap them in `getPreviewInvoiceItems`
```suggestion
      subscriptionNumber: SubscriptionNumber,
```

johnduffell (approved):
looks good, just one thing to improve the type safety

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
andrade0: Good call, done in 0c2990d. Typed the field as SubscriptionNumber so the filter compares directly, no unwrapping.

PR #1261, opened 2026-07-07
replace SOAP query reads with Zuora REST equivalents

andrade0:
### Why do we need this?

Part of the Zuora SOAP to REST migration for members-data-api. This is the `soap:query` side (the billing-preview/amend side was #1259 and #1260). The five remaining read methods in `ZuoraSoapService` (`getAccount`, `getAccountIds`, `getContact`, `getPaymentSummary`, `getPaymentMethod`) went through the Zuora SOAP query API. They now hit the equivalent REST endpoints (`object/{type}/{id}` and `action/query`).

### The changes

- New JSON readers (`ZuoraQueryReads`) that map the REST responses back to the existing SOAP query case classes, so the method signatures and return types are unchanged and no caller is affected.
- The five read methods now use the REST `SimpleClient`. Zuora returns some fields (e.g. credit card expiry) as JSON numbers where the SOAP path had strings, so the reader normalises them.
- These are core reads, so they still fail on a Zuora error like the old SOAP `queryOne`/`query` did, rather than degrading. This is not the best-effort side-preview from #1259.

Validated against real CODE data: all five methods return the same values as before.

Asana: Zuora Migrate remaining SOAP calls to REST API. Follows #1259 and #1260.

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -109,7 +109,7 @@ class TouchpointComponents(
         metrics = zuoraMetrics,
       )
 
-    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraSoapClient) with HealthCheckableService {
+    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraSoapClient, zuoraRestClient) with HealthCheckableService {
       override def checkHealth: Boolean = zuoraSoapClient.isReady
     }
 
--- a/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
+++ b/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
@@ -57,8 +57,7 @@ class ZuoraSoapService(soapClient: soap.Client, restClient: rest.SimpleClient[Fu
       .map(_.valueOr(error => throw QueryError(s"Zuora REST query '$zoql' failed: $error")).records)
 
   def getAccountIds(contactId: ContactId)(implicit logPrefix: LogPrefix): Future[List[AccountId]] =
-    query[AccountIdRecord](s"select Id from account where crmId = '${contactId.salesforceAccountId}'")
-      .map(_.map(record => AccountId(record.id)))
+    query[AccountId](s"select Id from account where crmId = '${contactId.salesforceAccountId}'")
 
   def getAccount(accountId: AccountId)(implicit logPrefix: LogPrefix): Future[SoapQueries.Account] =
     getObject[SoapQueries.Account](s"object/account/${accountId.get}")
@@ -130,11 +129,21 @@ class ZuoraSoapService(soapClient: soap.Client, restClient: rest.SimpleClient[Fu
     } yield result
   }
 
-  def getPaymentSummary(subscriptionNumber: S.SubscriptionNumber, accountCurrency: Currency)(implicit logPrefix: LogPrefix): Future[PaymentSummary] =
-    query[SoapQueries.InvoiceItem](
-      s"select Id, ChargeAmount, TaxAmount, ServiceStartDate, ServiceEndDate, ChargeNumber, ProductName, SubscriptionId " +
-        s"from invoiceitem where SubscriptionNumber = '${subscriptionNumber.getNumber}'",
-    ).map(invoiceItems => PaymentSummary(latestInvoiceItems(invoiceItems), accountCurrency))
+  def getPaymentSummary(subscriptionNumber: S.SubscriptionNumber, accountCurrency: Currency)(implicit
+      logPrefix: LogPrefix,
+  ): Future[PaymentSummary] = {
+    val zoql = List(
+      "select Id, ChargeAmount, TaxAmount, ServiceStartDate, ServiceEndDate, ChargeNumber, ProductName, SubscriptionId",
+      "from invoiceitem",
+      s"where SubscriptionNumber = '${subscriptionNumber.getNumber}'",
+    ).mkString(" ")
+    for {
+      invoiceItems <- query[SoapQueries.InvoiceItem](zoql)
+    } yield {
+      val filteredInvoices = latestInvoiceItems(invoiceItems)
+      PaymentSummary(filteredInvoices, accountCurrency)
+    }
+  }
 
   def getPaymentMethod(id: String)(implicit logPrefix: LogPrefix): Future[SoapQueries.PaymentMethod] =
     getObject[SoapQueries.PaymentMethod](s"object/payment-method/$id")
--- a/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
+++ b/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
@@ -45,16 +45,16 @@ class ZuoraSoapService(soapClient: soap.Client, restClient: rest.SimpleClient[Fu
 
   import ZuoraSoapService._
 
-  // These reads used to go through the SOAP query API; they now hit the equivalent Zuora REST endpoints (object/{type}/{id} and action/query),
-  // mapping the responses back to the same case classes so callers are unchanged. A missing record or REST error fails the Future, matching the
-  // old queryOne behaviour.
+  /* These reads used to go through the SOAP query API; they now hit the equivalent Zuora REST endpoints (object/{type}/{id} and action/query),
+     mapping the responses back to the same case classes so callers are unchanged. getObject fails the Future if the object is missing or the call
+     errors, like the old queryOne; query returns the records (empty if none) and only fails on a REST error, like the old plural query. */
   private def getObject[A: Reads](url: String)(implicit logPrefix: LogPrefix): Future[A] =
-    restClient.get[A](url).map(_.valueOr(error => throw new RuntimeException(s"Zuora REST get '$url' failed: $error")))
+    restClient.get[A](url).map(_.valueOr(error => throw QueryError(s"Zuora REST get '$url' failed: $error")))
 
   private def query[A: Reads](zoql: String)(implicit logPrefix: LogPrefix): Future[List[A]] =
     restClient
       .post[RestQuery, QueryResponse[A]]("action/query", RestQuery(zoql))
-      .map(_.valueOr(error => throw new RuntimeException(s"Zuora REST query '$zoql' failed: $error")).records)
+      .map(_.valueOr(error => throw QueryError(s"Zuora REST query '$zoql' failed: $error")).records)
 
   def getAccountIds(contactId: ContactId)(implicit logPrefix: LogPrefix): Future[List[AccountId]] =
     query[AccountIdRecord](s"select Id from account where crmId = '${contactId.salesforceAccountId}'")
--- a/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
+++ b/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
@@ -17,6 +17,8 @@ import com.gu.zuora.soap.models.Results.{CreateResult, UpdateResult}
 import com.gu.zuora.soap.models.errors._
 import com.gu.zuora.soap.models.{PaymentSummary, Queries => SoapQueries}
 import com.gu.zuora.soap.writers.Command.createPaymentMethodWrites
+import com.gu.zuora.rest.ZuoraQueryReads._
+import play.api.libs.json.Reads
 
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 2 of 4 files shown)


shtukas (approved):

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -109,7 +109,7 @@ class TouchpointComponents(
         metrics = zuoraMetrics,
       )
 
-    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraSoapClient) with HealthCheckableService {
+    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraSoapClient, zuoraRestClient) with HealthCheckableService {
```
johnduffell, line 112: seems a bit odd that the "soap" service takes a "rest" client?

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
```diff
@@ -0,0 +1,136 @@
...
+import com.gu.i18n.Currency
+import com.gu.zuora.ZuoraLookup
+import com.gu.zuora.api.PaymentGateway
+import com.gu.zuora.soap.models.Queries
+import org.joda.time.LocalDate
+import play.api.libs.json._
+
+/** JSON readers that let ZuoraSoapService's read methods hit Zuora REST (object/{type}/{id} and action/query) instead of SOAP query, mapping the
+  * responses back to the existing SOAP query case classes so callers are unaffected. Field names and parsing mirror the old SOAP readers in
+  * com.gu.zuora.soap.Readers. See https://developer.zuora.com/api-references/api/operation/Object_GETAccount
```
johnduffell: this seems to be a 404 for me!

Image

johnduffell: I don't think one docs link would cover them all anyway, better to put the individual link on each function/type/reads  rather than at the top

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
```diff
@@ -0,0 +1,136 @@
...
+  implicit val accountIdRecordReads: Reads[AccountIdRecord] = (__ \ "Id").read[String].map(AccountIdRecord)
+
+  /* Zuora returns some fields (e.g. credit card expiry) as JSON numbers; the SOAP readers treated everything as strings, so normalise to String. */
+  private def optString(json: JsValue, field: String): JsResult[Option[String]] =
+    (json \ field)
+      .validateOpt[JsValue]
+      .map(_.flatMap {
+        case JsString(s) => Some(s)
+        case JsNumber(n) => Some(n.toBigInt.toString)
+        case _ => None
```
johnduffell: shouldn't we return a `case unknown => JsError("this wasn't a string or number as expected: $unknown"` rather than just filtering out anything we don't understand?

johnduffell: also if we genuinely want a string and it's defined as a number in the api, we don't need the "try both" logic, we can just validate as a JsNumber and convert to a string, which makes it deterministic and avoids the error handling necessity.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
```diff
@@ -0,0 +1,136 @@
...
+    } yield Queries.Account(
+      id = id,
+      billToId = billToId,
+      soldToId = soldToId,
+      billCycleDay = billCycleDay,
+      creditBalance = creditBalance,
+      currency = currency.flatMap(Currency.fromString),
+      defaultPaymentMethodId = defaultPaymentMethodId,
+      sfContactId = sfContactId,
+      paymentGateway = paymentGateway.flatMap(PaymentGateway.getByName),
```
johnduffell, line 48: will this just filter it out if it's a new/unknown gateway?  is that what we want?

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  case class RestQuery(queryString: String)
+  implicit val restQueryWrites: Writes[RestQuery] = Json.writes[RestQuery]
+
+  case class QueryResponse[A](records: List[A])
+  implicit def queryResponseReads[A](implicit r: Reads[A]): Reads[QueryResponse[A]] =
+    (__ \ "records").read[List[A]].map(QueryResponse(_))
+
+  case class AccountIdRecord(id: String)
+  implicit val accountIdRecordReads: Reads[AccountIdRecord] = (__ \ "Id").read[String].map(AccountIdRecord)
```
johnduffell: it looks like the AccountIdRecords is just a temporary wrapper which it then unwraps and puts it into an AccountId instead.  Why not just wrap it in an AccountId in the first place and get rid of the temporary wrapper?

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
```diff
@@ -117,14 +131,12 @@ class ZuoraSoapService(soapClient: soap.Client)(implicit ec: ExecutionContext) e
...
   def getPaymentSummary(subscriptionNumber: S.SubscriptionNumber, accountCurrency: Currency)(implicit logPrefix: LogPrefix): Future[PaymentSummary] =
-    for {
-      invoiceItems <- soapClient.query[SoapQueries.InvoiceItem](SimpleFilter("SubscriptionNumber", subscriptionNumber.getNumber))
-    } yield {
-      val filteredInvoices = latestInvoiceItems(invoiceItems)
-      PaymentSummary(filteredInvoices, accountCurrency)
-    }
+    query[SoapQueries.InvoiceItem](
+      s"select Id, ChargeAmount, TaxAmount, ServiceStartDate, ServiceEndDate, ChargeNumber, ProductName, SubscriptionId " +
+        s"from invoiceitem where SubscriptionNumber = '${subscriptionNumber.getNumber}'",
```
johnduffell: might be slightly easier to read like this?
```suggestion
      List(
        "select Id, ChargeAmount, TaxAmount, ServiceStartDate, ServiceEndDate, ChargeNumber, ProductName, SubscriptionId",
        "from invoiceitem",
        s"where SubscriptionNumber = '${subscriptionNumber.getNumber}'",
      ).mkString(" ")
```

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
```diff
@@ -117,14 +131,12 @@ class ZuoraSoapService(soapClient: soap.Client)(implicit ec: ExecutionContext) e
...
+    ).map(invoiceItems => PaymentSummary(latestInvoiceItems(invoiceItems), accountCurrency))
```
johnduffell: could have left it as a for comprehension with the old lines 123/124 split out?

johnduffell (commented):
great work, thanks for switching those!  I think the existing naming is not great that it's a "soap" or "rest" client rather than it being the functionality that matters.  I can see the logic of keeping the methods in place and unchanged.  I wonder whether a pre- or post- refactor might make sense to move them into the rest client and tidy up the types, for consistency?
I've left a few small comments and/or questions about bits that were unclear to me.

membership-attribute-service/app/components/TouchpointComponents.scala
andrade0: Yeah, it does read oddly. It's the transitional state: the service still owns the SOAP write methods (createPaymentMethod etc.), so it needs both clients until those move too. I kept the reads in place to keep the diff small and the callers untouched while the service is on its way out, but I agree the name stops making sense here. Happy to do the move into the rest client (and the type tidy-up) as a follow-up so this one stays focused on the SOAP removal.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
andrade0: Good catch, sorry about that. The object endpoints live under the older API section, so the path needs `older-api` rather than `api`, e.g. https://developer.zuora.com/api-references/older-api/operation/Object_GETAccount. I'll fix it.

andrade0: Agreed, one link was never going to cover them all. I'll drop the top-level one and put the specific operation link on each reads/method instead.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
andrade0: Fair point, silently dropping an unexpected shape isn't great. I'll make it fail with a JsError instead of coercing to None.

andrade0: Yes, that's cleaner. The fields aren't actually mixed: the card mask is a string and the expiry month/year are numbers, so I'll type each one deterministically (String for the mask, Int converted to String for the expiry) and drop the try-both helper altogether. That also removes the swallow you flagged above.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
andrade0: This mirrors the existing SOAP reader, which also did flatMap(PaymentGateway.getByName) and dropped an unknown gateway to None, so it's like-for-like and the field was already Option. I'd rather not change that behaviour in this PR, but you're right it's a latent question, happy to raise failing loud on an unknown gateway separately if we think it's worth it.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
andrade0: Good point, the wrapper is pointless. I'll read straight into an AccountId and get rid of AccountIdRecord.

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
andrade0: Nice, that does read better. I'll take the suggestion.

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
andrade0: Sure, I'll put it back as a for comprehension with the filtered invoices split out.

andrade0:
Thanks for the thorough look. Agreed the soap/rest naming is the weak spot here, the class name is a bit of a misnomer now it does both. I kept the methods in place deliberately to keep the diff small and the callers untouched while the service is being wound down, but I think you're right that moving these into the rest client and tidying the types is the cleaner end state. I'll do the small fixes from the inline comments in this PR and raise the move/rename as a follow-up so this one stays focused on removing the SOAP calls.

membership-attribute-service/app/components/TouchpointComponents.scala
johnduffell: sounds good, followup makes sense to keep each PR clear and self contained.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
johnduffell: fair enough, would be useful to have a GH link to the existing code that you copied, to make it easier to compare.(in a GH comment is fine)

As a more general point, it's useful when you raise your own PR to go through and self-review/comment on these things yourself with references where necessary to make it easier for reviewers to focus on the important parts.  This provides the additional context without permanently cluterring the code with information that can become outdated, and it's still available in future in case we're investigating something in future.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
```diff
@@ -9,7 +10,7 @@ import play.api.libs.json._
 
 /** JSON readers that let ZuoraSoapService's read methods hit Zuora REST (object/{type}/{id} and action/query) instead of SOAP query, mapping the
   * responses back to the existing SOAP query case classes so callers are unaffected. Field names and parsing mirror the old SOAP readers in
-  * com.gu.zuora.soap.Readers. See https://developer.zuora.com/api-references/api/operation/Object_GETAccount
+  * com.gu.zuora.soap.Readers. Each reader links the Zuora operation it parses.
```
johnduffell, line 13: nitpick: there's no need to say `Each reader links the Zuora operation it parses.`, the links can speak for themselves.

As a side point unrelated to this PR, I guess you added it because it's good context for future AI, but it's a lot harder for humans to quickly read comments, so we should make sure they have maximum impact for each line.
I wonder if there's some kind of convention for e.g. ai-context.md files, so this kind of thing can still exist but not affect humans?

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala
```diff
@@ -20,19 +21,10 @@ object ZuoraQueryReads {
...
-  /* Zuora returns some fields (e.g. credit card expiry) as JSON numbers; the SOAP readers treated everything as strings, so normalise to String. */
-  private def optString(json: JsValue, field: String): JsResult[Option[String]] =
-    (json \ field)
-      .validateOpt[JsValue]
-      .map(_.flatMap {
-        case JsString(s) => Some(s)
-        case JsNumber(n) => Some(n.toBigInt.toString)
-        case _ => None
-      })
+  /** @see https://developer.zuora.com/api-references/api/operation/Action_POSTquery */
```
johnduffell, line 24: Where are you getting the links?  for some reason I get redirected immediately from your link:
`https://developer.zuora.com/api-references/api/operation/Action_POSTquery` to:
`https://developer.zuora.com/v1-api-reference/api/actions/action_postquery`
normally I would find the right part of the docs and copy the link from the link thing

Image

PR #1263, opened 2026-07-13
replace SOAP payment method writes with Zuora REST equivalents

andrade0:
### Why do we need this?

The last part of the Zuora SOAP to REST migration for members-data-api (after #1259 / #1260 billing-preview and #1261 the query reads). The remaining SOAP calls were the payment-method writes in `ZuoraSoapService`: creating a payment method and setting it as the account default, used by the update-card and update-direct-debit flows. They now go through the equivalent Zuora REST endpoints, so `ZuoraSoapService` no longer makes any SOAP calls.

### The changes

- Account payment updates (set/clear default payment method, gateway, autoPay) now use `PUT /v1/accounts/{id}`. A null `defaultPaymentMethodId` clears the default, the REST equivalent of the SOAP `fieldsToNull`.
- Payment method creation now uses `POST /v1/object/payment-method`, which takes the same zObject fields the SOAP create used (Direct Debit bank transfer, and Stripe credit-card reference).
- The three-step ordering (clear + set gateway, create, set default) is kept, because Zuora still validates that the account gateway matches its default method's gateway.
- New writers in `ZuoraPaymentWrites` plus unit tests; the SOAP client is no longer passed to `ZuoraSoapService`.

Tested end to end on the deployed CODE build against the real Zuora sandbox. Both write flows go through manage cleanly: updating to a direct debit (GoCardless gateway) and updating a card (Stripe gateway), each one created in Zuora and set as the account default via the three-step dance, then confirmed server side on `/api/me/mma`. The reads (account, contact, payment summary, payment method) come back correct too, including the card expiry that Zuora REST returns as numbers rather than strings. Reads and writes all good.

### Asana / PRs

Asana: Zuora Migrate remaining SOAP calls to REST API. Follows #1259, #1260, #1261. Follow-up: delete the now-dead `com.gu.zuora.soap` package and move the health check off the SOAP client.

### DIFF
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -8,11 +8,10 @@ import com.gu.identity.play.IdentityPlayAuthService
 import com.gu.memsub.subsv2.Catalog
 import com.gu.memsub.subsv2.services.{CatalogService, FetchCatalog, SubscriptionService}
 import com.gu.monitoring.SafeLogger.LogPrefix
-import com.gu.monitoring.{SafeLogging, ZuoraMetrics}
+import com.gu.monitoring.SafeLogging
 import com.gu.okhttp.RequestRunners
 import com.gu.touchpoint.TouchpointBackendConfig
 import com.gu.zuora.rest.SimpleClient
-import com.gu.zuora.soap.Client
 import com.gu.zuora.{ZuoraSoapService, rest}
 import com.typesafe.config.Config
 import configuration.Stage
@@ -37,7 +36,6 @@ import software.amazon.awssdk.auth.credentials.{
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.dynamodb.{DynamoDbAsyncClient, DynamoDbAsyncClientBuilder}
 
-import java.util.concurrent.TimeUnit.SECONDS
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -50,7 +48,7 @@ class TouchpointComponents(
     subscriptionServiceOverride: Option[SubscriptionService[Future]] = None,
     zuoraRestServiceOverride: Option[ZuoraRestService] = None,
     catalogServiceOverride: Option[Future[Catalog]] = None,
-    zuoraServiceOverride: Option[ZuoraSoapService with HealthCheckableService] = None,
+    zuoraServiceOverride: Option[ZuoraSoapService] = None,
     patronsStripeServiceOverride: Option[BasicStripeService] = None,
     chooseStripeOverride: Option[ChooseStripe] = None,
 )(implicit
@@ -99,20 +97,8 @@ class TouchpointComponents(
   lazy val supporterProductDataService: SupporterProductDataService =
     supporterProductDataServiceOverride.getOrElse(dynamoSupporterProductDataService)
 
-  private val zuoraMetrics = new ZuoraMetrics(stage.value, configuration.ApplicationName.applicationName)
-
   lazy val zuoraSoapService = {
-    lazy val zuoraSoapClient =
-      new Client(
-        apiConfig = backendConfig.zuoraSoap,
-        httpClient = RequestRunners.configurableFutureRunner(timeout = Duration(30, SECONDS)),
-        metrics = zuoraMetrics,
-      )
-
-    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraRestClient) with HealthCheckableService {
-      override def checkHealth: Boolean = zuoraSoapClient.isReady
-    }
-
+    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraRestClient)
     zuoraServiceOverride.getOrElse(simpleZuoraSoapService)
   }
 
--- a/membership-attribute-service/app/components/TouchpointComponents.scala
+++ b/membership-attribute-service/app/components/TouchpointComponents.scala
@@ -109,7 +109,7 @@ class TouchpointComponents(
         metrics = zuoraMetrics,
       )
 
-    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraSoapClient, zuoraRestClient) with HealthCheckableService {
+    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraRestClient) with HealthCheckableService {
       override def checkHealth: Boolean = zuoraSoapClient.isReady
     }
 
--- a/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
+++ b/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
@@ -41,9 +41,8 @@ class ZuoraSoapService(restClient: rest.SimpleClient[Future])(implicit ec: Execu
 
   import ZuoraSoapService._
 
-  /* These reads used to go through the SOAP query API; they now hit the equivalent Zuora REST endpoints (object/{type}/{id} and action/query),
-     mapping the responses back to the same case classes so callers are unchanged. getObject fails the Future if the object is missing or the call
-     errors, like the old queryOne; query returns the records (empty if none) and only fails on a REST error, like the old plural query. */
+  /* getObject fetches a single object via object/{type}/{id} and fails the Future if it is missing or the call errors; query runs a ZOQL query via
+     action/query and returns the records (empty if none), failing only on a REST error. */
   private def getObject[A: Reads](url: String)(implicit logPrefix: LogPrefix): Future[A] =
     restClient.get[A](url).map(_.valueOr(error => throw QueryError(s"Zuora REST get '$url' failed: $error")))
 
@@ -61,9 +60,8 @@ class ZuoraSoapService(restClient: rest.SimpleClient[Future])(implicit ec: Execu
   def getContact(contactId: String)(implicit logPrefix: LogPrefix): Future[SoapQueries.Contact] =
     getObject[SoapQueries.Contact](s"object/contact/$contactId")
 
-  /* These payment-method writes used to go through the SOAP create/update API; they now hit the equivalent Zuora REST endpoints: the account payment
-     fields via PUT accounts/{id}, and the payment method itself via POST object/payment-method (which takes the same zObject fields the SOAP create
-     used). A REST error or an unsuccessful response fails the Future, matching the old behaviour where a non-success SOAP response raised. */
+  /* The account payment fields are updated via PUT accounts/{id}, and the payment method is created via POST object/payment-method. A REST error or an
+     unsuccessful response fails the Future. */
   private def updateAccountPayment(
       accountId: AccountId,
       defaultPaymentMethodId: Option[String],
--- a/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
+++ b/membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
@@ -9,14 +9,14 @@ import com.gu.salesforce.ContactId
 import com.gu.stripe.Stripe
 import com.gu.zuora.api.{PaymentGateway}
 import com.gu.zuora.soap._
-import com.gu.zuora.soap.models.Commands.CreatePaymentMethod
+import com.gu.zuora.soap.models.Commands.{CreatePaymentMethod, CreditCardReferenceTransaction}
 import com.gu.zuora.soap.models.Results.UpdateResult
 import com.gu.zuora.soap.models.errors._
 import com.gu.zuora.soap.models.{PaymentSummary, Queries => SoapQueries}
 import com.gu.zuora.rest.ZuoraQueryReads._
 import com.gu.zuora.rest.ZuoraPaymentWrites._
 import com.gu.zuora.rest.{ZuoraResponse, zuoraResponseReads}
-import play.api.libs.json.{JsObject, Reads}
+import play.api.libs.json.Reads
 
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -79,11 +79,14 @@ class ZuoraSoapService(restClient: rest.SimpleClient[Future])(implicit ec: Execu
         else throw QueryError(s"Zuora account update for ${accountId.get} was unsuccessful: ${response.error.getOrElse("")}")
       }
 
-  private def createObjectPaymentMethod(zObject: JsObject)(implicit logPrefix: LogPrefix): Future[String] =
+  private def createObjectPaymentMethod(request: CreatePaymentMethodObject)(implicit logPrefix: LogPrefix): Future[String] =
     restClient
-      .post[JsObject, ObjectCreateResponse]("object/payment-method", zObject)
+      .post[CreatePaymentMethodObject, ObjectCreateResponse]("object/payment-method", request)
       .map(_.valueOr(error => throw QueryError(s"Zuora REST create payment method failed: $error")))
-      .map(response => if (response.success) response.id else throw QueryError("Zuora create payment method was unsuccessful"))
+      .map { response =>
+        if (response.success) response.id.getOrElse(throw QueryError("Zuora create payment method succeeded but returned no Id"))
+        else throw QueryError("Zuora create payment method was unsuccessful")
+      }
 
   private def setDefaultPaymentMethod(accountId: AccountId, paymentMethodId: String, paymentGateway: PaymentGateway)(implicit
       logPrefix: LogPrefix,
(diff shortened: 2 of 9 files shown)


shtukas (approved):

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
```diff
@@ -0,0 +1,99 @@
...
+  * and payment-method creation via POST /v1/object/payment-method. The object API takes the same PascalCase zObject fields the SOAP create used, so
+  * the field mapping mirrors com.gu.zuora.soap.writers.Command and the CreateCreditCardReferencePaymentMethod action.
+  */
+object ZuoraPaymentWrites {
+
+  implicit val jsObjectWrites: Writes[JsObject] = Writes(identity)
+
+  /** PUT /v1/accounts/{id}. A None defaultPaymentMethodId is written as an explicit JSON null, which Zuora treats as "clear the default" — the REST
+    * equivalent of the SOAP fieldsToNull. See https://developer.zuora.com/api-references/api/operation/PUT_Account
+    */
```
graham228221: I feel like these comments would be better as just comments in the PR rather than in the CODE

Once we complete this migration, a future developer reading this code is not likely to have any context around the SOAP API

graham228221 (approved):
Tested in CODE and looks to be working well

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
andrade0: Good point, thanks. I've stripped the SOAP references out of the comments here and in ZuoraSoapService, keeping just what stays useful going forward (the null-clears-the-default behaviour and the doc links). Done in 758a771.

membership-attribute-service/app/components/TouchpointComponents.scala
```diff
@@ -109,7 +109,7 @@ class TouchpointComponents(
         metrics = zuoraMetrics,
       )
 
-    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraSoapClient, zuoraRestClient) with HealthCheckableService {
+    lazy val simpleZuoraSoapService = new ZuoraSoapService(zuoraRestClient) with HealthCheckableService {
       override def checkHealth: Boolean = zuoraSoapClient.isReady
     }
```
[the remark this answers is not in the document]
johnduffell: true, can line 105-110 be removed completely?

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
```diff
@@ -0,0 +1,98 @@
...
+
+/** JSON for the Zuora payment-method write operations: account payment updates via PUT /v1/accounts/{id}, and payment-method creation via POST
+  * /v1/object/payment-method (which takes the PascalCase zObject fields of the payment method).
+  */
+object ZuoraPaymentWrites {
+
+  implicit val jsObjectWrites: Writes[JsObject] = Writes(identity)
+
+  /** PUT /v1/accounts/{id}. A None defaultPaymentMethodId is written as an explicit JSON null, which Zuora treats as clearing the default. See
+    * https://developer.zuora.com/api-references/api/operation/PUT_Account
```
johnduffell: where did you get that URL?

I ask because same as the last PR, this immediately redirects to https://developer.zuora.com/v1-api-reference/api/accounts/put_account
I'd rather put the up to date urls in.  But I don't even know where the outdated one came from.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
```diff
@@ -0,0 +1,98 @@
...
+import com.gu.zuora.soap.models.Commands
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+
+/** JSON for the Zuora payment-method write operations: account payment updates via PUT /v1/accounts/{id}, and payment-method creation via POST
+  * /v1/object/payment-method (which takes the PascalCase zObject fields of the payment method).
+  */
+object ZuoraPaymentWrites {
+
+  implicit val jsObjectWrites: Writes[JsObject] = Writes(identity)
```
johnduffell: why did you need a Writes for JsObject?  it seems an odd technique, normally I'd expect to model things with domain types rather than mixing low level and high level types
Looking at function paymentMethod, I think that is the actual Writes as it is taking the business object and returning a JsObject.

johnduffell: this is what copilot suggested, which is a reasonable analysis:

---

`jsObjectWrites` as an identity `Writes[JsObject]` isn’t adding any domain meaning, and the real serialization logic appears to live in `paymentMethod`, which is what should be modeled as the `Writes`.

So, yes, it would make sense to move that logic into a proper `Writes[PaymentMethodType]` (or whatever the domain type is) and remove the identity writer.

## Why this is better
- Clear intent: a Writes should describe how a domain model becomes JSON.
- Type safety: callers pass domain values, not pre-shaped JsObjects.
- Testability: you can unit-test one canonical writer instance.
- Less indirection: avoids “fake” implicit writers that just pass through.

## Suggested direction
Refactor from:

identity `Writes[JsObject]`
custom `paymentMethod(...) => JsObject`
to:

`implicit val paymentMethodWrites: Writes[PaymentMethodRequest]` (or equivalent)
use `Json.toJson(paymentMethodRequest)` at call sites
If there are two wire formats (e.g., direct debit vs card reference), model that with ADTs and pattern matching inside the writer.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
```diff
@@ -0,0 +1,98 @@
...
+      "autoPay" -> update.autoPay,
+    )
+  }
+
+  /** POST /v1/object/payment-method returns {"Id": "...", "Success": true}. See
+    * https://developer.zuora.com/api-references/older-api/operation/Object_POSTPaymentMethod
+    */
+  case class ObjectCreateResponse(id: String, success: Boolean)
+  implicit val objectCreateResponseReads: Reads[ObjectCreateResponse] =
+    ((__ \ "Id").read[String] and (__ \ "Success").read[Boolean])(ObjectCreateResponse.apply _)
```
johnduffell: if you're defining your own case classes, I'd recommend to use automatic derivation as it is less error prone.  (You can still use things like `AccountId` value classes or other custom Reads for the child properties, as long as their implicit Reads is in scope.)
https://www.playframework.com/documentation/3.0.x/ScalaJsonAutomated

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  // Zuora only accepts a fixed set of CreditCardType values; anything else is omitted (matches the old SOAP action).
```
johnduffell: please can you add a PR comment to the old code that this is based on in the SOAP action, so it's easy to check they match?

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
```diff
@@ -0,0 +1,99 @@
...
+    )
+  }
+
+  /** POST /v1/object/payment-method returns {"Id": "...", "Success": true}. See
+    * https://developer.zuora.com/api-references/older-api/operation/Object_POSTPaymentMethod
+    */
+  case class ObjectCreateResponse(id: String, success: Boolean)
+  implicit val objectCreateResponseReads: Reads[ObjectCreateResponse] =
+    ((__ \ "Id").read[String] and (__ \ "Success").read[Boolean])(ObjectCreateResponse.apply _)
+
```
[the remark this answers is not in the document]
johnduffell: not sure if the RestClient already handles the Success=false case and returns an Exception/failed Future.  Worth checking either way.

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
```diff
@@ -65,41 +60,48 @@ class ZuoraSoapService(soapClient: soap.Client, restClient: rest.SimpleClient[Fu
...
+      autoPay: Boolean,
+  )(implicit logPrefix: LogPrefix): Future[UpdateResult] =
+    restClient
+      .put[AccountPaymentUpdate, ZuoraResponse](
+        s"accounts/${accountId.get}",
+        AccountPaymentUpdate(defaultPaymentMethodId, paymentGateway.gatewayName, autoPay),
+      )
+      .map(_.valueOr(error => throw QueryError(s"Zuora REST account update for ${accountId.get} failed: $error")))
+      .map { response =>
+        if (response.success) UpdateResult(accountId.get)
```
johnduffell, line 78: seems a bit odd that we re-wrap the account id that was passed in.  Looking at the code, it's actually unused
https://github.com/guardian/members-data-api/blob/0fbaf7a50b04a0e7110a0c9ef94a7351b1147e9a/membership-attribute-service/app/controllers/PaymentUpdateController.scala#L188-L193

I assume there's a couple of cleanup PRs incoming after this one, to tidy up the "soap" naming and align the types with the REST api rather than using the SOAP types - if so, this is a prime thing to simplify at that point.

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
```diff
@@ -65,41 +60,48 @@ class ZuoraSoapService(soapClient: soap.Client, restClient: rest.SimpleClient[Fu
...
-    soapClient.authenticatedRequest(
-      action = UpdateAccountPayment(
-        accountId = accountId.get,
-        defaultPaymentMethodId = Clear,
-        paymentGatewayName = paymentGateway.gatewayName,
-        autoPay = Some(false),
-      ),
-    )
-  }
+  private def createObjectPaymentMethod(zObject: JsObject)(implicit logPrefix: LogPrefix): Future[String] =
```
johnduffell: (I think this is related to my previous comment - this should take a business  object rather than an arbitrary JsObject, and then it will use the Writes to serialise.)

membership-attribute-service/app/components/TouchpointComponents.scala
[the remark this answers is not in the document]
andrade0: Yep, removed it. It was only powering the health check's isReady, which is a SOAP login probe that doesn't mean anything now we're all REST, and could even fail the health check on a SOAP auth blip for something we no longer call. So I've dropped Zuora from the health check (Salesforce only now, and the comment already said Zuora wasn't critical). The dead soap package and the ZuoraSoapService rename I'll do in the follow up.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
andrade0: Good catch, no idea where the old one came from either. Swapped it for the one it redirects to (https://developer.zuora.com/v1-api-reference/api/accounts/put_account). The object-create link just below is already on the older-api path and doesn't redirect, so I left that one.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
andrade0: Agreed, that identity writer was a smell. Modelled it properly now: a Writes[CreatePaymentMethodObject] that pattern matches the Commands.PaymentMethod ADT, and createObjectPaymentMethod takes the business object and lets the writer serialise it. The Writes[JsObject] is gone.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
andrade0: Done for ObjectCreateResponse, it's Json.reads now with a PascalCase JsonConfiguration so it lines up with Id/Success. I kept AccountPaymentUpdate as a hand-written Writes though, because a None default has to serialise to an explicit null to clear the default and Json.writes would just omit the field.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
andrade0: The equivalent is createPaymentMethodWrites in membership-common/src/main/scala/com/gu/zuora/soap/writers/Command.scala. One thing worth flagging while comparing: the old action sent the card type raw with spaces removed, whereas here I normalise the common brands to Zuora's casing and pass anything else through spaces-removed, so it's a superset rather than byte-identical.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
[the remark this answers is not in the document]
andrade0: Made id an Option, so a failure response without an Id reads cleanly and hits the "unsuccessful" branch instead of blowing up in the parser. I checked the RestClient too and it already returns a Left (failed Future) on a non-2xx before it tries to parse, so this only really bites on a 200 with Success=false.

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
andrade0: Yeah, agreed. There's a cleanup PR coming right after this to drop the soap naming and move onto the REST types, I'll simplify this there.

membership-common/src/main/scala/com/gu/zuora/ZuoraSoapService.scala
andrade0: Done, same change as above: createObjectPaymentMethod now takes the CreatePaymentMethodObject business object and the Writes does the serialisation.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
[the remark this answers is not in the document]
johnduffell: ok that will work with reduced type safety, and we will have to deal with the Option unnecessarily.

To get the fully type safety the return type should be an Either or similar with success as the discriminator, so the Left is success=false and has no fields and the Right is success=true and has the Id field as required rather than Option.
Then the Either (or similar) can be flattened into the for comprehension thus making sure that things bail out when success is false without having to write extra code to handle it in the business logic.

(Normally in our code bases we deal with the success flag in the zuora client as it's a bit fiddly so it's a shame this repo presumably doesn't do that.)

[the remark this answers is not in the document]
andrade0: Yeah, fair point, done. It's a small sealed result now: Created(id) with the Id required, and Failed(reason) with no Id, and a Reads that discriminates on Success. createObjectPaymentMethod just pattern matches it, so no Option to unwrap and the success=false case bails out on its own. Dropped the PascalCase config too since the reader is explicit now.

Agreed on the wider point that this would sit better as success handling in a proper Zuora REST client. That's really the same thread as aligning these off the SOAP types, so I'll pull it in with the cleanup PR rather than half-do it here.

membership-common/src/main/scala/com/gu/zuora/rest/ZuoraPaymentWrites.scala
johnduffell: @andrade0  why do we need to normalise the common brands?  was it not being handled alright with the existing logic
https://github.com/guardian/members-data-api/blob/39baf406f8881699223c3328674d0b7fb990bf11/membership-common/src/main/scala/com/gu/zuora/soap/writers/Command.scala#L52

johnduffell: also, btw the most helpful way to link to existing code is by highlighting it in github ui and copying the "permalink".
image

Then you can easily see the referenced code inline like this:
 
https://github.com/guardian/members-data-api/blob/39baf406f8881699223c3328674d0b7fb990bf11/membership-common/src/main/scala/com/gu/zuora/soap/writers/Command.scala#L107-L124

andrade0: Good question, and no, there wasn't a good reason. The old action just stripped spaces and that was fine, I'd added the brand normalisation on top without any real justification. I've dropped it and gone back to just removing spaces (matching the old action) in the cleanup PR #1264. Thanks for pushing on it.

andrade0: Correction on this, having actually tested it in CODE: there is a good reason after all. Zuora's CreditCardType is a case-sensitive enum, and Stripe returns the brand lower-case now (e.g. "mastercard"), so sending it through unchanged makes Zuora reject the create. A Mastercard card update 500s without the normalisation and succeeds with it (confirmed both ways on the deployed CODE build), so I've kept it in the cleanup PR. Sorry for the back and forth, and thanks for making me verify it properly.

johnduffell: ok got it, sounds like we do need it.  Thanks for confirming and writing up the reasoning here for future reference!

PR #1264, opened 2026-07-15
Remove the SOAP layer from the Zuora integration

andrade0:
Follow-up to #1263, which moved the last Zuora calls off SOAP onto REST. This finishes the job and removes the SOAP layer entirely, split into commits so it's easy to follow:

- delete the now-dead SOAP transport (the Client, the actions, the XML writers and readers, the deserializers, `ServiceHelpers`, `ZuoraFilters`, and their tests)
- rename `ZuoraSoapService` to `ZuoraService`, and the `SoapClient` trait to `ZuoraClient`, now they make no SOAP calls
- move the data models out of `com.gu.zuora.soap.models` to `com.gu.zuora.models`, so the `com.gu.zuora.soap` package is gone completely
- remove the model types that were only used by the SOAP transport: the unused `Commands` (subscribe, contribute, amend, renew, ...) and the SOAP fault-parsing types in `errors` (the `ErrorType` enum, `ErrorHandler`, `ZuoraFault`, ...)

It also drops the unused `UpdateResult` from the payment write flow so those methods just return `Unit`, and makes the zuora `Error` types report their message via `getMessage` (they were logging as null, which hid the underlying Zuora reason).

No behaviour change. The models that are still used keep the same shapes, only their package moved. Around 2400 lines of dead code removed.

Note: I briefly tried simplifying the credit card type to plain space-stripping, but testing in CODE showed Zuora rejects a lower-case `CreditCardType` (e.g. a Mastercard update 500s), so the brand normalisation is kept. That change and its revert are both in the history.


shtukas (approved):

membership-attribute-service/app/components/TouchpointComponents.scala
johnduffell, line 110: we now have both ZuoraService and ZuoraRestService which are actually both REST so it could be confusing - of course it's nice to keep things simpler/smaller where possible, so maybe breaking it down simplifies things but is it worth looking at that in a subsequent PR?
If there's a clear difference in what the two do they could simply be given more focussed names.

johnduffell (approved):
yes looks great, thanks for coming back to the tidyup.  Just one comment about naming that doesn't quite make sense any more, possible for a future PR if it's not trivial.


Something missing which you are probably planning for a later PR is to move away from the legacy shapes of the SOAP style types, which means we can use the derived Reads/Writes rather than custom logic for the simple cases.
e.g. membership-common/src/main/scala/com/gu/zuora/rest/ZuoraQueryReads.scala but I didn't keep track of them all.

Things like a `Currency` Reads would still be needed, but the top level derived one would use it automatically.

Then that removes the extra mapping/abstraction between the actual JSON and the fields used in the code.

🚀 great work on the migration overall, this has been needing to do for a long time.

PR #1265, opened 2026-07-16
Read the card expiry as an Int instead of round-tripping through a String

andrade0:
A small step towards John's suggestion on #1264 of moving off the SOAP-shaped types so the JSON maps more directly to the fields used in the code.

Zuora returns the card expiry month/year as JSON numbers, but `Queries.PaymentMethod` stored them as `Option[String]` (the old SOAP shape) and `PaymentService` then parsed them straight back to `Int`. This reads them as `Option[Int]` directly and drops the round-trip (and the little `asInt` helper). No behaviour change.

The fuller type alignment / derived Reads is a bigger follow-up: some of the remaining fields have lenient parsing (an unrecognised currency/gateway/country becomes `None` rather than failing) and mismatched JSON keys (`WorkEmail`, `sfContactId__c`, `PaypalEmail`), so it needs the SOAP-shaped types redesigning rather than a like-for-like swap.


johnduffell (approved):
nice simplification you spotted there, also thanks for making it small and easy to review 👏

PR #1268, opened 2026-07-16
Read the payment method through the existing REST service

andrade0:
First step of the follow-up John suggested on #1264: move the ex-SOAP reads onto the existing REST functions to remove the duplication between `ZuoraService` and `ZuoraRestService`.

`ZuoraService.getPaymentMethod` and `ZuoraRestService.getPaymentMethod` already hit the same `object/payment-method/{id}` endpoint, just parsed into two different types (the SOAP-shaped `Queries.PaymentMethod` vs the thinner `PaymentMethodResponse`). This merges them: `PaymentMethodResponse` gains the card / direct-debit / PayPal fields it was missing, `PaymentService` reads through `ZuoraRestService`, and the ex-SOAP `getPaymentMethod`, `Queries.PaymentMethod` and its reader are deleted.

No behaviour change on the read: same JSON, same fields, same graceful handling, and the error path stays like-for-like (a failed fetch still fails rather than silently returning nothing). The card expiry flows through as Ints, following on from #1265. Coverage for the card/direct-debit parsing moves to a `PaymentMethodResponse` reads test.

More of the ex-SOAP reads (`getAccount`, `getContact`, `getPaymentSummary`) will move over in follow-up PRs.

### DIFF
--- a/membership-attribute-service/app/services/zuora/payment/PaymentService.scala
+++ b/membership-attribute-service/app/services/zuora/payment/PaymentService.scala
@@ -84,7 +84,6 @@ class PaymentService(zuoraService: ZuoraService, restService: ZuoraRestService)(
         buildBankTransferPaymentMethod(defaultMandateIdIfApplicable, bankTransfer, response)
       case paypal: ZuoraRestService.PaymentMethodDetails.PayPal =>
         Some(PayPalMethod(paypal.email, response.numConsecutiveFailures, response.paymentMethodStatus))
-      case _: ZuoraRestService.PaymentMethodDetails.Other => None
     }
 
   private def getNextBill(subscriptionNumber: SubscriptionNumber, account: Account, targetDate: LocalDate)(implicit
--- a/membership-attribute-service/app/services/zuora/payment/PaymentService.scala
+++ b/membership-attribute-service/app/services/zuora/payment/PaymentService.scala
@@ -55,18 +55,18 @@ class PaymentService(zuoraService: ZuoraService, restService: ZuoraRestService)(
 
   private def buildBankTransferPaymentMethod(
       defaultMandateIdIfApplicable: Option[String],
-      m: ZuoraRestService.PaymentMethodResponse,
+      response: ZuoraRestService.PaymentMethodResponse,
   ): Option[PaymentMethod] = {
     for {
-      mandateId <- m.mandateId.orElse(defaultMandateIdIfApplicable)
-      accountName <- m.bankTransferAccountName
-      accountNumber <- m.bankTransferAccountNumberMask
+      mandateId <- response.mandateId.orElse(defaultMandateIdIfApplicable)
+      accountName <- response.bankTransferAccountName
+      accountNumber <- response.bankTransferAccountNumberMask
       paymentMethod <-
-        (m.bankTransferType, m.bankCode) match {
+        (response.bankTransferType, response.bankCode) match {
           case (Some("SEPA"), _) =>
-            Some(Sepa(mandateId, accountName, accountNumber, m.numConsecutiveFailures, m.paymentMethodStatus))
+            Some(Sepa(mandateId, accountName, accountNumber, response.numConsecutiveFailures, response.paymentMethodStatus))
           case (_, Some(sortCode)) =>
-            Some(GoCardless(mandateId, accountName, accountNumber, sortCode, m.numConsecutiveFailures, m.paymentMethodStatus))
+            Some(GoCardless(mandateId, accountName, accountNumber, sortCode, response.numConsecutiveFailures, response.paymentMethodStatus))
           case _ => None
         }
     } yield paymentMethod
@@ -74,18 +74,18 @@ class PaymentService(zuoraService: ZuoraService, restService: ZuoraRestService)(
 
   private def buildPaymentMethod(
       defaultMandateIdIfApplicable: Option[String] = None,
-      m: ZuoraRestService.PaymentMethodResponse,
+      response: ZuoraRestService.PaymentMethodResponse,
   ): Option[PaymentMethod] =
-    m.paymentMethodType match {
+    response.paymentMethodType match {
       case "CreditCard" | "CreditCardReferenceTransaction" =>
-        val isReferenceTransaction = m.paymentMethodType == "CreditCardReferenceTransaction"
+        val isReferenceTransaction = response.paymentMethodType == "CreditCardReferenceTransaction"
         val details =
-          (m.creditCardNumber |@| m.creditCardExpirationMonth |@| m.creditCardExpirationYear)(PaymentCardDetails)
-        Some(PaymentCard(isReferenceTransaction, m.creditCardType, details, m.numConsecutiveFailures, m.paymentMethodStatus))
+          (response.creditCardNumber |@| response.creditCardExpirationMonth |@| response.creditCardExpirationYear)(PaymentCardDetails)
+        Some(PaymentCard(isReferenceTransaction, response.creditCardType, details, response.numConsecutiveFailures, response.paymentMethodStatus))
       case "BankTransfer" =>
-        buildBankTransferPaymentMethod(defaultMandateIdIfApplicable, m)
+        buildBankTransferPaymentMethod(defaultMandateIdIfApplicable, response)
       case "PayPal" =>
-        Some(PayPalMethod(m.payPalEmail.get, m.numConsecutiveFailures, m.paymentMethodStatus))
+        Some(PayPalMethod(response.payPalEmail.get, response.numConsecutiveFailures, response.paymentMethodStatus))
       case _ => None
     }
 
@@ -131,11 +131,11 @@ class PaymentService(zuoraService: ZuoraService, restService: ZuoraRestService)(
       paymentMethodId <- maybePaymentMethodId
     } yield restService
       .getPaymentMethod(paymentMethodId)
-      .withLogging(s"get payment method for $maybePaymentMethodId")
       .map {
         case \/-(paymentMethod) => buildPaymentMethod(defaultMandateIdIfApplicable, paymentMethod)
         case -\/(error) => throw new RuntimeException(s"Failed to get payment method $paymentMethodId: $error")
-      })
+      }
+      .withLogging(s"get payment method for $maybePaymentMethodId"))
       .getOrElse(Future.successful(None))
 
 }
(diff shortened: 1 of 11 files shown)


shtukas (approved):

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
```diff
@@ -72,19 +74,18 @@ class PaymentService(zuoraService: ZuoraService, restService: ZuoraRestService)(
 
   private def buildPaymentMethod(
       defaultMandateIdIfApplicable: Option[String] = None,
-      soapPaymentMethod: Queries.PaymentMethod,
+      m: ZuoraRestService.PaymentMethodResponse,
```
johnduffell: `m` looks like a placeholder, consider a more descriptive name (same above on line 58, looks like it was an existing issue)

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
```diff
@@ -324,13 +324,62 @@ object ZuoraRestService {
...
+      payPalEmail: Option[String] = None,
+      bankTransferType: Option[String] = None,
+      bankTransferAccountName: Option[String] = None,
+      bankTransferAccountNumberMask: Option[String] = None,
+      bankCode: Option[String] = None,
+      creditCardNumber: Option[String] = None,
+      creditCardExpirationMonth: Option[Int] = None,
+      creditCardExpirationYear: Option[Int] = None,
+      creditCardType: Option[String] = None,
+      paymentMethodStatus: Option[String] = None,
```
johnduffell: are we just being very conservative by making them all optional apart from type?
If it is because the underlying data is a discriminated union on the type, we can make it into a sealed trait to avoid that (e.g. presuambly payPalEmail is required for type paypal, but not for anything else)

johnduffell (commented):
looks good, my only concern is that although the PaymentMethodResponse type is conservative, we end up with a lot of Options.
If we need to look at the zuora docs and/or bigquery to get a feel for whether they're really optional and which come together, then it would be worth doing I think.

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
```diff
@@ -128,9 +129,13 @@ class PaymentService(zuoraService: ZuoraService, restService: ZuoraRestService)(
...
-    } yield for {
-      soapPaymentMethod <- zuoraService.getPaymentMethod(paymentMethodId).withLogging(s"get payment method for $maybePaymentMethodId")
-    } yield buildPaymentMethod(defaultMandateIdIfApplicable, soapPaymentMethod))
+    } yield restService
+      .getPaymentMethod(paymentMethodId)
+      .withLogging(s"get payment method for $maybePaymentMethodId")
+      .map {
+        case \/-(paymentMethod) => buildPaymentMethod(defaultMandateIdIfApplicable, paymentMethod)
+        case -\/(error) => throw new RuntimeException(s"Failed to get payment method $paymentMethodId: $error")
+      })
```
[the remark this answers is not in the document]
andrade0: Good catch, moved `withLogging` after the `.map` so a `-\/` (which comes back as a successful `Future`) is now logged as a failure rather than a success.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
```diff
@@ -324,13 +324,62 @@ object ZuoraRestService {
...
-      (JsPath \ "Type").read[String] and
-      (JsPath \ "LastTransactionDateTime").read[String].map(isoDateStringAsDateTime)
-  )(PaymentMethodResponse.apply _)
+  implicit val paymentMethodReads: Reads[PaymentMethodResponse] = Reads { json =>
+    for {
+      numConsecutiveFailures <- (json \ "NumConsecutiveFailures").validateOpt[Int]
+      paymentMethodType <- (json \ "Type").validate[String]
+      lastTransactionDateTime <- (json \ "LastTransactionDateTime").validateOpt[String].map(_.map(isoDateStringAsDateTime))
+      mandateId <- (json \ "MandateID").validateOpt[String]
+      tokenId <- (json \ "TokenId").validateOpt[String]
```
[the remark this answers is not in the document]
andrade0: Fixed, `LastTransactionDateTime` now parses through a `JsResult` helper (`isoDateStringAsDateTimeResult`), so a malformed value becomes a `JsError` instead of a thrown exception that escapes the `Reads`. A missing field still reads as `None`. Added a test for the malformed case.

membership-attribute-service/app/services/zuora/payment/PaymentService.scala
andrade0: Renamed to `response` in both build methods. Thanks, yes it was carried over from the old code.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
andrade0: Done. The type-specific part is now a sealed trait `PaymentMethodDetails` (`Card` / `BankTransfer` / `PayPal` / `Other`), discriminated on the payment method type. The fields that are genuinely common (`numConsecutiveFailures`, `lastTransactionDateTime`, `paymentMethodStatus` and the type) stay at the top level, since Zuora returns them for every type. That lets the type-specific fields be non optional where the type guarantees them, and the PayPal email is required now which also removes the `.get`. An unrecognised type parses as `Other` so we never fail the whole payload. Added tests per case, including a PayPal with no email failing to parse.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
```diff
@@ -43,6 +44,15 @@ object ZuoraRestService {
   def jsStringOrNull(value: Option[String]) = value.map(JsString(_)).getOrElse(JsNull)
   def isoDateStringAsDateTime(dateString: String): DateTime = ISODateTimeFormat.dateTimeParser().parseDateTime(dateString)
 
+  /** Parse an ISO datetime into the JsResult error channel: a malformed value becomes a JsError rather than a thrown exception that would escape the
+    * Reads and crash parsing of the whole payload.
+    */
+  def isoDateStringAsDateTimeResult(dateString: String): JsResult[DateTime] =
+    Try(isoDateStringAsDateTime(dateString)).fold(
+      error => JsError(s"could not parse '$dateString' as an ISO 8601 datetime: ${error.getMessage}"),
```
johnduffell: if you do error.toString instead of getMessage, it also adds the name of the exception which is more helpful.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
```diff
@@ -324,60 +334,83 @@ object ZuoraRestService {
...
+        expirationYear: Option[Int],
+        cardType: Option[String],
+        isReferenceTransaction: Boolean,
+    ) extends PaymentMethodDetails
+    case class BankTransfer(
+        mandateId: Option[String],
+        transferType: Option[String],
+        accountName: Option[String],
+        accountNumberMask: Option[String],
+        bankCode: Option[String],
```
johnduffell, line 354: good to have these grouped so they don't appear where they don't make sense, but are these genuinely optional still?

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
```diff
@@ -324,60 +334,83 @@ object ZuoraRestService {
...
+    ) extends PaymentMethodDetails
+    case class PayPal(email: String) extends PaymentMethodDetails
+    // A payment method type we don't render (e.g. a new gateway's type). Kept so parsing never fails on an unknown type.
+    case class Other(paymentMethodType: String) extends PaymentMethodDetails
```
johnduffell: it looks like ultimately we return a 500 error and log that the payment method was missing, which is misleading.  It thinks that it was lost when it's actually just an unknown type.
https://github.com/guardian/members-data-api/blob/cf61aed6995fc0b4c4edfe8269c80def435cf639/membership-attribute-service/app/controllers/PaymentUpdateController.scala#L126-L128
I suspect that failing parsing would produce a more useful error and stack trace, and the code would be simpler.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
johnduffell: thanks yes that looks a lot better, now the type system gives us the guarantees once parsing has completed.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
andrade0: Good idea, switched to `error.toString` so the exception type shows up too.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
andrade0: These ones are genuinely optional. Zuora returns the direct debit fields per subtype (a SEPA method has the mandate and IBAN, a Bacs / GoCardless one has the mandate, sort code and account), so a given method won't carry all of them. `buildBankTransferPaymentMethod` already degrades to no payment method rather than rendering a half-built one when a field it needs is missing. Making them required would risk a 500 on `/mma` if Zuora omits one, which is the thing we've been trying to avoid on the read path, so I've kept them optional.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
andrade0: Agreed, done. Dropped the `Other` case and now fail parsing on a type we don't model, with a `JsError` that names the actual type. That surfaces it loudly with a useful stack trace instead of looking like a lost payment method, and the match in `buildPaymentMethod` is exhaustive now so it's simpler. None of our current types hit this (Card, CreditCardReferenceTransaction, BankTransfer and PayPal cover everything, PayPal Complete Payments included since the object API reports it as `PayPal`), so it only bites if Zuora ever returns a genuinely new type, which is exactly when we'd want to know.

membership-attribute-service/app/services/zuora/rest/ZuoraRestService.scala
johnduffell: ok fair enough either way I think it's a good improvement already and trying to reduce it further into subtypes might be more effort than it's worth. 👍

johnduffell (approved):
thanks for going through and making all these improvements 👏 this should leave us in a better state going forward 👍

andrade0:
Thanks John, glad it's in a better shape now. I'm off on holiday today and tomorrow, so feel free to merge it whenever suits you if you want it in sooner, otherwise I'll merge it when I'm back. Either way works for me.

